PR #13, opened 2014-06-08
Added the ability to cache responses to HEAD requests

TJ-Cutajar:
A subsequent GET request to the same URI will result in a call to origin. A cache entry that exists from a previous GET can be used to satisfy a HEAD request.


fxbonnet:
Can someone close this pull request? The changes have been merged and the JIRA is closed.

PR #25, opened 2015-01-07
RequestBuilder add Patch support

timja:
Adds the Http Patch method to the RequestBuilder


ok2c:
Committed to SVN trunk
http://svn.apache.org/r1650292

Oleg

PR #27, opened 2015-02-23
Being able to inject RequestConfig with CDI

nykolaslima:
CDI specification requires default constructor to being able to inject the bean.

Since CDI is a Oracle specification and is getting used even more, I believe `httpclient` should support it.
This PR adds a default constructor to `RequestConfig` class, allowing it to be injected with CDI.

### DIFF
--- a/httpclient/src/main/java/org/apache/http/client/config/RequestConfig.java
+++ b/httpclient/src/main/java/org/apache/http/client/config/RequestConfig.java
@@ -60,6 +60,14 @@ public class RequestConfig implements Cloneable {
     private final int socketTimeout;
     private final boolean decompressionEnabled;
 
+    /**
+     * CDI eyes only
+     * @deprecated CDI eyes only
+    */
+    public RequestConfig() {
+		this(false, null, null, false, null, false, false, false, 0, false, null, null, 0, 0, 0, false);
+    }
+    
     RequestConfig(
             final boolean expectContinueEnabled,
             final HttpHost proxy,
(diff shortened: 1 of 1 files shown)


ok2c:
Does the constructor need to be public to make CDI happy? Can it be protected or package private?

nykolaslima:
CDI reference:
```
5.4.1. Unproxyable bean types

Certain legal bean types cannot be proxied by the container:

classes which don't have a non-private constructor with no parameters,
classes which are declared final or have final methods,
primitive types,
and array types.
If an injection point whose declared type cannot be proxied by the container resolves to a bean with a normal scope, the container automatically detects the problem and treats it as a deployment problem.
```

So I believe it can be protected, I'll change it.

nykolaslima:
@ok2c done, what do you think?

ok2c:
According to my reading of the above section of the spec the constructor could also be package private which I would very much prefer.

nykolaslima:
@ok2c Actually it can't: `classes which don't have a non-private constructor with no parameters`.

it needs a `non-private constructor`. Actually there is a discussion about it here: https://developer.jboss.org/thread/240412

ok2c:
All right. I'll merge the change-set in as soon as we have 4.5 branch. Please feel free to nag me in two weeks, if it does not happen by then.

nykolaslima:
Ok, I'll wait for this. Thank you @ok2c

ok2c:
Committed to SVN trunk
http://svn.apache.org/r1662807

nykolaslima:
Thank you @ok2c. Can I close this PR?

ok2c:
Yes, please do so.

nykolaslima:
Thank you.

PR #29, opened 2015-03-13
Provide convenience functions for constructing the default cookie spec r...

grumpyjames:
...egistry and to build on top of the default entries.

Use them to construct the default registry from HttpClientBuilder when no cookie spec registry has been provided.

This allows clients to easily add their own CookieSpecProviders without having to duplicate the list of defaults.

Two further entirely cosmetic changes followed.


ok2c:
I committed your changes to SVN trunk with some minor tweaks
http://svn.apache.org/r1666565

Please review.

grumpyjames:
That looks great, thank you :-)

PR #32, opened 2015-06-03
Cache request line in HttpRequestWrapper

hirthwork:
Copy of https://github.com/apache/httpclient/pull/31 for 4.5.x

For each request HttpAsyncClient calls HttpRequestWrapper.getRequestLine() 8 times during single request execution. For URIs containing non-ASCII characters this causes tons of allocations and completely ruins performance.
According to out tests RequestLine caching successfully addresses the issue with the cost of single long living object.


ok2c:
Committed to SVN trunk and 4.5.x branch.
http://svn.apache.org/r1683475

PR #36, opened 2015-10-12
Allow usage of MultipartFormEntity in HttpAsyncClient.

Slikey:
There currently is only a workaround for the usage of MultipartFormEntity which requires sending the binary data of the MultipartFormEntity to a NByteArrayEntity. Actually I couldn't get this workaround working and it appears to not capture all part of the purpose of MultipartFormEntity.
See: http://mail-archives.apache.org/mod_mbox/hc-httpclient-users/201311.mbox/%3C13034392.132.1384195516775.JavaMail.Hal@Harold-Rosenbergs-MacBook-Pro.local%3E

This fix allows the usage of the MultipartFormEntity in the HttpAsyncClient because it calls #getContent() instead of #writeTo().

See people who have a similar problem:
http://stackoverflow.com/questions/4720077/how-can-i-see-the-content-of-a-multipartform-request
https://github.com/Mashape/unirest-java/issues/31

Something I want to point out is this quote: "That's the problem with trying to mix an asynchronous transport with an inherently blocking content producer. MultipartFormEntity can only produce its content using #writeTo method and cannot be used asynchronously. You basically have two options. (1) Write out entity content to ByteArrayOutputStream (or similar) and then create NByteArrayEntity using the resultant byte array. (2) Create multipart entity implementation that can produce its content asynchronously. " - Source: http://mail-archives.apache.org/mod_mbox/hc-httpclient-users/201311.mbox/%3C13034392.132.1384195516775.JavaMail.Hal@Harold-Rosenbergs-MacBook-Pro.local%3E

I'd like to get some discussion going on this approach and end up having a solution to this problem because I had to recreate part of the HttpMime in my project to support MultipartFormEntity in my HttpAsyncClient.

Thank you for reading!

### DIFF
--- a/httpmime/src/main/java/org/apache/http/entity/mime/MultipartFormEntity.java
+++ b/httpmime/src/main/java/org/apache/http/entity/mime/MultipartFormEntity.java
@@ -33,7 +33,11 @@ import org.apache.http.entity.ContentType;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.protocol.HTTP;
 
-import java.io.*;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 
 class MultipartFormEntity implements HttpEntity {
 
@@ -91,10 +95,11 @@ class MultipartFormEntity implements HttpEntity {
 
     @Override
     public InputStream getContent() throws IOException {
-        try (ByteArrayOutputStream outstream = new ByteArrayOutputStream()) {
-            writeTo(outstream);
-            return new ByteArrayInputStream(outstream.toByteArray());
-        }
+        ByteArrayOutputStream outstream = new ByteArrayOutputStream();
+        writeTo(outstream);
+        InputStream instream = new ByteArrayInputStream(outstream.toByteArray());
+        outstream.close();
+        return instream;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


ok2c:
If you want this patch committed to the stable 4.5 branch could you please make your code Java 1.6 compatible and also avoid using * imports?

Any development related discussions should be held at dev@hc.apache.org

Slikey:
Thank you for the response. I've change the code to compile against 1.6 and removed the * imports.

ok2c:
I would add one last thing here. I think it would be reasonable to not attempt to buffer the entity content if it is too large or its length is unknown. I would put a cap on the content length to 20K or something like that.

Slikey:
Hello, I don't want to disable this feature for requests above 20K, because the content can easily get higher than 20K due to the binary data which is carried inside. I'd rather propose that I instantiate the InputStream and fill it from a different Thread if blocking is a concern.
Are there any Utilities in this repository for that? Since the code has to be 1.6 compatible, the PipedInputStream is not available for my use.

ok2c:
One can always use BufferedHttpEntity to explicitly buffer entities of arbitrary length or explicitly write multipart content out to a byte[]. What I avoid is causing substantial memory allocation for unsuspected clueless users.

ok2c:
Committed to SVN trunk and 4.5.x branch with some modifications
http://svn.apache.org/viewvc?rev=1709814&view=rev

jacohend:
The 20k restriction is much *more* of a problem for unsuspecting users trying to use a library that is literally intended to, among other functions, upload binary data. At no point should my network library be making extremely low content length restrictions for me. If you want users to be aware that the library makes a full additional copy and to use BufferedHttpEntity in the case of large binary files, then I would suggest putting it in the documentation.

ok2c:
You are welcome to contribute improved javadocs / documentation.

PR #38, opened 2015-12-20
String format fix

meggermo:
There are two issues in the toString method of OSGiProxyConfiguration
- the number of provided arguments does not match the format string
- the proxyExceptions argument is an array so it will not pretty-print.

This pull request fixes both issues and demonstrates the fix with a JUnit test.

Possible improvement for the JUnit test:
Set access modifier to package for string constants in OSGiProxyConfiguration so I can refer to them in the JUnit test. This would make the test less brittle to (internal) code modifications.


ok2c:
Committed to SVN trunk and 4.5.x branch
http://svn.apache.org/viewvc?rev=1721386&view=rev

PR #40, opened 2015-12-28
Delete obsolete clone method

meggermo:
In svn revision 980937 the Cloneable interface was removed from the class CacheEntity but the clone method was not. Assuming the removal of the Cloneable interface was intentional, the clone method should have been removed as well.

See https://fisheye6.atlassian.com/changelog/httpcomponents?cs=980937 for the entire change set.


garydgregory:
Committed to git master. Please verify and close. In Apache's Jira as [HTTPCLIENT-1709].

PR #41, opened 2015-12-31
HTTPCLIENT-1710 - fixes - examine all exceptions, correct capture group count

enenum:
Also in case there are more than one ProxyConfiguration services registered, the proxy from the first enabled one encountered will be used. From the user perspective, they have no control over which one that is. The whole idea of multiple ProxyConfigurations (its a factory) seems largely unsupported at this time.


ok2c:
Committed to SVN trunk and 4.5.x branch.
http://svn.apache.org/viewvc?rev=1722632&view=rev

PR #46, opened 2016-02-13
HTTPCLIENT-1718 and HTTPCLIENT-1719 changes to allow proxy authentication

enenum:
Some changes to OSGiCredentialsProvider to allow user and pass to be read from configuration, also change to allow user value in config to be accessible to the class


ok2c:
@enenum Is 967056f necessary?

ok2c:
@enenum Essentially, I should only merge 8ac1fe5? Please confirm.

ok2c:
Committed to SVN trunk and 4.5.x branch. Please verify and close.

PR #47, opened 2016-03-02
HTTPCLIENT-1726: Copy the SNI fix from SSLConnectionSocketFactory to …

dbaxa:
…the deprecated SSLSocketFactory class.

Signed-off-by: David Black dblack@atlassian.com


dbaxa:
cc @ok2c

PR #55, opened 2016-07-10
HTTPCLIENT-1748: fix for trunk

sebastiencaille:
Hello,
as discussed before, here is a patch proposal for the trunk.
(I added a basic test case)

Cheers
Sébastien

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/SystemDefaultCredentialsProvider.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/SystemDefaultCredentialsProvider.java
@@ -27,7 +27,9 @@
 package org.apache.hc.client5.http.impl.auth;
 
 import java.net.Authenticator;
+import java.net.MalformedURLException;
 import java.net.PasswordAuthentication;
+import java.net.URL;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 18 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/SystemDefaultCredentialsProvider.java
```diff
@@ -88,35 +93,51 @@ public void setCredentials(final AuthScope authscope, final Credentials credenti
...
+            final HttpContext context) {
         final String hostname = authscope.getHost();
         final int port = authscope.getPort();
         final HttpHost origin = authscope.getOrigin();
         final String protocol = origin != null ? origin.getSchemeName() :
                 (port == 443 ? "https" : "http");
+
+        final URL targetHostURL;
+        if (context != null) {
+            final HttpUriRequest httpUriRequest = (HttpUriRequest) context.getAttribute(HttpCoreContext.HTTP_REQUEST);
```
ok2c: One cannot assume the request object in the context to be always of type of HttpUriRequest. It would be safer to use the generic HttpRequest type here. 

Otherwise everything looks good to me.

sebastiencaille:
is it ok to use httpRequest.getRequestLine().getUri() instead ?

ok2c:
Committed to SVN trunk. Please review and close the PR.

PR #56, opened 2016-07-22
HTTPCLIENT-1752 - Allow to configure the OSGI clients with relaxed SSL checks

simonetripodi:
This is my proposal to add HTTPCLIENT-1752 feture.
Feedbacks, suggestions, hints, ... are always welcome!

### DIFF
--- a/httpclient5-osgi/src/main/java/org/apache/hc/client5/http/osgi/impl/RelaxedLayeredConnectionSocketFactory.java
+++ b/httpclient5-osgi/src/main/java/org/apache/hc/client5/http/osgi/impl/RelaxedLayeredConnectionSocketFactory.java
@@ -38,15 +38,21 @@ import org.apache.hc.client5.http.socket.LayeredConnectionSocketFactory;
 import org.apache.hc.client5.http.ssl.SSLConnectionSocketFactory;
 import org.apache.hc.core5.http.HttpHost;
 import org.apache.hc.core5.http.protocol.HttpContext;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
 
 final class RelaxedLayeredConnectionSocketFactory implements LayeredConnectionSocketFactory {
 
     private final SSLConnectionSocketFactory defaultSocketFactory = getSocketFactory();
 
-    private final TrustedHostsConfiguration configuration;
+    private final BundleContext bundleContext;
 
-    public RelaxedLayeredConnectionSocketFactory(final TrustedHostsConfiguration configuration) {
-        this.configuration = configuration;
+    private final ServiceRegistration trustedHostConfiguration;
+
+    public RelaxedLayeredConnectionSocketFactory(final BundleContext bundleContext,
+                                                 final ServiceRegistration trustedHostConfiguration) {
+        this.bundleContext = bundleContext;
+        this.trustedHostConfiguration = trustedHostConfiguration;
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


garydgregory:
Hi Simone!

Nice to see a patch from you.

Is there a way all this new code can be unit tested?

I know that running an OSGi container in a "unit" test is heavy-handed but it is do-able. What about for some of the other pieces?

Gary

simonetripodi:
Thanks a lot @garydgregory for jumping in, very appreciated!
I just added another missing test for the `HostMatcherTest` class, for the `OSGiProxyConfiguration` it was already there. 

Do you have any hint on how to setup a unit test for the `RelaxedLayeredConnectionSocketFactory` class? It is a `LayeredConnectionSocketFactory` which accepts SSL connection with server with self-signed certificates... Many thanks in advance!

tmaret:
bq. Allows the new HTTP Client instance to blindly trust all SSL self-signed certificates. Pay attention on no enabling this feature in production environment as it is totally insecure!

In order to avoid production environments to use this insecure configuration, how about splitting the code in two bundles ? 
The base bundle `httpclient5-osgi` would feature an extension mechanism that allow to plug a new bundle containing the insecure code and configuration.

simonetripodi:
@tmaret I have to think a little on how to wire internal component in order to implement your idea - if it is not an urgent constraint, I'd defer that "feature" once the pull is merged in the codebase - WDYT?
Thanks a lot for your hint, really appreciated!

tmaret:
@simonetripodi fine with me if it is possible to split later on.

httpclient5-osgi/src/main/java/org/apache/hc/client5/http/osgi/impl/RelaxedLayeredConnectionSocketFactory.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.hc.client5.http.socket.LayeredConnectionSocketFactory;
+import org.apache.hc.client5.http.ssl.SSLConnectionSocketFactory;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.protocol.HttpContext;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
+
+final class RelaxedLayeredConnectionSocketFactory implements LayeredConnectionSocketFactory {
+
+    private final SSLConnectionSocketFactory defaultSocketFactory = getSocketFactory();
```
ok2c: @simonetripodi If you inject the default `LayeredConnectionSocketFactory` through constructor instead of creating it internally this class could be unit-tested using Mockito. Both `BundleContext` and `LayeredConnectionSocketFactory` can be mocked out with verified with Mockito.

simonetripodi: Thanks for the hint, I am going to add this behaviour!

simonetripodi: @ok2c I just implemented the required behaviour in my last commit, tests are up and running - any feedback would be more than appreciated, thanks in advance!

simonetripodi:
@tmaret do you already have an idea on how to split the _trust all_ in a separated configuration belonging in a different bundle, while keeping the core in this bundle?

tmaret:
@simonetripodi I did not dig this much but you may be able to add a ProviderType api in the `httpclient5-osgi` bundle. 
The API may be enough if it allows to provide a `LayeredConnectionSocketFactory`.

The separate bundle would contain and provide the RelaxedLayeredConnectionSocketFactory together with its configuration.
If the second bundle is not provided, a default `LayeredConnectionSocketFactory` (safe one) implementation would be selected.

PR #60, opened 2016-08-24
Support changing system default ProxySelector

PissedCapslock:
The `ProxySelectorRoutePlanner` class which got deprecated in favor of the `SystemDefaultRoutePlanner` could deal with:
- `null` as default `ProxySelector`
- a change in the default `ProxySelector`

This change ports that behavior to the `SystemDefaultRoutePlanner`.


ok2c:
Committed to SVN trunk and 4.5.x branch. Please review and close.

PR #61, opened 2016-09-13
HTTPCLIENT-1767: Null pointer dereference in EofSensorInputStream and ResponseEntityProxy

ansell:
This fixes HTTPCLIENT-1767 by adding null checks to ResponseEntityProxy and dereferencing instance variables before null checks in EofSensorInputStream.

For reference here, the stacktrace, against httpclient-4.5.2, which I can't reproduce reliably, but which should be fixed by these changes is:

```
java.lang.NullPointerException: null
	at org.apache.http.impl.execchain.ResponseEntityProxy.streamClosed(ResponseEntityProxy.java:140)
	at org.apache.http.conn.EofSensorInputStream.checkClose(EofSensorInputStream.java:228)
	at org.apache.http.conn.EofSensorInputStream.close(EofSensorInputStream.java:174)
```


ok2c:
Committed to SVN trunk and 4.5.x branch. Please review and close.

ansell:
Thanks for patching this so promptly! The commit you applied looks correct.

PR #65, opened 2017-02-21
Make some package protected classes public as we need to extend them in current jmeter

pmouawad:
Hello,
Would it be possible to make those classes public as for now in JMeter, to use validation mechanism introduced in 4.3 we had to create a custom class located in  package "org.apache.http.impl.conn" which is clearly a hack.
In the future we will probably migrate directly to HttpClient5.

Thanks

### DIFF
--- a/RELEASE_NOTES.txt
+++ b/RELEASE_NOTES.txt
@@ -8,6 +8,9 @@ Please note that as of 4.4 HttpClient requires Java 1.6 or newer.
 Changelog:
 -------------------
 
+* [HTTPCLIENT-1817] Add a "Trust All" TrustStrategy implementation.
+  Contributed by Gary Gregory <ggregory at apache.org>
+
 * [HTTPCLIENT-1816] Update Apache Commons Codec 1.9 to 1.10.
   Contributed by Gary Gregory <ggregory at apache.org>
 
--- a/httpclient/src/main/java-deprecated/org/apache/http/impl/conn/HttpConnPool.java
+++ b/httpclient/src/main/java-deprecated/org/apache/http/impl/conn/HttpConnPool.java
@@ -43,7 +43,7 @@ import org.apache.http.pool.ConnFactory;
  * @deprecated (4.3) no longer used.
  */
 @Deprecated
-class HttpConnPool extends AbstractConnPool<HttpRoute, OperatedClientConnection, HttpPoolEntry> {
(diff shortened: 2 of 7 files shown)


fxbonnet:
Hi,
It looks like these classes are deprecated. I think the best way for you to implement your validation mechanism would be to create a subclass of org.apache.http.pool.AbstractConnPool

ok2c:
+1 to @fxbonnet's comment

pmouawad:
@ok2c , @fxbonnet , I am aware of the fact that classes are deprecated, see how we use them and the javadocs:

- https://github.com/apache/jmeter/blob/trunk/src/protocol/http/org/apache/http/impl/conn/JMeterPoolingClientConnectionManager.java

Doing what you propose means for me  copy/pasting the 3 classes in JMeter, which I don't find it better.
Migrating fully to 4.5 new code is identified:

- https://bz.apache.org/bugzilla/show_bug.cgi?id=58757

Still, the migration is not a little piece of cake and for us does not bring enough ROI to be done.
JMeter relied on "advanced" api of HC4 and although I know you make a lot of effort, documentation for migration was not complete at all for such parts.

@ok2c helped a lot on Cookie migration which was important feature for us. We're using new APIs for new code , like Async Backend Listener that interracts with InfluxDB, but as you know we have to make some choice as time is not infinite :-) 

In the future and if possible, it would be nice when you deprecate an API to mention in javadoc its new class equivalent. It's not done everywhere.

As Http2 is expected in 5.0, for now I must say I prefer to do migration at that time.

But @ok2c  if you don't want to merge the PR, I'll do that.

ok2c:
@pmouawad I'll happily help with the migration (either to 4.5 or 5.0 depending on your preference) but opening up deprecated APIs looks wrong to me.

pmouawad:
@ok2c , thanks for your proposal. I understand. As I have seen some PR on deprecated API (SNI retro fix) I thought it was possible.

I'll be very happy if you can help on migration. 5.0 being in alpha mode for now, do you think we should wait for migration or migrate already to more up to date API of 4.5.X ? Will it reduce the amount of work at end ?

Thanks

ok2c:
@pmouawad it is one thing to back-port a fix or a feature that has no API impact but this is a whole different thing to start opening up private methods. 
It will take a while for 5.0 to mature. It really depends if you want to migrate early and have influence on API development or prefer convenience of API stable releases.

PR #67, opened 2017-03-19
HTTPCLIENT-1830: await termination of the IdleConnectionEvictor when it is shut down

theangrydev:
See [HTTPCLIENT-1830](https://issues.apache.org/jira/browse/HTTPCLIENT-1830) for more details.

I also noticed something weird with the original implementation of this in [HTTPCLIENT-1583](https://issues.apache.org/jira/browse/HTTPCLIENT-1583) that I think means that `HttpClientBuilder.evictExpiredConnections` is not working as intended, see my comment on that ticket.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/IdleConnectionEvictor.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/IdleConnectionEvictor.java
@@ -27,11 +27,13 @@
 
 package org.apache.hc.client5.http.impl;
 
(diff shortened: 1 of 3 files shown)


theangrydev:
As discussed in [HTTPCLIENT-1830](https://issues.apache.org/jira/browse/HTTPCLIENT-1830), updated this pull request to keep await termination separate from the shutdown. Keeping the uncaught exception logging in trunk but not in 4.5.x and 4.6.x.

theangrydev:
This has been merged.

PR #68, opened 2017-03-19
HTTPCLIENT-1830: await termination of the IdleConnectionEvictor when it is shut down (4.5.x)

theangrydev:
Same as #67 but for 4.5.x branch.


theangrydev:
As discussed in [HTTPCLIENT-1830](https://issues.apache.org/jira/browse/HTTPCLIENT-1830), updated this pull request to keep await termination separate from the shutdown and keeping the change set to a minimum.

theangrydev:
This has been merged.

PR #70, opened 2017-03-21
HTTPCLIENT-1835: evictExpiredConnections no longer causes the evictIleConnections behaviour to be enabled when evictIdleConnections is not explicitly enabled

theangrydev:
See [HTTPCLIENT-1835](https://issues.apache.org/jira/browse/HTTPCLIENT-1835).


theangrydev:
Merged

PR #72, opened 2017-03-21
HTTPCLIENT-1835: evictExpiredConnections no longer causes the evictIleConnections behaviour to be enabled when evictIdleConnections is not explicitly enabled (4.6.x)

theangrydev:
Same as #70 but for 4.6.x


theangrydev:
Merged

PR #76, opened 2017-05-09
Allow to add exec interceptor as the first and the last

pavolloffay:
Related to OpenTracing integration #74 and https://issues.apache.org/jira/browse/HTTPCLIENT-1842?focusedCommentId=15996807&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15996807 

cc @ok2c

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
@@ -176,7 +176,7 @@ public class HttpAsyncClientBuilder {
 
     private static class ExecInterceptorEntry {
 
-        enum Postion { BEFORE, AFTER, REPLACE }
+        enum Postion { BEFORE, AFTER, REPLACE, FIRST, LAST }
 
         final ExecInterceptorEntry.Postion postion;
         final String name;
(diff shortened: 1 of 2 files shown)


ok2c:
Looks good to me. Could you please do the same for the async client builder?

pavolloffay:
Last/first interceptor added to the `AsyncClient`.

pavolloffay:
When is going the release of 5.0 happen?

ok2c:
HttpClient 5.0-alpha2 vote is ongoing.

pavolloffay:
> That seems quite reasonable. Some unit tests would be nice as well.

@garydgregory Could you please point me to existing tests for `HttpClientBuilder#addExecInterceptorBefore` and its alternatives? I wasn't able to find them.

> HttpClient 5.0-alpha2 vote is ongoing.

When is the GA planned for? I have to look at async client, what is necessary for tracing.

garydgregory:
I do not see any either . You can be the first ;-)

HttpCore 5 GA has to come first. Oleg can speak to schedule. I think it will be a while though. We are in alpha mode.

pavolloffay:
Could it go in without tests? I don't know how the test should look like. 

To implement tests I need to access `NamedElementChain<ExecChainHandler>` which is constructed inside `HttpClientBuilder.build()`. I could define a new property with a package visibility getter in the builder, but it's not so nice.

ok2c:
@pavolloffay Please bear with me a few more days. I'll merge your changes as soon as we are done migrating to Git

ok2c:
@pavolloffay The PR got automatically closed as a result of migration from svn to git. I'll get it merged anyway

pavolloffay:
@ok2c thanks!

PR #77, opened 2017-05-18
Avoid fetching entity from cache's datastore twice

leandronunes85:
This is just a simple fix for the problem being discussed in the mailing list (subject of the thread is "HttpCacheStorage#getEntry called twice on cache hit"). 

I applied the fix proposal on branch 4.5.x because that's the version I'm using and opened this PR just to get some feedback on the changes before patching other branches (4.6.x and master). 

I separated the patch in two different commits:

- 10ab37b ("Removing unnecessary checks for cache invalidation calls."): removed a few references to "cacheInvalidatorWasCalled()" from tests where this didn't seem to be the primary focus of the test (basically after applying the change to the production code all these tests went red for no apparent reason).
- f59181e ("Avoid fetching the cached entity twice on cache hit."): actual change and unit test to avoid regression. If the changes to the tests in 10ab37b are correct then the change to production code should not have broken anything :).

Notes: 
1. Locally I cannot simply run "mvn install" as there's some problems reported by maven-javadoc-plugin (had to comment it out to try this locally but didn't commit those changes for obvious reasons). Am I missing some configuration to be able to run this?
2. Probably this module deserved a big refactor but as this is the first time I'm contributing to ASF I wanted to keep things simple and the changes as small as possible.
3. I'm not sure about what the process to contributing to this project is, so I'm trying this approach. Please correct me if this is not the way to go.

### DIFF
--- a/httpclient-cache/src/main/java/org/apache/http/impl/client/cache/CachingExec.java
+++ b/httpclient-cache/src/main/java/org/apache/http/impl/client/cache/CachingExec.java
@@ -263,10 +263,9 @@ public class CachingExec implements ClientExecChain {
         requestCompliance.makeRequestCompliant(request);
         request.addHeader("Via",via);
 
-        flushEntriesInvalidatedByRequest(context.getTargetHost(), request);
-
         if (!cacheableRequestPolicy.isServableFromCache(request)) {
             log.debug("Request is not servable from cache");
+            flushEntriesInvalidatedByRequest(context.getTargetHost(), request);
             return callBackend(route, request, context, execAware);
         }
 
--- a/httpclient-cache/src/test/java/org/apache/http/impl/client/cache/TestCachingExec.java
+++ b/httpclient-cache/src/test/java/org/apache/http/impl/client/cache/TestCachingExec.java
@@ -402,6 +402,26 @@ public class TestCachingExec extends TestCachingExecChain {
         verifyMocks();
     }
 
+    @Test
+    public void testDoesNotFlushCachesOnCacheHit() throws Exception {
+        requestPolicyAllowsCaching(true);
+        requestIsFatallyNonCompliant(null);
+
+        getCacheEntryReturns(mockCacheEntry);
+        doesNotFlushCache();
+        cacheEntrySuitable(true);
+        cacheEntryValidatable(true);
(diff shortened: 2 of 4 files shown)


ok2c:
>  Am I missing some configuration to be able to run this?

HC 4.x still requires Java 1.6. You are probably using Java 1.8 to build the project. Try downgrading to Java 1.7

> Probably this module deserved a big refactor but as this is the first time I'm contributing to ASF I wanted to keep things simple and the changes as small as possible

I guess we would all welcome volunteers willing to improve HttpClient cache code

> I'm not sure about what the process to contributing to this project is, so I'm trying this approach. Please correct me if this is not the way to go

Just keep PRs coming. For the time being Github integration appears broken. I am trying to get it fixed. So, you might want to subscribe to dev@hc.apache.org as well.

ok2c:
John, would you be able to review the patch?

comcast-jonm:
@leandronunes85 : first, thanks for structuring your commits this way; makes it really easy to review. I agree that the cache flushes are redundant here, and as long as `TestProtocolRequirements` and `TestProtocolRecommendations` are passing, then I'm confident this change should be approved and merged. I don't have a working dev environment at the moment, so I suggest that if @ok2c can see the tests pass then we should merge.

Thanks, @leandronunes85! 👍

ok2c:
@leandronunes85 all PRs have been pulled. Could you please review and close those PRs that are still open?

leandronunes85:
Thanks for the support @ok2c, @comcast-jonm  and everyone else. Is there an ETA for releasing these patches? When can I expect having them available for use (specially in the 4.5.x branch)? Also, is there anything I could do to speed this up?

I'm short on time for extra projects ATM but I plan to take a look into refactoring the caching module and propose a 'plan' for this (that I would like someone to review before actually starting implementing it). Does this sound a good plan or there is already some plan in progress?

ok2c:
@leandronunes85 
> When can I expect having them available for use (specially in the 4.5.x branch)?
Realistically, not sooner than Q3 2017
> Does this sound a good plan or there is already some plan in progress?
Presently no one is working on the caching module, so this plan is as good as any. Any improvements would be quite welcome

PR #81, opened 2017-08-07
BasicCookieStore no longer synchronizes on reads

cakofony:
BasicCookieStore uses a ReentrantReadWriteLock to avoid
synchronization on getCookies/toString while maintaining
thread safety.


ok2c:
Please review and close.

MarkMielke:
I am going to follow up here on the mailing list for best direction. This change broke some of my code that relies on BasicCookieStore being Serializable. This change breaks the Serializable contract by introducing a transient "lock" variable (without qualifying it as transient), doesn't provide readObject()/writeObject() methods to deal with the transient variable, and doesn't update the serialVersionUID so it continues to try and read the objects already written that have the same serialVersionUID. I think this change should probably be reverted?

PR #84, opened 2017-09-19
Fix to HTTPCLIENT-1869

pmouawad:
Hello, 
Fix is to apply to 4.5.X

Thank you !
Regards
@philmdot


ok2c:
The work-around looks misplaced to me. If this issue affects GZIP encoded content only the work-around should be placed in `GZIPInputStream` only.

ok2c:
@pmouawad I urge you to close this PR and completely re-think the approach to solving this issue (see HTTPCLIENT-1869 for details)

PR #85, opened 2017-09-28
HTTPCLIENT-293 Fix proposal based on RFC 7578

sermojohn:
I implemented the fix in two commits, because I believe that some refactoring was required in order to handle the part header field parameters (name, filename) properly in the Content-Disposition part header field. Also a unit test was implemented that fails on the first commit but succeeds on the second that includes the actual patch.

I did some research about the percent encoding, which as I can understand is quite loose concerning the characters that it should always encode depending on the context. The actual percent character encoding implementation was copied from the common-codecs' URLCodec but it could not be reused as it was implemented, because that class includes URL specific encoding (e.g. ' ' -> '+')

### DIFF
--- a/httpmime/src/main/java/org/apache/http/entity/mime/FormBodyPartBuilder.java
+++ b/httpmime/src/main/java/org/apache/http/entity/mime/FormBodyPartBuilder.java
@@ -27,13 +27,14 @@
 
 package org.apache.http.entity.mime;
 
-import java.util.List;
-
 import org.apache.http.entity.ContentType;
 import org.apache.http.entity.mime.content.AbstractContentBody;
 import org.apache.http.entity.mime.content.ContentBody;
 import org.apache.http.util.Args;
 import org.apache.http.util.Asserts;
+
+import java.util.List;
+import java.util.Map;
 
 /**
  * Builder for individual {@link org.apache.http.entity.mime.FormBodyPart}s.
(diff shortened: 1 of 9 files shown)


ok2c:
@sermojohn Do you mind making a few relatively minor changes to your patch? 
1. Could you please keep `MinimalField` immutable and move all construction logic to `FormBodyPartBuilder`?
2. `FormBodyPartBuildeencodeForHeader` appears to be no longer needed. It should probably be removed.
3. Could you please avoid different formatting of imports? You might need to tweak your IDE settings for that.

sermojohn:
@ok2c pushed the suggested changes. Thank you for the feedback.

Removed both FormBodyPartBuilder.encodeForHeader and the corresponding unit test, as I did not find anywhere in the spec this encoding rule.

Fyi, I could not build the 4.6.x branch with java 1.6 version, as it is reported by animal-sniffer-maven-plugin that undefined references are used (AuthSchemeRegistry.java:127, SchemeRegistry.java:151,  CookieSpecRegistry.java:139). It worked when I switched to java 1.8 source compilation and target.

ok2c:
@sermojohn Is there any reason why you encode parameters of the CONTENT_DISPOSITION field only? Why not all fields?

sermojohn:
Only the filename parameter of the Content-Disposition field is currently encoded, because it is defined by the specification (RFC 7578) that should support non US-ASCII characters. The other parameters, "name" (in Content-Disposition) and "charset" (in Content-Type) should only contain US-ASCII characters.

I could modify the code that creates the "charset" parameter of Content-Type field, but I noticed that this field always appears without quotes around the value, which makes it a bit incompatible with the handling of the Content-Disposition parameters that always appear quoted. So, I believe I should avoid changing that code, as it also seems irrelevant with the current fix. Hope you agree.

ok2c:
@sermojohn
I committed your changes with some modifications. I fixed multiple style check violations, escaped non Latin chars in test cases, and redesigned your code to generate fewer intermediate objects / garbage on the heap.

Please review and close this PR. Feel free to raise another PR with follow-up changes.

ok2c:
@sermojohn Please close this PR.

ok2c:
@vlsi This change-set is present in the 5.x code line only.

vlsi:
Would you please adjust "fix version" in https://issues.apache.org/jira/browse/HTTPCLIENT-293 accordingly?

PR #89, opened 2017-12-01
add dropwizard metrics to httpclient

xiaoshuang-lu:


ok2c:
@xiaoshuang-lu We cannot just add an extra dependency to the stable (GA) release branch.

xiaoshuang-lu:
All right, @ok2c. Which branch shall be used?

ok2c:
@xiaoshuang-lu master. I have a question, though. Why should HttpClient depend on dropwizard in the first place?

xiaoshuang-lu:
It would be great if httpclient has the ability to collect critical metrics (requests-per-seconds, response-time, etc.) and dropwizard metrics provides what we want.

ryantxu:
Isn't this already supported via:
http://metrics.dropwizard.io/3.2.3/manual/httpclient.html

ok2c:
@xiaoshuang-lu Have you looked at the new 5.0 APIs, more specifically at the exec interceptors? If you find exec interceptors insufficient to collect all the requires metrics feel free to propose additional plug-in points, but please do not drag in a new dependency without a very good reason

xiaoshuang-lu:
ExecChainHandler means that one user need to implement the metrics collectors by self and another user need to do the same work too. These things are duplicated. From my perspective, users love the frameworks which are capable of native metric functionalities. Think about gRPC.

xiaoshuang-lu:
Thanks, @ryantxu. I will investigate metrics-httpclient.

ok2c:
@xiaoshuang-lu I do not see gRPC require dropwizard

xiaoshuang-lu:
@ok2c gRPC needs opencensus, something like dropwizard metrics.

ok2c:
@xiaoshuang-lu and what if I _want_ to use dropwizard?

xiaoshuang-lu:
Do you mean dependency conflicts? :-)

ok2c:
@xiaoshuang-lu No, I mean forcing all users to use one particular framework without any good reason.

petermant:
Should this be closed? We have dependencies on httpclient 4.5.x as well as dropwizard 1.3.5, which uses metrics-core 4.0.2. Making httpclient dependent on metrics 3.2.5 risks incompatibilities if breaking API changes are made between 3.x and 4.x.

ok2c:
@petermant @xiaoshuang-lu I also would prefer to have this PR closed.

PR #91, opened 2018-01-14
HTTPCLIENT-1690: Discarded any Content-Encoding headers coming with 304 status

catchsudheera:
Discarded any Content-Encoding headers coming with 304 status code while updating the cache entry. Update performed on cache entry is not updating the content, so we should not update the Content-Encoding header as well ( even if the 304 response has it)

### DIFF
--- a/httpclient-cache/src/main/java/org/apache/http/impl/client/cache/CacheEntryUpdater.java
+++ b/httpclient-cache/src/main/java/org/apache/http/impl/client/cache/CacheEntryUpdater.java
@@ -116,7 +116,7 @@ class CacheEntryUpdater {
                 continue;
             }
             final Header[] matchingHeaders = headerGroup.getHeaders(responseHeader.getName());
-            for (Header matchingHeader : matchingHeaders) {
+            for (final Header matchingHeader : matchingHeaders) {
                 headerGroup.removeHeader(matchingHeader);
             }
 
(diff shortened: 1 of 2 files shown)


httpclient-cache/src/main/java/org/apache/http/impl/client/cache/CacheEntryUpdater.java
```diff
@@ -103,6 +103,8 @@ public HttpCacheEntry updateCacheEntry(
 
     protected Header[] mergeHeaders(final HttpCacheEntry entry, final HttpResponse response) {
 
+        // since we do not expect a content in the response, remove any related headers if exists
+        response.removeHeaders(HTTP.CONTENT_ENCODING);
```
ok2c: @catchsudheera Sudheera, can we avoid mutation of the original response message? It would be cleaner to ignore this header when copying updated headers from the response message instead of deleting it from the original message itself. To make things easier you might want to port `mergeHeaders` method from master
https://github.com/apache/httpcomponents-client/blob/master/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheUpdateHandler.java#L139

ok2c:
@catchsudheera Committed to 4.5.x, 4.6.x and master. Please review and close this PR.

catchsudheera:
@ok2c Actually, in master branch commit. you missed one change. Surprised why tests didn't fail and found out that test is not complete. Updated test case for 4.5.x and 4.6.x and added missing change for master branch. Sent separate PRs for each branch.. closing this PR

PR #94, opened 2018-01-15
HTTPCLIENT-1690: Avoid updating the cache entry with Content-Encoding header 

catchsudheera:
Avoid updating the cache entry with Content-Encoding headers too. (missing piece from last commit). Updated the test case


ok2c:
@catchsudheera Committed to master. Please review and close this PR.

PR #95, opened 2018-01-20
Workaround for misbehaved servers that return HTTP 204 responses with a content

agherardi:
Also see https://github.com/apache/httpcomponents-core/pull/57


ok2c:
@agherardi Looks good to me. However this fix should go into HttpCore [DefaultConnectionReuseStrategy](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/impl/DefaultConnectionReuseStrategy.java). A unit test would also be nice

PR #100, opened 2018-03-07
Information responses (1xx) processing implemented.

kirill-usov:


ok2c:
@kirill-usov Kirill, could you please briefly describe what you are trying to accomplish with these changes? I think HttpClient 5.0 already supports 1xx message processing through `AsyncClientExchangeHandler` interface.

kirill-usov:
Hello.
I need to handle 103 Early Hints response from the server (RFC 8297). But I did not found any way handle it in async client when I am creating default, minimal or build custom one with help HttpAsyncClientBuilder. 
It is what I have implemented.

ok2c:
@kirill-usov It most definitely works with minimal HttpAsyncClient
```
final IOReactorConfig ioReactorConfig = IOReactorConfig.custom()
        .setSoTimeout(Timeout.ofSeconds(5))
        .build();

final MinimalHttpAsyncClient client = HttpAsyncClients.createMinimal(
        HttpVersionPolicy.NEGOTIATE,
        H2Config.DEFAULT,
        H1Config.DEFAULT,
        ioReactorConfig);

client.start();

final URI requestUri = new URI("http://httpbin.org/get");
final HttpHost target = new HttpHost(
        requestUri.getHost(),
        requestUri.getPort() > 0 ? requestUri.getPort() : 80,
        requestUri.getScheme());
final Future<AsyncClientEndpoint> leaseFuture = client.lease(target, null);
final AsyncClientEndpoint endpoint = leaseFuture.get(30, TimeUnit.SECONDS);
try {

    final CountDownLatch latch = new CountDownLatch(1);
    endpoint.execute(new BasicClientExchangeHandler<SimpleHttpResponse>(
                             SimpleRequestProducer.create(SimpleHttpRequest.get(requestUri)),
                             SimpleResponseConsumer.create(),
                             new FutureCallback<SimpleHttpResponse>() {

                                 @Override
                                 public void completed(final SimpleHttpResponse response) {
                                     latch.countDown();
                                     System.out.println(requestUri + "->" + response.getCode());
                                 }

                                 @Override
                                 public void failed(final Exception ex) {
                                     latch.countDown();
                                     System.out.println(requestUri + "->" + ex);
                                 }

                                 @Override
                                 public void cancelled() {
                                     latch.countDown();
                                     System.out.println(requestUri + " cancelled");
                                 }

                             }) {

                         @Override
                         public void consumeInformation(final HttpResponse response) throws HttpException, IOException {
                             System.out.println("infornation response:" + response);
                         }

                     },
            HttpClientContext.create());

    latch.await();
} finally {
    endpoint.releaseAndReuse();
}

System.out.println("Shutting down");
client.shutdown(ShutdownType.GRACEFUL);
```

ok2c:
What I do not like about your patch is that it introduces another callback interface (`HttpClientAsyncExecutionCallback`). I believe it would be cleaner to add `#informationResponse` to the `AsyncResponseConsumer` interface instead.

garydgregory:
Wouldn't adding a method to an existing interface break BC? Are we OK with that or would we/should we introduce a subinterface?

ok2c:
@garydgregory I see no problem with making small profile API changes to 5.0 while we are still in BETA

kirill-usov:
AsyncResponseConsumer interface is a part of core. I did not want to change core, because it can be a reason of regression on products using it. And I tried to keep all existing interfaces unchanged to avoid regressions.

ok2c:
@kirill-usov HC 5.0 is still BETA. Some API changes are still to be expected. We are still many months away from the API freeze. I suggest that we do the right thing and fix what needs fixing. Please raise a PR for HttpCore and close this one.

garydgregory:
@ok2c : Oops, I did not realize we were talking about 5.0, sorry. So yeah, changing the current interface makes sense.

PR #101, opened 2018-04-29
Ignore cached ETag and Last-Modified headers of HEAD cache entries

gesellix:
Since httpclient-cache 4.4 `HEAD` requests can be cached with a `null` response body. Subsequent `GET` requests to the same target then use the headers "If-None-Match" and "If-Modified-Since" to leverage caching via HTTP protocol. In that case a server can respond with status code 304 and an empty response body, which leads to an invalid cache entry.

The issue seems to be introduced with PR #13. The repo https://github.com/gesellix/httpclient-cache-demo shows the behaviour for HttpClients 4.3, 4.4, and 4.5. 

This PR, based on the 4.5.x tag, tries to recognise the described invalid state and suppresses caching on the HTTP layer, so that the server has to return the complete response.

JIRA issue: https://issues.apache.org/jira/browse/HTTPCLIENT-1920

Signed-off-by: Tobias Gesellchen <tobias@gesellix.de>

### DIFF
--- a/httpclient-cache/src/main/java/org/apache/http/impl/client/cache/ConditionalRequestBuilder.java
+++ b/httpclient-cache/src/main/java/org/apache/http/impl/client/cache/ConditionalRequestBuilder.java
@@ -58,14 +58,19 @@ class ConditionalRequestBuilder {
             throws ProtocolException {
         final HttpRequestWrapper newRequest = HttpRequestWrapper.wrap(request.getOriginal());
         newRequest.setHeaders(request.getAllHeaders());
-        final Header eTag = cacheEntry.getFirstHeader(HeaderConstants.ETAG);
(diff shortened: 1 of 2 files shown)


gesellix:
We might consider not to add cache entries with a missing response body in the first place, which would obsolete PR #13 to some extend (initial `HEAD` requests for a specific target won't ever be cached anymore).

ok2c:
Related to HTTPCLIENT-1920

comcast-jonm:
Yes, this approach is not the right one, as @fxbonnet mentioned. 👎 

I'll comment on https://issues.apache.org/jira/browse/HTTPCLIENT-1920.

ok2c:
Closing as outdated.

PR #103, opened 2018-05-11
fixed incorrect connection close on shutdown + fixed corresponding test

overquota:
due to https://issues.apache.org/jira/browse/HTTPCLIENT-1923

close() method was changed due to Oleg@ suggestion

But thats was not enough, because it isnt called at all when connection manager is shutting down [entering in connection manager shutdown on client close](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/HttpClientBuilder.java#L1233) -> [shutdown method called](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/conn/BasicHttpClientConnectionManager.java#L376)

### DIFF
--- a/httpclient/src/main/java/org/apache/http/impl/conn/BasicHttpClientConnectionManager.java
+++ b/httpclient/src/main/java/org/apache/http/impl/conn/BasicHttpClientConnectionManager.java
@@ -157,7 +157,9 @@ public class BasicHttpClientConnectionManager implements HttpClientConnectionMan
 
     @Override
     public void close() {
-        closeConnection();
+        if (this.isShutdown.compareAndSet(false, true)) {
+            closeConnection();
+        }
     }
 
     HttpRoute getRoute() {
@@ -206,7 +208,7 @@ public class BasicHttpClientConnectionManager implements HttpClientConnectionMan
         };
     }
 
-    private void closeConnection() {
+    private synchronized void closeConnection() {
         if (this.conn != null) {
             this.log.debug("Closing connection");
             try {
@@ -214,20 +216,6 @@ public class BasicHttpClientConnectionManager implements HttpClientConnectionMan
             } catch (final IOException iox) {
                 if (this.log.isDebugEnabled()) {
                     this.log.debug("I/O exception closing connection", iox);
-                }
-            }
-            this.conn = null;
-        }
-    }
-
-    private void shutdownConnection() {
-        if (this.conn != null) {
-            this.log.debug("Shutting down connection");
-            try {
-                this.conn.shutdown();
-            } catch (final IOException iox) {
-                if (this.log.isDebugEnabled()) {
-                    this.log.debug("I/O exception shutting down connection", iox);
                 }
             }
             this.conn = null;
@@ -373,10 +361,8 @@ public class BasicHttpClientConnectionManager implements HttpClientConnectionMan
     }
 
     @Override
-    public synchronized void shutdown() {
-        if (this.isShutdown.compareAndSet(false, true)) {
-            closeConnection();
-        }
+    public void shutdown() {
+        close();
     }
 
 }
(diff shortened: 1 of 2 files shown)


httpclient/src/main/java/org/apache/http/impl/conn/BasicHttpClientConnectionManager.java
```diff
@@ -375,7 +375,7 @@ public synchronized void closeIdleConnections(final long idletime, final TimeUni
     @Override
     public synchronized void shutdown() {
         if (this.isShutdown.compareAndSet(false, true)) {
-            shutdownConnection();
```
ok2c: @overquota Should not `#close` do `#closeConnection` and `#shutdown` do `#shutdownConnection`?

overquota: I think no because `#shutdownConnection` closes connection with `RST` (https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/impl/BHttpConnectionBase.java#L304) - this is not normal tcp connection closing

While `#closeConnection` doing it more correctly (FIN+ACK): https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/impl/BHttpConnectionBase.java#L315

`#shutdown` may do `#shutdownConnection` only if `#shutdownConnection` will do close() on connection instance i think... but if so - i think there are no difference what shutdown method will do close instead of shutdown =)

ok2c: @overquota I am not sure I agree. `#shutdown` is supposed to be used to execute an abnormal connection termination (with `RST`), while `#close` is to be used for orderly connection termination (with `FIN+ACK`).

ok2c: @overquota Actually I just realized that the pooling connection manager does not distinguish between normal or abnormal pool shutdown. Persistent connections get closed out regardless whether the manager gets closed or shut down. For consistency sake I'll make the basic manager behave the same.

overquota: just for example:
`#shutdown` in [PoolingHttpClientConnectionManager](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.java#L410) really do `this.pool.shutdown();`
*BUT*
`this.pool.shutdown` doing [entry.close](https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/pool/AbstractConnPool.java#L148) - which is the same normal `conn.close()`

In addition: [HttpClientBuilder](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/HttpClientBuilder.java#L1232) knows only about `#close`, and on close it is doing... `#shutdown` :) - thats the main reason i've changed `#shutdown` method

In addition 2: [ClientConnectionManager declaration](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/conn/ClientConnectionManager.java#L112) does not know anything about `#close`, only `#shutdown`. And there are no docs about how connections would be terminated

Do you want me to refactor all these misleadings? No problems, really, but I don't know is it worth it?

ok2c:
@overquota Committed to 4.5.x and 4.6.x with some minor tweaks. Please review and close the PR.

overquota:
Thank you very much! Is there any deadline for publishing version with this fix in maven repo?

ok2c:
Q2/Q3 2018

PR #105, opened 2018-07-09
add 'oraclejdk10' to Travis build matrix

sullis:


ok2c:
The build was failing due to conflicts with Maven Surefire plugin and Mockito.

PR #107, opened 2018-07-15
AsyncPushConsumer now also accepts a HttpContext

ato:
These are the corresponding client changes for the core interface change in apache/httpcomponents-core#70.

This patch is not expected to compile without the corresponding core patch.

@ok2c should I add context as a parameter to start() as well and update the push example? Or would you prefer to keep the change minimal like this and let user code which needs access to the context override consumePromise() themselves?


ok2c:
@ato HttpClient 5.0 has been upgraded to HttpCore 5.0-beta3 with 1614c2516806780c80ddb94fc1414233da1e5da2. Could you please rebase your change-set off the latest master?

ok2c:
@ato Could you please close this PR?

PR #111, opened 2018-08-31
HTTPCLIENT-1944: Add hardCancellationEnabled option to RequestConfig

rschmitt:


rschmitt:
You can merge it as-is. Changing the default behavior may in turn require test code changes, so I'll send you a follow-up PR for that.

PR #118, opened 2018-11-13
Use connectTimeout as TLS handshake timeout

rschmitt:


ok2c:
@rschmitt Looks good to me.

PR #133, opened 2019-01-08
Shutdown executorService on AbstractHttpAsyncClientBase shutdown

Tinow:


Tinow:
Shutdown executorService on AbstractHttpAsyncClientBase shutdown.

PR #134, opened 2019-02-06
Always use the cached entity's content length.

JaysonRaymond:
Some well known proxies respond with Content-Length=0, when returning 304. For robustness, always use the cached entity's content length, as modern browsers do.


ok2c:
@JaysonRaymond I presume this one can be closed?

ok2c:
@JaysonRaymond  Could you please squash the commits into a single one?

ok2c:
@JaysonRaymond Committed as d150f5abae3fc4a51f18a63d2807bf703fa67dd7

PR #137, opened 2019-02-15
URIBuilder and URIUtils improvements 

ok2c:


httpclient-cache/src/main/java/org/apache/http/impl/client/cache/CacheKeyGenerator.java
```diff
@@ -63,45 +114,23 @@
      * @return String the extracted URI
      */
     public String getURI(final HttpHost host, final HttpRequest req) {
-        if (isRelativeRequest(req)) {
-            return canonicalizeUri(String.format("%s%s", host.toString(), req.getRequestLine().getUri()));
+        try {
+            final URI uri = normalize(getRequestUri(req, host));
```
cstamas: Why `uri` local var?

ok2c: @cstamas No reason. Just old habit. It is not a deal breaker, is it?

cstamas: Nope, is fine, I just found it strange 😄

ok2c: @cstamas I cant help it. `return normalize(getRequestUri(req, host)).toASCIIString();` just looks ugly to me.

cstamas: Agreed

PR #140, opened 2019-03-01
HTTPCLIENT-1969: Filter out weak cipher suites

artem-smotrakov:
Please consider a patch for [HTTPCLIENT-1969](https://issues.apache.org/jira/browse/HTTPCLIENT-1969):
- Defined a list of weak algorithms which may be used in a TLS connection. The list is based on the latest settings in modern OpenJDK, see [java.security](https://hg.openjdk.java.net/jdk/jdk/file/1019c97e1bde/src/java.base/share/conf/security/java.security#l678) file (EXPORT ciphers are also disabled in modern OpenJDK by default)
- Updated `SSLConnectionSocketFactory` to filter out weak ciphers if cipher suites are not explicitly set.

Please note that the test passes with latest Java versions even without patching `SSLConnectionSocketFactory` because latest Java versions disable weak ciphers by default. The filtering mechanism blocks weak ciphers in case older Java versions are used.

### DIFF
--- a/httpclient/src/main/java/org/apache/http/conn/ssl/SSLConnectionSocketFactory.java
+++ b/httpclient/src/main/java/org/apache/http/conn/ssl/SSLConnectionSocketFactory.java
@@ -36,7 +36,9 @@ import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import javax.net.SocketFactory;
 import javax.net.ssl.HostnameVerifier;
@@ -163,6 +165,15 @@ public class SSLConnectionSocketFactory implements LayeredConnectionSocketFactor
     public static final X509HostnameVerifier STRICT_HOSTNAME_VERIFIER
         = StrictHostnameVerifier.INSTANCE;
 
+    private static final String WEAK_KEY_EXCHANGES
+            = "^(TLS|SSL)_(NULL|ECDH_anon|DH_anon|DH_anon_EXPORT|DHE_RSA_EXPORT|DHE_DSS_EXPORT|"
+            + "DSS_EXPORT|DH_DSS_EXPORT|DH_RSA_EXPORT|RSA_EXPORT|KRB5_EXPORT)_(.*)";
+    private static final String WEAK_CIPHERS
+            = "^(TLS|SSL)_(.*)_WITH_(NULL|DES_CBC|DES40_CBC|DES_CBC_40|3DES_EDE_CBC|RC4_128|RC4_40|RC2_CBC_40)_(.*)";
+    private static final List<Pattern> WEAK_CIPHER_SUITE_PATTERNS = Collections.unmodifiableList(Arrays.asList(
+            Pattern.compile(WEAK_KEY_EXCHANGES, Pattern.CASE_INSENSITIVE),
+            Pattern.compile(WEAK_CIPHERS, Pattern.CASE_INSENSITIVE)));
+
     private final Log log = LogFactory.getLog(getClass());
 
     /**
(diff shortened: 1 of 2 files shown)


httpclient/src/main/java/org/apache/http/conn/ssl/SSLConnectionSocketFactory.java
```diff
@@ -183,6 +187,15 @@ public static SSLConnectionSocketFactory getSocketFactory() throws SSLInitializa
         return new SSLConnectionSocketFactory(SSLContexts.createDefault(), getDefaultHostnameVerifier());
     }
 
+    private static boolean isWeakCipherSuite(final String cipherSuite) {
+        for (final String weakAlgorithm : WEAK_ALGORITHMS) {
+            if (cipherSuite.contains(weakAlgorithm)) {
```
ok2c: @artem-smotrakov Could this be replaced with regexp matching?

artem-smotrakov: @ok2c I thought a list of weak algorithms is more straightforward. Regex would allow to build more complex constraints but not sure if it's necessary. TLS cipher suites have a pretty simple format where a particular crypto algorithm may be used only in particular part of a cipher suite.

ok2c: @artem-smotrakov Matching things by a simple occurrence of a expression at any position in the algorithm name looks really shoddy to me. I have no idea if this is a common practice but it looks sloppy.

artem-smotrakov: @ok2c I think strings from `WEAK_ALGORITHMS` can occur only in specific parts of a cipher suite. I am not sure about common practices either, but in this particular case, I think that having a list of weak algorithms looks clearer and simpler to understand rather than a regex. Anyway, if you think it has to be updated to use a regex, I'll do that. Please let me know.

ok2c: > WEAK_ALGORITHMS can occur only in specific parts of a cipher suite

@artem-smotrakov This is precisely what bothers me. The weak algorithms can occur at specific parts but the propose change-set simply tests for occurrence of those expressions at any arbitrary position in the cipher name. This looks sloppy.

artem-smotrakov: @ok2c The weak algorithms can't occur at arbitrary position in the cipher suites defined in TLS specifications. That's why I think it's enough to search for specific weak algorithms in a cipher suite. Again, no problem with updating it to use regex, I just don't think it's worth it.

artem-smotrakov: Just noticed one more thing. It may be better to use convert the string to upper-case before search or matching. The JSSE spec currently mentions only _anon_ but not sure if _ANON_ may not appear.

ok2c: @artem-smotrakov Yes, this must be fixed.

artem-smotrakov:
Addressed comments from @ok2c 
- Use regexp to catch weak cipher suites
- Normalize cipher suites names prior matching
- Added more weak algorithms even them they are not supported by the default JSSE implementation
- Updated tests

Please take a look.

httpclient/src/main/java/org/apache/http/conn/ssl/SSLConnectionSocketFactory.java
```diff
@@ -183,6 +193,16 @@ public static SSLConnectionSocketFactory getSocketFactory() throws SSLInitializa
         return new SSLConnectionSocketFactory(SSLContexts.createDefault(), getDefaultHostnameVerifier());
     }
 
+    static boolean isWeakCipherSuite(final String cipherSuite) {
+        final String normalizedCipherSuite = cipherSuite.toUpperCase();
```
ok2c: @artem-smotrakov Please use `#toUpperCase` with  `Locale#ROOT` parameter. Alternatively you might want to use non-case sensitive regex matching now.

artem-smotrakov: Good point!

artem-smotrakov:
Updated the patch:
- Use non-case sensitive regex matching
- Squash commits into a single one

@ok2c Please take a look.

ok2c:
@artem-smotrakov Would you also be willing to port this change-set to `master`?

artem-smotrakov:
To be honest, I am not convinced that regexp is better here, but it's not a big deal to me :) Let's agree to disagree, you have more experience with this project :) Sure, I'll port it to the master branch. Thanks for the review @ok2c !

PR #143, opened 2019-03-13
Changed route tracking of proxy to not default to false

gknapowski:
Changed route tracking of proxy to not default to false and instead to
use the route's secure value. this allow routes to proxies to be https and
not tunnel/CONNECT.  All test pass with this change.

### DIFF
--- a/httpclient/src/main/java/org/apache/http/impl/execchain/MainClientExec.java
+++ b/httpclient/src/main/java/org/apache/http/impl/execchain/MainClientExec.java
@@ -404,7 +404,7 @@ public class MainClientExec implements ClientExecChain {
                         timeout > 0 ? timeout : 0,
                         context);
                 final HttpHost proxy  = route.getProxyHost();
-                tracker.connectProxy(proxy, false);
+                tracker.connectProxy(proxy, route.isSecure() && !route.isTunnelled());
                 break;
             case HttpRouteDirector.TUNNEL_TARGET: {
                 final boolean secure = createTunnelToTarget(


ok2c:
@gknapowski I am not sure this is the right thing to do here. If the entire route can be considered secure once complete it does not mean its intermediate hops can be considered secure.
What is it exactly you are trying to achieve?

gknapowski:
Thank you for looking at this.

My use case is a secure break inspecting  proxy.=:  I must connect the proxy over HTTPS.  I must not use the CONNECT verb.

With a custom RoutePlanner I can achieve this with the exception of that one line.

I think in addition to the ‘secure’ field on the route, their needs to be a ‘secureProxy’ field?  That seemed like a massive change.

Is there a better way to achieve what I am trying to accomplish?

Thanks

ok2c:
@gknapowski I suppose for individual hops one can assume `https` scheme as being secure.

httpclient/src/main/java/org/apache/http/impl/execchain/MainClientExec.java
```diff
@@ -404,7 +404,7 @@ void establishRoute(
                         timeout > 0 ? timeout : 0,
                         context);
                 final HttpHost proxy  = route.getProxyHost();
-                tracker.connectProxy(proxy, false);
+                tracker.connectProxy(proxy, route.isSecure());
```
ok2c: @gknapowski it should most likely be `route.isSecure() && !route.isTunnelled()`, that is a direct (non-tunneled) connection to the proxy can be considered secure if the route is considered secure.

gknapowski: I agree.  I tested with use cases.  I also tested with a secure connection to the Proxy with Tunneling.  Both work.

ok2c: @gknapowski Could you please squash the commits into a single one?

gknapowski: Sorry about that.

gknapowski:
Thank You Oleg!

PR #146, opened 2019-04-23
Fix bug in URIBuilder#isPathEmpty method 

varunnvs92:
isPathEmpty should verify if encodedPath member is an empty string.
Otherwise, [this code](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/utils/URIUtils.java#L229) in URIUtils.rewriteURI will evaluate to false when normalization is disabled.

Before 4.5.6, URIUtils.rewriteURI method used [TextUtils#isEmpty method](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/util/TextUtils.java#L42) which checked for both null and empty string case.

* Added a new unit test. The test failed before including the empty string check

### DIFF
--- a/httpclient/src/main/java/org/apache/http/client/utils/URIBuilder.java
+++ b/httpclient/src/main/java/org/apache/http/client/utils/URIBuilder.java
@@ -34,7 +34,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-
 import org.apache.http.Consts;
 import org.apache.http.NameValuePair;
 import org.apache.http.conn.util.InetAddressUtils;
@@ -521,7 +520,8 @@ public class URIBuilder {
      * @since 4.5.8
      */
     public boolean isPathEmpty() {
-        return (this.pathSegments == null || this.pathSegments.isEmpty()) && this.encodedPath == null;
+        return (this.pathSegments == null || this.pathSegments.isEmpty()) &&
+               (this.encodedPath == null || this.encodedPath.isEmpty());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


varunnvs92:
@ok2c 
Thank you for the quick action. Is it possible to release `v4.5.9` in April. I am from AWS Java SDK team. Our customers are waiting for us to move to apache `4.5.8`. But we need this fix to upgrade as we need to disable normalization. Please let us know your release plans. I am happy to help if required.
Relavant Issues:
https://github.com/aws/aws-sdk-java/pull/1966
https://github.com/aws/aws-sdk-java/issues/1919

ok2c:
@varunnvs92 Usually we tend to have a release every quarter. This is a very trivial bug with a fairly trivial workaround. I do not think it warrants an immediate release.

varunnvs92:
@ok2c 
This bug is trivial. But we are effected by the normalization behavior change (see below links) which is causing [multiple failures](https://github.com/aws/aws-sdk-java/issues/1919) for our customers. The workaround suggested was to upgrade to `4.5.8` and disable normalization. But this bug is blocking us to move to `4.5.8` as it effects normalization disabled path. You can consider this bug as part of adoption blocker for `4.5.8`
https://issues.apache.org/jira/browse/HTTPCLIENT-1968
https://issues.apache.org/jira/browse/HTTPCLIENT-1983

I hope you reconsider this and do a release soon as this is a major priority for us. Lot of customers are waiting on us upgrading to `4.5.8`.

varunnvs92:
@ok2c 
Any thoughts on my last comment? This is a very important change for us to upgrade to latest version. I would be happy to help in doing code changes for releasing a new version.

ok2c:
@varunnvs92 I am not sure what you expect me to say in addition to what I have already said. I wish I had spare bandwidth for doing a release right now but I simply do not. There are other committers who you may what to ask to be a release manager for this release.

PR #147, opened 2019-04-28
MultipartEntity limit 25Kb Size

teahyuk:
Hello.
I use this library for access other server which file uploading is important part of one

so I want upload file over 25Kb with this entity but this entity limited 25Kb-(hardCode)
and I don't understand why are you guys limit this things

please give me your intent

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartFormEntity.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartFormEntity.java
@@ -95,9 +95,13 @@ class MultipartFormEntity implements HttpEntity {
     public InputStream getContent() throws IOException {
         if (this.contentLength < 0) {
             throw new ContentTooLongException("Content length is unknown");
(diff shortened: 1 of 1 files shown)


ok2c:
@teahyuk  The `getContent` method should never be called and this limit should never apply as long as the library is being used correctly. One should always be using `#writeTo` method to stream out the content of the entity.

teahyuk:
@ok2c thank you answer my question
but When I used with AsyncHttpClient with multipartEntity which builded with Filebody,
the AsyncHttpClient used `getContent` and threw EntityTooLongException from inside.

How can I use AsyncHttpClient with using body `writeTo`???

ok2c:
@teahyuk entities generated by `MultipartEntityBuilder` rely on classic (inherently blocking) `InputStream` / `OutputStream` APIs. Those do not work well with non-blocking agents such as `HttpAsyncClient`. You should either consider using classic `HttpClient` or buffer the entire entity in memory with `BufferedHttpEntity` wrapper class.

PR #148, opened 2019-05-16
Http Client part of information responses (1xx) processing.

kirill-usov:
I think it is a time to commit client part of the feature.


ok2c:
@kirill-usov Could you please rebase your changes off the master and squash them into a single commit?

PR #149, opened 2019-05-27
Information responses (1xx) processing implemented.

kirill-usov:
Http Client part of information responses (1xx) processing.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/DefaultAsyncCacheRevalidator.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/DefaultAsyncCacheRevalidator.java
@@ -131,6 +131,12 @@ class DefaultAsyncCacheRevalidator extends CacheRevalidatorBase {
                                 }
 
                                 @Override
+                                public void handleInformationResponse(final HttpResponse response) throws HttpException, IOException {
+                                    responseRef.set(response);
+                                    asyncExecCallback.handleInformationResponse(response);
+                                }
+
+                                @Override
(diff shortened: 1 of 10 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/DefaultAsyncCacheRevalidator.java
```diff
@@ -130,6 +130,12 @@ public AsyncDataConsumer handleResponse(
                                     return asyncExecCallback.handleResponse(response, entityDetails);
                                 }
 
+                                @Override
+                                public void handleInformationResponse(final HttpResponse response) throws HttpException, IOException {
+                                    responseRef.set(response);
```
ok2c: @kirill-usov Are you sure this is the right thing to do here? Should informational response have any impact on cache re-validation?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncRedirectExec.java
```diff
@@ -187,6 +192,19 @@ public AsyncDataConsumer handleResponse(
                 return asyncExecCallback.handleResponse(response, entityDetails);
             }
 
+            @Override
+            public void handleInformationResponse(final HttpResponse response) throws HttpException, IOException {
+                redirect(response);
```
ok2c: @kirill-usov Are you sure about doing a redirect on an informational response? This does not look right to me.

ok2c:
Superseded by b717fde2897c338566a95494dd392be046e5050f

PR #150, opened 2019-05-29
Escape DEL character when tracing

dowy:
DEL characters should be converted to [0x7f] in Wire traces otherwise they are difficult to see in logs.


garydgregory:
LGTM

PR #153, opened 2019-06-11
HTTPCLIENT-1992: Impossible to access trailer-headers 

serkanturgut:
1. MainClientExec changed in a way to recognize chunked encoding and create correct http entity wrapper for it, implemented a new http entity wrapper which works with ChunkedInputStream to reach trailers and propagate them to HttpEntity.getTrailers() API
2. Fixed a typo in ResponseEntityProxy.enhance() method name

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MainClientExec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MainClientExec.java
@@ -135,7 +135,7 @@ public final class MainClientExec implements ExecChainHandler {
                 execRuntime.releaseEndpoint();
                 return new CloseableHttpResponse(response, null);
             }
-            ResponseEntityProxy.enchance(response, execRuntime);
+            ResponseEntityProxy.enhance(response, execRuntime);
             return new CloseableHttpResponse(response, execRuntime);
         } catch (final ConnectionShutdownException ex) {
             final InterruptedIOException ioex = new InterruptedIOException(
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MinimalHttpClient.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MinimalHttpClient.java
@@ -160,7 +160,7 @@ public class MinimalHttpClient extends CloseableHttpClient {
                 execRuntime.releaseEndpoint();
                 return new CloseableHttpResponse(response, null);
             }
-            ResponseEntityProxy.enchance(response, execRuntime);
+            ResponseEntityProxy.enhance(response, execRuntime);
             return new CloseableHttpResponse(response, execRuntime);
         } catch (final ConnectionShutdownException ex) {
             final InterruptedIOException ioex = new InterruptedIOException("Connection has been shut down");
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ResponseEntityProxy.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ResponseEntityProxy.java
@@ -31,10 +31,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.SocketException;
+import java.util.Arrays;
+import java.util.List;
 
 import org.apache.hc.client5.http.classic.ExecRuntime;
+import org.apache.hc.core5.function.Supplier;
 import org.apache.hc.core5.http.ClassicHttpResponse;
+import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.impl.io.ChunkedInputStream;
 import org.apache.hc.core5.http.io.EofSensorInputStream;
 import org.apache.hc.core5.http.io.EofSensorWatcher;
 import org.apache.hc.core5.http.io.entity.HttpEntityWrapper;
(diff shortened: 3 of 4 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ChunkAwareResponseEntityProxy.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    ChunkAwareResponseEntityProxy(final HttpEntity entity, final ExecRuntime execRuntime, final ChunkedInputStream chunkedInputStream) {
+        super(entity, execRuntime);
+        this.internalChunkedInputStream = chunkedInputStream;
+    }
+
+    @Override
+    public Supplier<List<? extends Header>> getTrailers() {
```
ok2c: @serkanturgut Could not the `getTrailer` method be added to the `ResponseEntityProxy` eliminating the need for `ChunkAwareResponseEntityProxy`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ChunkAwareResponseEntityProxy.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.impl.io.ChunkedInputStream;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+class ChunkAwareResponseEntityProxy extends ResponseEntityProxy {
+
+    private ChunkedInputStream internalChunkedInputStream;
```
ok2c: @serkanturgut This instance variable can be made final.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ChunkAwareResponseEntityProxy.java
serkanturgut: getTrailers() return a supplier which reaches to ChunkedInputStream.getFooters(). So getTrailers() in ChunkAwareResponseEntityProxy is aware of the underlying stream is ChunkedInputStream. 

If we embed getTrailers() into ResponseEntityProxy, we have to check whether the underlying stream is ChunkedInputStream (ChunkAwareResponseEntityProxy is doing this check during initialization) if it is not a IllegalStateException should be thrown there to notify caller. However this approach contradicts with throw-early-catch-late practice. Given that getTrailers() (or Supplier.get()) typically will be called at the end of HTTP communication by the user of client, It would be a unpleasant surprise to caller to see an exception 

Another reason is that I wanted to leave ResponseEntityProxy stream agnostic as it is currently. Placing new getTrailers() implementation in ResponseEntityProxy brings certain responsibility to ResponseEntityProxy to recognize the type of underlying stream to some extent. 

If you think this is still worth it, I could give it a try to eliminate it. I am open to suggestions, please let me know what you think.

Thank you

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ChunkAwareResponseEntityProxy.java
serkanturgut: Missed this one. I will fix this in next revision.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ChunkAwareResponseEntityProxy.java
ok2c: @serkanturgut I personally see no problem if `ResponseEntityProxy` returned `null` from `getTrailers` if for whatever reason the underlying content stream is not `ChunkedInputStream`. This approach would simplify things considerably. I see no great value in throwing an unchecked exception in the middle of an HTTP message exchange.

serkanturgut: I added ``getTrailers()`` method into ``ResponseEntityProxy``, removed the if statement in ``MainClientExec`` and did not need another static initializer for ``ResponseEntityProxy``.

The following will be the only addition to the code which will be in ``ResponseEntityProxy``, all other changes in this commit is eliminated now;

```
   @Override
    public Supplier<List<? extends Header>> getTrailers() {
            return new Supplier<List<? extends Header>>() {
                @Override
                public List<? extends Header> get() {
                    final Header[] footers;
                    InputStream underlyingStream = null;
                    try {
                        underlyingStream = getContent();
                    } catch (IOException e) {
                        throw new IllegalStateException("Unable to retrieve input stream");
                    }
                    if (underlyingStream instanceof ChunkedInputStream) {
                        ChunkedInputStream chunkedInputStream = (ChunkedInputStream) underlyingStream;
                        footers = chunkedInputStream.getFooters();
                    } else {
                        footers = new Header[0];
                    }
                    return Arrays.asList(footers);
                }
            };
    }
```

I can't say it looks cleaner however it reduces the amount of changes and need for new class considerably as you mentioned. If this looks good, I will change the unit tests accordingly and push the commit.

ok2c: @serkanturgut As far as I can say simpler is _better_. Looks good to me.

ok2c:
@serkanturgut Looks very good to me. Could you please squash all changes into a single commit though?

serkanturgut:
> @serkanturgut Looks very good to me. Could you please squash all changes into a single commit though?

Squashed and updated the PR. Thank you @ok2c

ok2c:
@serkanturgut Many thanks for contributing this change-set.

PR #157, opened 2019-07-27
Make Accept-Encoding header handling thread-safe

coxlinton:


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ContentCompressionExec.java
```diff
@@ -70,16 +71,20 @@
...
     private final Lookup<InputStreamFactory> decoderRegistry;
     private final boolean ignoreUnknown;
 
     public ContentCompressionExec(
             final List<String> acceptEncoding,
             final Lookup<InputStreamFactory> decoderRegistry,
             final boolean ignoreUnknown) {
-        this.acceptEncoding = acceptEncoding != null ? acceptEncoding.toArray(
-                new String[acceptEncoding.size()]) : new String[] {"gzip", "x-gzip", "deflate"};
+        // MessageSupport.format is not thread-safe, because it sorts the
```
ok2c: @coxlinton I will fix `MessageSupport#format` in HttpCore master. Do you mind dropping the comment about `MessageSupport#format` thread safety from this change-set?

coxlinton: Not at all - that definitely seems like an excellent idea. Updated made.

ok2c: @coxlinton Could you please squash those two commits into a single one?

coxlinton: Sorry about that - done.

PR #160, opened 2019-08-17
Execute Socket#connect under doPrivileged

s1monw:
In order to allow users to run under a security manager that only grants
connect permission to the httpcore codebase the connect methods should
be executed in a doPriveledged block.
This is certainly not the only issue that users run into when they
install a SecurityManager with strict permissions but certainly the
most prominent and most likely one. Upstream components like the client
might also need to protect places accessing the proxy selector etc.


s1monw:
Relates to https://github.com/apache/httpcomponents-core/pull/135

PR #161, opened 2019-08-21
Add RequestConfig builder javadocs

rhernandez35:
I linked each builder method to its corresponding getter in the built object. However, the builder methods are the primary interface for users, so I think it should hold the primary javadocs instead.


ok2c:
@rhernandez35 Thank you for contributing this PR. It would really cool if you could make similar changes to the config classes in HttpCore.

rschmitt:
@ok2c Do you agree with his suggestion that the Javadoc should be moved to the builder methods? I like that idea.

ok2c:
@rschmitt I am fine with it as long as the same change get applied consistently to all similar classes in core and client.

PR #162, opened 2019-08-22
Modify RequestConfig Javadoc

bekker:
Cleaned up duplicate sentences and removed 'system default' from getSocketTimeout comment ([HTTPCLIENT-1765](https://issues.apache.org/jira/browse/HTTPCLIENT-1765)).

### DIFF
--- a/httpclient/src/main/java/org/apache/http/client/config/RequestConfig.java
+++ b/httpclient/src/main/java/org/apache/http/client/config/RequestConfig.java
@@ -264,11 +264,10 @@ public class RequestConfig implements Cloneable {
 
     /**
      * Returns the timeout in milliseconds used when requesting a connection
-     * from the connection manager. A timeout value of zero is interpreted
-     * as an infinite timeout.
+     * from the connection manager.
      * <p>
      * A timeout value of zero is interpreted as an infinite timeout.
-     * A negative value is interpreted as undefined (system default).
+     * A negative value is interpreted as undefined (system default if applicable).
      * </p>
      * <p>
      * Default: {@code -1}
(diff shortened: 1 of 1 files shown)


httpclient/src/main/java/org/apache/http/client/config/RequestConfig.java
```diff
@@ -295,11 +293,11 @@ public int getConnectTimeout() {
 
     /**
      * Defines the socket timeout ({@code SO_TIMEOUT}) in milliseconds,
-     * which is the timeout for waiting for data  or, put differently,
+     * which is the timeout for waiting for data or, put differently,
      * a maximum period inactivity between two consecutive data packets).
      * <p>
      * A timeout value of zero is interpreted as an infinite timeout.
-     * A negative value is interpreted as undefined (system default).
+     * A negative value is interpreted as undefined.
```
ok2c: @bekker Thank you for cleaning up the javadocs but why do you want to remove `(system default)` from socket timeout description and leave it everywhere else?

bekker: I was not so sure about other timeouts. Would it be better to make it same for other timeouts?

garydgregory: I think we should keep the docs as is or expand on it, as opposed to removing information.

bekker: @garydgregory You're right. I should add more info, not removing it.

But I'm not sure about meaning of 'undefined' here - is it 'undefined behavior'? Then there should be a warning about unexpected results.

bekker:
OK, it seems inappropriate to fix timeout comments right now.

I reverted timeout default comments and only javadoc cleanup left.

ok2c:
@bekker We could phrase it like ` A negative value is interpreted as undefined (system default if applicable)` if that makes it sound better to you.
Could you please also squash your commits into a single one?

PR #166, opened 2019-09-26
Enforce h2 TLS rules after negotiating TLS, not before

rschmitt:


httpclient5/src/main/java/org/apache/hc/client5/http/ssl/AbstractClientTlsStrategy.java
```diff
@@ -130,7 +131,15 @@ public void initialize(final NamedEndpoint endpoint, final SSLEngine sslEngine)
             @Override
             public TlsDetails verify(final NamedEndpoint endpoint, final SSLEngine sslEngine) throws SSLException {
                 verifySession(host.getHostName(), sslEngine.getSession());
-                return createTlsDetails(sslEngine);
+                final TlsDetails tlsDetails = createTlsDetails(sslEngine);
+                final String negotiatedCipherSuite = sslEngine.getSession().getCipherSuite();
+                if (tlsDetails != null && "h2".equals(tlsDetails.getApplicationProtocol())) {
```
ok2c: There is `ApplicationProtocols` enum for supported application protocols. It is used by all application protocol negotiators in core. You might want it to use here as well.

rschmitt: OK

PR #174, opened 2019-11-22
HTTPCLIENT-2030: Fix PublicSuffixMatcher::getDomainRoot on invalid hostnames

nielsbasjes:
This fixes the behaviour of PublicSuffixMatcher::getDomainRoot in the case when you feed it a hostname for which NO rule has been defined; and you specify an expected domaintype (i.e. either ICANN or PRIVATE).

Notable changes for reviewers:
- PublicSuffixMatcher load(final InputStream in)  is now public to make testing a lot easier and also useful for downstream users.
- I had to change the test for HTTPCLIENT-1997 also because that used an invalid domain name in their test. So instead of "cloud.a" it now uses "cloud.com" instead.

### DIFF
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
@@ -30,6 +30,7 @@ package org.apache.hc.client5.http.psl;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 
 import org.junit.Assert;
 import org.junit.Before;
@@ -37,7 +38,7 @@ import org.junit.Test;
 
 public class TestPublicSuffixMatcher {
 
-    private static final String SOURCE_FILE = "suffixlist.txt";
+    private static final String SOURCE_FILE = "suffixlistmatcher.txt";
 
     private PublicSuffixMatcher matcher;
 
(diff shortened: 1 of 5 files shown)


httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
```diff
@@ -46,36 +44,78 @@ public void setUp() throws Exception {
...
+        // Private
         Assert.assertEquals("example.xx", matcher.getDomainRoot("example.XX"));
         Assert.assertEquals("example.xx", matcher.getDomainRoot("www.example.XX"));
         Assert.assertEquals("example.xx", matcher.getDomainRoot("www.blah.blah.example.XX"));
-        Assert.assertEquals(null, matcher.getDomainRoot("xx"));
-        Assert.assertEquals(null, matcher.getDomainRoot("jp"));
-        Assert.assertEquals(null, matcher.getDomainRoot("ac.jp"));
-        Assert.assertEquals(null, matcher.getDomainRoot("any.tokyo.jp"));
+        // Too short
+        Assert.assertNull(matcher.getDomainRoot("xx"));
```
ok2c: @nielsbasjes I would prefer to keep non-functional changes in a separate commit.

nielsbasjes:
Understood. I'll put up an updated pull request soon.

httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcherLoader.java
```diff
@@ -50,7 +50,8 @@
 @Contract(threading = ThreadingBehavior.SAFE)
 public final class PublicSuffixMatcherLoader {
 
-    private static PublicSuffixMatcher load(final InputStream in) throws IOException {
+    public static PublicSuffixMatcher load(final InputStream in) throws IOException {
```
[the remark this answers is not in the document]
nielsbasjes: I reverted this and used a different solution.

nielsbasjes:
I have added an optional DomainType parameter to the DefaultHostnameVerifier in case a PublicSuffixMatcher is specified. The default value is null if not specified because that matches the behavior as it currently is.
I handled the review comments.
Did a rebase against the latest master.
@ok2c Please review.

httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
nielsbasjes: I reverted this change to assertNull

PR #176, opened 2019-11-26
HTTPCLIENT-2030: Fix PublicSuffixMatcher::getDomainRoot on invalid hostnames (4.5.x)

nielsbasjes:
This is an as exact as possible backport of https://github.com/apache/httpcomponents-client/pull/174 to the 4.5.x branch.
Only real differences are file locations and the UTF_8 had to be defined manually.


nielsbasjes:
@ok2c I've backported it to 4.5.x

PR #181, opened 2019-12-10
Add reactive test coverage

rschmitt:
This commit adds test coverage for the `:httpcore5-reactive` bindings,
using different types of clients and protocols. Not all of these tests
pass, as some of them have uncovered new bugs. In particular:

* `testSequentialHeadRequests` always fails and is currently ignored
* The `testSequentialPostRequests` and `testConcurrentPostRequests` test
cases fail non-deterministically

### DIFF
--- a/httpclient5-testing/pom.xml
+++ b/httpclient5-testing/pom.xml
@@ -43,6 +43,11 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.httpcomponents.core5</groupId>
+      <artifactId>httpcore5-reactive</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.httpcomponents.client5</groupId>
       <artifactId>httpclient5</artifactId>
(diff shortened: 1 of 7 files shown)


ok2c:
@rschmitt Awesome! We could also keep this change-set on a separate branch as an alternative to disabling unstable tests.

garydgregory:
I think we should keep it in master and JUnit assumes to disable-enable them.

rschmitt:
I think the `@Ignore` annotation makes the most sense here, especially since these tests should only be disabled in the short-term. (As I mentioned, I've already fixed some of them.) I'm afraid that keeping them on a separate branch will cause bitrot and compilation failures after we rename everything again. I also don't see how assumptions make sense here, since you normally use them to check some sort of precondition about the test *environment* (e.g. the JVM version, availability of credentials for integration tests, the operating system), not about the code being tested and whether it is buggy. (Determining whether the code is buggy is why we have the tests!)

ok2c:
@rschmitt Fair enough. Feel free to merge the tests whenever you deem them ready.

PR #192, opened 2019-12-29
HTTPCORE-615: Implement HTTP-based cache serializer-deserializer.

scottgifford:
See https://issues.apache.org/jira/browse/HTTPCORE-615


ok2c:
@scottgifford Looks OK to me. I will merge the PR if there are no objections. 

What I do not understand is why we need `HttpByteArrayCacheEntrySerializerTestUtils`. Could you please fold it into `TestHttpByteArrayCacheEntrySerializer`?

scottgifford:
> @scottgifford Looks OK to me. I will merge the PR if there are no objections.
> 
> What I do not understand is why we need `HttpByteArrayCacheEntrySerializerTestUtils`. Could you please fold it into `TestHttpByteArrayCacheEntrySerializer`?

OK!  I re-read the PR this morning and pushed a few small cleanups, they can be squashed on merge, or I can push a new branch and squash them there (I don't know of a way to squash them myself without creating a new branch).

My thoughts with putting the HttpByteArrayCacheEntrySerializerTestUtils into its own file were twofold.  First, there is quite a bit of test infrastructure with this, and putting it into a separate file made it easy to read the tests to see what's being tested, and the utilities to see how the tests are implemented; if I move the test utilities into the tests, it will be almost 800 lines, which is a pretty big file to  grok IMO.  Second my benchmark program uses the test utilities, so it's convenient to have them in a separate file so a developer can drop in the benchmark when they want to think about the performance of this.  All that said, this is only a mild preference for me, and if you would like the files merged together I will happily make the change.

PR #198, opened 2020-01-06
DefaultHostnameVerifier: Match DNS and CN names against ICANN domains

rschmitt:
This change ensures that during hostname verification the public suffix
list is only used to prevent wildcard matching against entire TLDs (e.g.
`*.com`). Currently, private domains are also being matched against,
which is preventing reasonable wildcards (such as
`*.s3.eu-central-1.amazonaws.com`) from being respected.


ok2c:
@rschmitt It looks like a regression introduced by #174

PR #205, opened 2020-01-28
Now that ClassicHttpRequests is no longer an enum, we need to way to

garydgregory:
generically build requests from method names. Update all factory classes
with matching APIs for Method and String method name inputs.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/BasicHttpRequests.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/BasicHttpRequests.java
@@ -29,11 +29,11 @@ package org.apache.hc.client5.http.async.methods;
 
 import java.net.URI;
 import java.util.Locale;
-import java.util.Objects;
 
 import org.apache.hc.core5.http.HttpHost;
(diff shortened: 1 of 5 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/BasicHttpRequests.java
```diff
@@ -40,6 +42,38 @@
  */
 public final class BasicHttpRequests {
 
+    // TODO Next version of HttpCore:
+    // Method.normalizedValueOf(method)
+    private static Method normalizedValueOf(final String method) {
+        return Method.valueOf(Objects.requireNonNull(method, "method").toUpperCase(Locale.ROOT));
```
ok2c: @garydgregory Please use `Args.notNull` for consistency with the rest of the code base.

garydgregory: Done.

ok2c:
@garydgregory Looks good to me. Please add the release notes with the same commit if you have to, but please do not do that as a separate commit.

PR #210, opened 2020-02-12
re-add validateAfterInactivity to pool manager

h0nIg:
It is really ugly that this setting was removed with https://github.com/apache/httpcomponents-client/commit/e48091da3d945b056d9ad9ea46a62b91e142d33f#diff-66a448bdca2ba2ef9a7c3f8d93784078L187. This default value is responsible for checking bad connections and people will not notice this change if they upgrade, because it will not happen very frequently: https://github.com/apache/httpcomponents-client/blob/rel/v4.5.9/httpclient/src/main/java/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.java#L182

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -187,6 +187,7 @@ public class PoolingHttpClientConnectionManager
             final TimeValue timeToLive,
             final HttpConnectionFactory<ManagedHttpClientConnection> connFactory) {
         super();
+        this.setValidateAfterInactivity(2000);
         this.connectionOperator = Args.notNull(httpClientConnectionOperator, "Connection operator");
         switch (poolConcurrencyPolicy != null ? poolConcurrencyPolicy : PoolConcurrencyPolicy.STRICT) {
             case STRICT:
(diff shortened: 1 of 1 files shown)


ok2c:
@h0nIg Why? What is exactly the benefit of calling that method in the constructor?

h0nIg:
@ok2c to restore the original 4.x behaviour as outlined in my comment. Please take a look at the links how it looked before

ok2c:
> @ok2c to restore the original 4.x behaviour as outlined in my comment. Please take a look at the links how it looked before

@h0nIg And why would we do that exactly?

h0nIg:
@ok2c because the stale check will not get triggered if this value is empty. In 4.X it will get triggered and i don't see any reason why this should change: https://github.com/apache/httpcomponents-client/blob/bd22d7cf24f262fcbf5040ac50e262d1ca1b9974/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java#L291-L308

ok2c:
@h0nIg The check _will_ get triggered if configured correctly.

h0nIg:
@h0nIg "if configured properly". In 4.X it was enabled by default and in 5.X it is not enabled by default, therefore please enable it by default in 5.X as well or help me to understand why its no longer like in 4.X

ok2c:
@h0nIg The stale connection check is both ineffective and inefficient. A major release is the point when its application can be changed from having to opt-out to having to opt-in.

h0nIg:
@ok2c ok, can you elaborate about "ineffective and inefficient" to better understand what is not catched by the check?

ok2c:
@h0nIg The stale check is ineffective because the connection can go stale immediately after having passed the check. It can catch some stale connections but not all. It is also inefficient by being expensive in therms of execution overhead.

ok2c:
> @ok2c ok after looking at the code it is still not clear for me why you said it is inefficient. 

@h0nIg Because in reality socket timeout granularity of many JREs is about 10 ms or more. Therefore the stale connection check adds ~10 ms to each request execution time when activated.

h0nIg:
@ok2c i stumbled across this issue once again and i think i need more details. to be very specific: the 10ms will get added to all requests or just to the requests which are stale and therefore these 10ms will get added until the JRE will see that the connection is stale?

ok2c:
@h0nIg All requests.

PR #226, opened 2020-06-08
Disable SNI extension for a specific SSLConnectionSocketFactory

nrv:
Some website have badly configured their SSL/TLS and it is not correctly managed by java, thus having an exception instead of  simple warning : "javax.net.ssl.SSLHandshakeException: received handshake warning: unrecognized_name". This is a know problem for java > 8

The solution to bypass this problem is to disable the SNI extension but it could only be done at the JVM level with a -D option of programmatically with System.setProperty("jsse.enableSNIExtension", "false")

This patch allows to disable the SNI verification for a single SSLConnectionSocketFactory, as explained here : https://stackoverflow.com/questions/55903477/received-handshake-warning-unrecognized-name


ok2c:
@nrv The trouble is that HC 5.0 still requires Java 1.7 compatible runtime. We cannot use 1.8 specific code in HttpClient 5.0 yet. We will likely upgrade to Java 1.8 in the 5.1 or 5.2 release.

PR #227, opened 2020-06-09
HTTPCLIENT-2083: Fix NPE when classic client interceptors are added

carterkozak:
Previously HttpClientBuilder methods addExecInterceptorFirst and
addExecInterceptorLast resulted in NullPointerExceptions because
the execInterceptors list had not been initialized yet.


garydgregory:
@carterkozak You are missing unit tests ;-)

carterkozak:
That's true. I couldn't find prior art for test coverage across the client builder class, happy to add it :-)

ok2c:
@carterkozak Could you please fix style violations and make sure that the build passes style checks? Otherwise looks good.

PR #228, opened 2020-06-19
HTTPCLIENT-2082

rafiek:
Adding the connection socket lookup as a parameter for the builder makes it more flexible to use this builder.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java
@@ -41,9 +41,9 @@ import org.apache.hc.core5.http.config.Lookup;
 import org.apache.hc.core5.http.config.RegistryBuilder;
 import org.apache.hc.core5.http.io.HttpConnectionFactory;
 import org.apache.hc.core5.http.io.SocketConfig;
-import org.apache.hc.core5.pool.ManagedConnPool;
 import org.apache.hc.core5.pool.PoolConcurrencyPolicy;
 import org.apache.hc.core5.pool.PoolReusePolicy;
+import org.apache.hc.core5.pool.StrictConnPool;
 import org.apache.hc.core5.util.TimeValue;
 
 /**
@@ -80,7 +80,7 @@ public class PoolingHttpClientConnectionManagerBuilder {
     private SchemePortResolver schemePortResolver;
     private DnsResolver dnsResolver;
 
-    private ManagedConnPool<HttpRoute, ManagedHttpClientConnection> managedConnPool;
+    private StrictConnPool<HttpRoute, ManagedHttpClientConnection> managedConnPool;
     private PoolConcurrencyPolicy poolConcurrencyPolicy;
     private PoolReusePolicy poolReusePolicy;
     private SocketConfig defaultSocketConfig;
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java
```diff
@@ -70,10 +74,13 @@
  */
 public class PoolingHttpClientConnectionManagerBuilder {
 
+    private Lookup<ConnectionSocketFactory> connectionSocketFactoryRegistry;
     private HttpConnectionFactory<ManagedHttpClientConnection> connectionFactory;
     private LayeredConnectionSocketFactory sslSocketFactory;
     private SchemePortResolver schemePortResolver;
     private DnsResolver dnsResolver;
+
+    private StrictConnPool<HttpRoute, ManagedHttpClientConnection> managedConnPool;
```
ok2c, line 83: @rafiek Why do we need this variable? It invalidates `PoolConcurrencyPolicy` parameter.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java
```diff
@@ -70,10 +74,13 @@
  */
 public class PoolingHttpClientConnectionManagerBuilder {
 
+    private Lookup<ConnectionSocketFactory> connectionSocketFactoryRegistry;
```
ok2c, line 77: @rafiek Same problem. This variable invalidates `sslSocketFactory`. What does it get us?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java
rafiek: It is used in tests, so I added the option to make it part of the `build()`. When omitted, you can still use `PoolConcurrencyPolicy`.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java
rafiek: When omitting the use of this variable for `'build()`, you can still use `sslSocketFactory` as part of `build()`.

ok2c:
@rafiek I understand that. However we have a very similar situation with `HttpClientBuilder` in HC 4.x and it has been causing quite a bit of confusion and troubles with users who never get around to reading the API documentation. I do not want to repeat the same situation in HC 5.x.

rafiek:
> @rafiek I understand that. However we have a very similar situation with `HttpClientBuilder` in HC 4.x and it has been causing quite a bit of confusion and troubles with users who never get around to reading the API documentation. I do not want to repeat the same situation in HC 5.x.

Aah, did not know that. Then how about this. I was preparing this feature to make it easier to implement a programmatic SOCKS based HttpClient. The `HttpClientBuilder` uses the `PoolingHttpClientConnectionManagerBuilder` to build a default HttpClientConnectionManager, but doesn't take into consideration to use `systemProperties`. When it does, then it would be very convenient that the builder takes in `connectionSocketFactoryRegistry` as a step for `build()`. This was part of [HTTPCLIENT-1966](https://issues.apache.org/jira/browse/HTTPCLIENT-1966), but I closed the pull request because it was preferred to implement it in HC 5.x.

Should I also add the changes for HTTPCLIENT-1966 in this pull request. Making the scope bigger and the intention for these changes more clear. Maybe adding more javadoc also?

ok2c:
@rafiek I initially thought you were going to add support for SOCKS related system properties but with this PR you have veered into an unexpected direction.

rafiek:
> @rafiek I initially thought you were going to add support for SOCKS related system properties but with this PR you have veered into an unexpected direction.

No worries. Hang tight. I'll steer this boat back into familiar waters. I'll contact you in Jira when I've updated this pull request.

PR #229, opened 2020-06-23
HTTPCLIENT-2090: TLS Upgrade connections use the configured socket timeout

carterkozak:
Previously upgrading to tls resulted in the connect timeout being
used instead of the configured socket timeout.

I've pulled in a server library I'm familiar with to facilitate testing, but I'd be happy to use another library if you can point me toward an example.


ok2c:
Superseded by 445b61a01afed7b47926a3977cd58937be1831bf

carterkozak:
Thanks @ok2c!

PR #231, opened 2020-07-01
HTTPCLIENT-2094: ConnectionManager validateAfterInactivity zero duration agreement

carterkozak:
Both connection managers agree on the meaning of a zero-duration
validateAfterInactivity value. Previously the documentation for
both suggested that zero resulted in no validation, however
the classic client would validate prior to each request and
the async client would never validate.
This commit standardizes behavior on the classic client, which
is consistent with hc4.x, allowing zero to force validation prior
to every request.


ok2c:
Awesome! Thank you, @carterkozak

PR #233, opened 2020-07-03
Optimize MultipartEntityBuilder#generateBoundary

slisaasquatch:
I've changed generateBoundary to use `ThreadLocalRandom` instead of creating a new `Random` every time, and I've replaced the `StringBuilder` with the faster `CharBuffer`.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java
@@ -29,6 +29,7 @@ package org.apache.hc.client5.http.entity.mime;
 
 import java.io.File;
 import java.io.InputStream;
+import java.nio.CharBuffer;
 import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


ok2c:
@michael-o `ThreadLocal`s must be used with _extreme_ caution and generally should have a very specific, short life span, like duration of an HTTP request / response exchange in a Servlet environment. I need to take a very close look at `ThreadLocalRandom` before I can make up my mind. The rest looks just fine.

slisaasquatch:
@michael-o No, I don't have any numbers. I just thought it was an easy improvement to make.
@ok2c `ThreadLocalRandom` is not really a regular `ThreadLocal` that's used to store a temporary state. The purpose of `ThreadLocalRandom` is to avoid synchronization. Basically all the `new Random()` calls can be safely replaced with `ThreadLocalRandom.current()` without negative side effects. [Here's the official documentation](https://docs.oracle.com/javase/tutorial/essential/concurrency/threadlocalrandom.html).

ok2c:
@slisaasquatch Fair enough. Merged.

PR #237, opened 2020-07-06
PoolingAsyncClientConnectionManager handles null ping callback values

carterkozak:
This change brings back handling for null values that was lost
in the fix for HTTPCLIENT-2097.
Note that unlike the previous implementation, a null value
results in the connection being discarded and closed as opposed to
being added to the pool in an unknown state.


ok2c:
@carterkozak Superseded by 93e3d0f45204836a547602c56a51f0c32f0f3df4.

PR #241, opened 2020-07-11
HTTPCLIENT-2099: SSLConnectionSocketFactory timeout fix

carterkozak:
SSLConnectionSocketFactory no longer overrides the socket timeout
with the connect timeout when an unlimited socket timeout is
configured. This matches behavior of HTTPCLIENT-2091.

Note that in scenarios where SocketConfig sets an infinite timeout
and the RequestConfig sets a bounded timeout, this change results
in the connect-timeout no longer applying to the TLS handshake.
This behavior can be retained by setting the expected timeout in
the SocketConfig.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java
@@ -207,9 +207,6 @@ public class SSLConnectionSocketFactory implements LayeredConnectionSocketFactor
             sock.bind(localAddress);
         }
         try {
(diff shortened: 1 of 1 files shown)


ok2c:
@carterkozak Damn, the pull request was against master, not the 5.1.x branch. I will have to move it to the 5.1.x branch.

carterkozak:
oops, my fault, sorry about that. I can set up a revert pr and a second change if you like.

ok2c:
@carterkozak No, worries. I will fix it.

ok2c:
@carterkozak I apologize for causing such a confusion. I was just being old and stupid. HTTPCLIENT-2091 fix committed to master and not 5.1.x. So, we are all good.

carterkozak:
Not a problem, thanks again for your help!

PR #243, opened 2020-08-04
Implement an adapter from unmanaged to managed client connections

carterkozak:
Draft based on discussion in https://github.com/apache/httpcomponents-core/pull/206#issuecomment-668594741
If this seems helpful I will file a ticket and write tests.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/HttpClientConnectionFactoryManagedAdapter.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/HttpClientConnectionFactoryManagedAdapter.java
@@ -1,0 +1,215 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/HttpClientConnectionFactoryManagedAdapter.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Override
+    public String toString() {
+        return "HttpClientConnectionFactoryManagedAdapter{" + delegate + '}';
+    }
+
+    private static final class ManagedHttpClientConnectionAdapter implements ManagedHttpClientConnection {
+
+        private final HttpClientConnection delegate;
+        private final AtomicBoolean closed = new AtomicBoolean();
```
carterkozak, line 70: I'm not sure close tracking is necessary here, but I borrowed it from the DefaultManagedHttpClientConnection implementation.

ok2c:
@carterkozak Is this PR still relevant? Could you please re-base it off the latest master?

carterkozak:
@ok2c I'm not sure this is a helpful utility to provide since we define `DefaultManagedHttpClientConnection` to adapt the httpcomponents-core factory into a managed connection factory, but it was useful to validate new changes from core without a direct dependency from httpcomponents-client onto the snapshot of core. If there are other custom non-managed connection factories, this could be helpful.

I suppose I am leaning towards closing this out, but if you think it would be helpful I'd be more than happy to give it a jira issue and merge it. What do you think?

ok2c:
@carterkozak Let's just close it. We can always re-visit it at a later point.

carterkozak:
Sounds good to me, closed.

PR #244, opened 2020-08-04
HTTPCLIENT-2103: ManagedHttpClientConnectionFactory provides a fluent builder

carterkozak:


httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/ManagedHttpClientConnectionFactory.java
```diff
@@ -139,4 +139,71 @@ public ManagedHttpClientConnection createConnection(final Socket socket) throws
...
+            this.charCodingConfig = charCodingConfig;
+            return this;
+        }
+
+        public Builder incomingContentLengthStrategy(final ContentLengthStrategy incomingContentStrategy) {
+            this.incomingContentStrategy = incomingContentStrategy;
+            return this;
+        }
+
+        public Builder outgoingContentLengthStrategy(final ContentLengthStrategy outgoingContentStrategy) {
```
carterkozak, line 182: `incomingContentLengthStrategy` and `outgoingContentLengthStrategy` slightly differ from the names used in core, which uses `incomingContentStrategy` `outgoingContentStrategy`. I think the specificity is helpful, so I'll PR into core to update names there as well since 5.1 has not yet been released.

PR #257, opened 2020-10-05
4.5.x - Prevent sending Authorization header to other than the original host

marverix:
It's an easy fix and really important (from my perspective) - mainly for security reasons.


marverix:
May I know why?

ok2c:
@marverix What exactly do you mean? Why the PR was closed? Same reason as #258

PR #261, opened 2020-10-25
[HTTPCLIENT-2124] NullPointerException in MinimalHttpClient.doExecute(HttpHost, ClassicHttpRequest, HttpContext)

garydgregory:
`NullPointerException` in `org.apache.hc.client5.http.impl.classic.MinimalHttpClient.doExecute(HttpHost, ClassicHttpRequest, HttpContext)`

### DIFF
--- a/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestMinimalClientRequestExecution.java
+++ b/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestMinimalClientRequestExecution.java
@@ -108,9 +108,10 @@ public class TestMinimalClientRequestExecution extends LocalServerTestBase {
 
         for (int i = 0; i < 10; i++) {
             final HttpGet request = new HttpGet("/");
-            final CloseableHttpResponse response = this.httpclient.execute(target, request);
-            EntityUtils.consume(response.getEntity());
-            Assert.assertEquals(HttpStatus.SC_OK, response.getCode());
+            try (final CloseableHttpResponse response = this.httpclient.execute(target, request)) {
+                EntityUtils.consume(response.getEntity());
+                Assert.assertEquals(HttpStatus.SC_OK, response.getCode());
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
Looks good to me, thanks!

michael-o (commented):
Where is the NPE and why is there no try block in the second test?

carterkozak:
context is nullable where clientContext is not. The adaptClientContext line produces the clientContext from a nullable context.

carterkozak:
As an aside, we’ve used https://github.com/uber/NullAway at work to avoid NPEs with compile time checks, but it can be painful to retrofit into large existing projects.

garydgregory:
On Sun, Oct 25, 2020 at 1:36 PM Carter Kozak <notifications@github.com>
wrote:

> As an aside, we’ve used https://github.com/uber/NullAway at work to avoid
> NPEs with compile time checks, but it can be painful to retrofit into large
> existing projects.
>

I found this one using SpotBugs which you can run as a Maven or Eclipse
plugin.

Gary

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestMinimalClientRequestExecution.java
```diff
@@ -98,4 +100,18 @@ public void testNonCompliantURI() throws Exception {
...
 
+    @Test
+    public void testNonCompliantURIWithoutContext() throws Exception {
+        this.server.registerHandler("*", new SimpleService());
+        this.httpclient = HttpClients.createMinimal();
+        final HttpHost target = start();
+
+        for (int i = 0; i < 10; i++) {
+            final HttpGet request = new HttpGet("/");
+            final CloseableHttpResponse response = this.httpclient.execute(target, request);
```
michael-o, line 111: Why is there no try with resources block?

garydgregory: Fixed! :-)

michael-o (approved):

PR #263, opened 2020-10-30
Added a security policy

artem-smotrakov:
The process of reporting security issues should be documented and easy to find.

I'd like to propose adding a security policy to the project that would be integrated with other GitHub features

https://docs.github.com/en/free-pro-team@latest/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository

Although the project's page has the "Security" tab, I think `SECURITY.md` would make it a bit easier to find the docs that describe how security issues should be reported. Also, when someone creates an issue in the repository, they will see a link to the security policy.

The proposed `SECURITY.md` file just contains a link to https://www.apache.org/security/ but we can put more info if necessary.

I had a look at several Apache projects and looks like there is a common template for `SECURITY.md`. For example:

- https://github.com/apache/nifi/blob/main/SECURITY.md
- https://github.com/apache/logging-log4j2/blob/master/SECURITY.md
- https://github.com/apache/commons-text/blob/master/SECURITY.md

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


michael-o (approved):
Makes sense. Are you going to wire it with another PR?

artem-smotrakov:
> Makes sense. Are you going to wire it with another PR?

This PR is standalone. Or, maybe I didn't get what you mean. I can also open a similar PR for https://github.com/apache/httpcomponents-core

garydgregory:
I added a similar page to most Apache Commons Components, but the page there is a more generic pointer because this is a policy page: https://github.com/apache/commons-io/security/policy

ok2c:
@artem-smotrakov Looks good to me. Please do add ASLv2 header to the file in order to make Apache RAT happy. Refer to `README.md` as an example.

ok2c (approved):

PR #272, opened 2020-12-08
Add interceptors before MAIN_TRANSPORT so they won't be ignored

robtimus:
When `addExecInterceptorLast` is called on an `HttpClientBuilder`, `HttpAsyncClientBuilder` or `H2AsyncClientBuilder`, the given executor is added as last element of the executor chain. That may seem correct, but it causes them to be ignored. The reason is the `*MainClientExec` that's initially the last element. In all three cases, this does not delegate to the chain. As a result, anything added with `addExecInterceptorLast` is currently completely ignored.

This change adds the interceptor as last in the chain of actual interceptors, just before the `*MainClientExec` which I don't consider an interceptor. It's the same as the workaround I've been applying right now which works just fine.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/H2AsyncClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/H2AsyncClientBuilder.java
@@ -752,7 +752,9 @@ public class H2AsyncClientBuilder {
                         execChainDefinition.addFirst(entry.interceptor, entry.name);
                         break;
                     case LAST:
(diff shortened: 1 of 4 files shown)


robtimus:
I'll try and add some unit/integration tests like I did for https://github.com/apache/httpcomponents-core/pull/236

robtimus:
I managed to write one for `HttpClientBuilder`: `TestHttpClientBuilderInterceptors`. For `HttpAsyncClientBuilder` and `H2AsyncClientBuilder` setting up the test server is always tricky, so I won't be able to provide tests for those at this moment.

PR #281, opened 2021-02-05
Minor Improvement:

arturobernalg:
* Avoid duplicate code
* Simplify if else
* Inline variables


ok2c:
@arturobernalg If you really want to log those exceptions should they be logged at the ERROR priority?

arturobernalg:
> @arturobernalg If you really want to log those exceptions should they be logged at the ERROR priority?

Hi @ok2c 

IMO We should inform in somewhere that's an error occurs. What about a warning?

ok2c:
@arturobernalg I would even use DEBUG priority. Could you please take `Add log Catch` out of this patch and put it into a separate PR for now?

arturobernalg:
> @arturobernalg I would even use DEBUG priority. Could you please take `Add log Catch` out of this patch and put it into a separate PR for now?

HI @ok2c 

agree. We can talk about the LOG.XX by mail or later. removed from PR

PR #285, opened 2021-02-11

careylam:
1. At the beginning of the negotiate, no token is defined in "WWW-Authenticate: Negotiate".
2. Kerberos expects HTTP.


ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/GGSSchemeBase.java
```diff
@@ -133,9 +131,9 @@ protected GSSManager getManager() {
      * @since 4.4
      */
     protected byte[] generateGSSToken(
-            final byte[] input, final Oid oid, final String serviceName, final String authServer) throws GSSException {
+            final byte[] input, final Oid oid, final String authServer) throws GSSException {
```
ok2c, line 134: @careylam Could you please keep the original method signature for the backward compatibility sake?

ok2c (changes requested):

PR #293, opened 2021-03-15
Blocking connection managers to validate connections by default

ok2c:
@carterkozak @rschmitt Blocking connection managers now validate connections after inactivity of more than 2s by default. The behavior of async connection managers remains unchanged. 

Please review.


carterkozak (approved):
Looks good to me, thanks @ok2c.

rschmitt (approved):

PR #296, opened 2021-03-18
Connection management configuration

ok2c:
@rschmitt @carterkozak @michael-o @garydgregory I would like to propose the following changes to HttpClient 5.x configuration APIs. In essence all configuration parameters related to connection management and initialization have been moved from `RequestConfig` to a new config class. 

This is just a draft at this point. You are welcome to propose alternative ideas, better class names and so on.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
@@ -1,0 +1,205 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.http.config;
+
+import java.util.concurrent.TimeUnit;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.util.TimeValue;
+import org.apache.hc.core5.util.Timeout;
+
+/**
+ * Immutable class encapsulating connection initialization and management settings.
+ *
+ * @since 5.2
+ */
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+public class ConnectionConfig implements Cloneable {
+
+    private static final Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3);
+
+    public static final ConnectionConfig DEFAULT = new Builder().build();
+
+    private final Timeout connectTimeout;
+    private final Timeout socketTimeout;
+    private final TimeValue validateAfterInactivity;
+
+    /**
+     * Intended for CDI compatibility
+     */
+    protected ConnectionConfig() {
+        this(DEFAULT_CONNECT_TIMEOUT, null, null);
+    }
+
+    ConnectionConfig(
+            final Timeout connectTimeout,
+            final Timeout socketTimeout,
+            final TimeValue validateAfterInactivity) {
+        super();
+        this.connectTimeout = connectTimeout;
+        this.socketTimeout = socketTimeout;
+        this.validateAfterInactivity = validateAfterInactivity;
+    }
+
+    /**
+     * @see Builder#setSocketTimeout(Timeout)
+     */
+    public Timeout getSocketTimeout() {
+        return socketTimeout;
+    }
+
+    /**
+     * @see Builder#setConnectTimeout(Timeout)
+     */
+    public Timeout getConnectTimeout() {
+        return connectTimeout;
+    }
+
+    /**
+     * @see Builder#setValidateAfterInactivity(TimeValue)
+     */
+    public TimeValue getValidateAfterInactivity() {
+        return validateAfterInactivity;
+    }
+
+    @Override
+    protected ConnectionConfig clone() throws CloneNotSupportedException {
+        return (ConnectionConfig) super.clone();
+    }
+
+    @Override
+    public String toString() {
+        final StringBuilder builder = new StringBuilder();
+        builder.append("[");
+        builder.append(", connectTimeout=").append(connectTimeout);
+        builder.append(", socketTimeout=").append(socketTimeout);
+        builder.append(", validateAfterInactivity=").append(validateAfterInactivity);
+        builder.append("]");
+        return builder.toString();
+    }
+
+    public static ConnectionConfig.Builder custom() {
+        return new Builder();
+    }
+
+    public static ConnectionConfig.Builder copy(final ConnectionConfig config) {
+        return new Builder()
+                .setConnectTimeout(config.getConnectTimeout())
+                .setSocketTimeout(config.getSocketTimeout())
+                .setValidateAfterInactivity(config.getValidateAfterInactivity());
+    }
+
+    public static class Builder {
+
+        private Timeout socketTimeout;
+        private Timeout connectTimeout;
+        private TimeValue validateAfterInactivity;
+
+        Builder() {
+            super();
+            this.connectTimeout = DEFAULT_CONNECT_TIMEOUT;
+        }
+
+
+        /**
+         * @see #setSocketTimeout(Timeout)
+         */
+        public Builder setSocketTimeout(final int soTimeout, final TimeUnit timeUnit) {
+            this.socketTimeout = Timeout.of(soTimeout, timeUnit);
+            return this;
+        }
+
+        /**
+         * Determines the default socket timeout value for I/O operations.
+         * <p>
+         * Default: {@code null}
+         * </p>
+         *
+         * @return the default socket timeout value for I/O operations.
+         */
+        public Builder setSocketTimeout(final Timeout soTimeout) {
+            this.socketTimeout = soTimeout;
+            return this;
+        }
+
+        /**
+         * Determines the timeout until a new connection is fully established.
+         * This may also include transport security negotiation exchanges
+         * such as {@code SSL} or {@code TLS} protocol negotiation).
+         * <p>
+         * A timeout value of zero is interpreted as an infinite timeout.
+         * </p>
+         * <p>
+         * Default: 3 minutes
+         * </p>
+         */
+        public Builder setConnectTimeout(final Timeout connectTimeout) {
+            this.connectTimeout = connectTimeout;
+            return this;
+        }
+
+        /**
+         * @see #setConnectTimeout(Timeout)
+         */
+        public Builder setConnectTimeout(final long connectTimeout, final TimeUnit timeUnit) {
+            this.connectTimeout = Timeout.of(connectTimeout, timeUnit);
+            return this;
+        }
+
+        /**
+         * Defines period of inactivity after which persistent connections must
+         * be re-validated prior to being leased to the consumer. Negative values passed
+         * to this method disable connection validation.
+         * <p>
+         * Default: {@code null}
+         * </p>
+         */
+        public Builder setValidateAfterInactivity(final TimeValue validateAfterInactivity) {
+            this.validateAfterInactivity = validateAfterInactivity;
+            return this;
+        }
+
+        /**
+         * @see #setValidateAfterInactivity(TimeValue)
+         */
+        public Builder setValidateAfterInactivity(final long validateAfterInactivity, final TimeUnit timeUnit) {
+            this.validateAfterInactivity = TimeValue.of(validateAfterInactivity, timeUnit);
+            return this;
+        }
+
+        public ConnectionConfig build() {
+            return new ConnectionConfig(
+                    connectTimeout != null ? connectTimeout : DEFAULT_CONNECT_TIMEOUT,
+                    socketTimeout,
+                    validateAfterInactivity);
+        }
+
+    }
+
+}
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
@@ -43,7 +43,6 @@ import org.apache.hc.core5.util.Timeout;
 public class RequestConfig implements Cloneable {
 
     private static final Timeout DEFAULT_CONNECTION_REQUEST_TIMEOUT = Timeout.ofMinutes(3);
-    private static final Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3);
     private static final TimeValue DEFAULT_CONN_KEEP_ALIVE = TimeValue.ofMinutes(3);
 
     public static final RequestConfig DEFAULT = new Builder().build();
@@ -69,7 +68,7 @@ public class RequestConfig implements Cloneable {
     */
     protected RequestConfig() {
         this(false, null, null, false, false, 0, false, null, null,
-                DEFAULT_CONNECTION_REQUEST_TIMEOUT, DEFAULT_CONNECT_TIMEOUT, null, DEFAULT_CONN_KEEP_ALIVE, false, false);
+                DEFAULT_CONNECTION_REQUEST_TIMEOUT, null, null, DEFAULT_CONN_KEEP_ALIVE, false, false);
     }
 
     RequestConfig(
@@ -115,7 +114,11 @@ public class RequestConfig implements Cloneable {
 
     /**
      * @see Builder#setProxy(HttpHost)
+     *
+     * @deprecated Use {@link org.apache.hc.client5.http.impl.routing.DefaultProxyRoutePlanner}
+     * or a custom {@link org.apache.hc.client5.http.routing.HttpRoutePlanner}.
      */
+    @Deprecated
     public HttpHost getProxy() {
         return proxy;
     }
@@ -178,7 +181,10 @@ public class RequestConfig implements Cloneable {
 
     /**
      * @see Builder#setConnectTimeout(Timeout)
+     *
+     * @deprecated Use {@link ConnectionConfig#getConnectTimeout()}.
      */
+    @Deprecated
     public Timeout getConnectTimeout() {
         return connectTimeout;
     }
@@ -286,7 +292,6 @@ public class RequestConfig implements Cloneable {
             this.maxRedirects = 50;
             this.authenticationEnabled = true;
             this.connectionRequestTimeout = DEFAULT_CONNECTION_REQUEST_TIMEOUT;
-            this.connectTimeout = DEFAULT_CONNECT_TIMEOUT;
             this.contentCompressionEnabled = true;
             this.hardCancellationEnabled = true;
         }
@@ -323,7 +328,11 @@ public class RequestConfig implements Cloneable {
          * <p>
          * Default: {@code null}
          * </p>
+         *
+         * @deprecated Use {@link org.apache.hc.client5.http.impl.routing.DefaultProxyRoutePlanner}
+         * or a custom {@link org.apache.hc.client5.http.routing.HttpRoutePlanner}.
          */
+        @Deprecated
         public Builder setProxy(final HttpHost proxy) {
             this.proxy = proxy;
             return this;
@@ -444,7 +453,10 @@ public class RequestConfig implements Cloneable {
          * <p>
          * Default: 3 minutes
          * </p>
+         *
+         * @deprecated Use {@link ConnectionConfig.Builder#setConnectTimeout(Timeout)}.
          */
+        @Deprecated
         public Builder setConnectTimeout(final Timeout connectTimeout) {
             this.connectTimeout = connectTimeout;
             return this;
@@ -452,7 +464,10 @@ public class RequestConfig implements Cloneable {
 
         /**
          * @see #setConnectTimeout(Timeout)
+         *
+         * @deprecated Use {@link ConnectionConfig.Builder#setConnectTimeout(long, TimeUnit)}.
          */
+        @Deprecated
         public Builder setConnectTimeout(final long connectTimeout, final TimeUnit timeUnit) {
             this.connectTimeout = Timeout.of(connectTimeout, timeUnit);
             return this;
@@ -570,7 +585,7 @@ public class RequestConfig implements Cloneable {
                     targetPreferredAuthSchemes,
                     proxyPreferredAuthSchemes,
                     connectionRequestTimeout != null ? connectionRequestTimeout : DEFAULT_CONNECTION_REQUEST_TIMEOUT,
-                    connectTimeout != null ? connectTimeout : DEFAULT_CONNECT_TIMEOUT,
+                    connectTimeout,
                     responseTimeout,
                     connectionKeepAlive != null ? connectionKeepAlive : DEFAULT_CONN_KEEP_ALIVE,
                     contentCompressionEnabled,
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalH2ConnPool.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalH2ConnPool.java
@@ -1,0 +1,102 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl.async;
+
+import org.apache.hc.client5.http.config.ConnectionConfig;
+import org.apache.hc.core5.concurrent.CallbackContribution;
+import org.apache.hc.core5.concurrent.FutureCallback;
+import org.apache.hc.core5.function.Resolver;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.nio.ssl.TlsStrategy;
+import org.apache.hc.core5.http2.nio.pool.H2ConnPool;
+import org.apache.hc.core5.io.CloseMode;
+import org.apache.hc.core5.io.ModalCloseable;
+import org.apache.hc.core5.reactor.ConnectionInitiator;
+import org.apache.hc.core5.reactor.IOSession;
+import org.apache.hc.core5.util.TimeValue;
+import org.apache.hc.core5.util.Timeout;
+
+import java.net.InetSocketAddress;
+import java.util.concurrent.Future;
+
+class InternalH2ConnPool implements ModalCloseable {
+
+    private final H2ConnPool connPool;
+
+    private volatile Resolver<HttpHost, ConnectionConfig> connectionConfigResolver;
+
+    InternalH2ConnPool(final ConnectionInitiator connectionInitiator,
+                       final Resolver<HttpHost, InetSocketAddress> addressResolver,
+                       final TlsStrategy tlsStrategy) {
+        this.connPool = new H2ConnPool(connectionInitiator, addressResolver, tlsStrategy);
+    }
+
+    public void close(final CloseMode closeMode) {
+        connPool.close(closeMode);
+    }
+
+    public void close() {
+        connPool.close();
+    }
+
+    private ConnectionConfig resolveConnectionConfig(final HttpHost httpHost) {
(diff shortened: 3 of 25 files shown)


michael-o:
Massive, I will need some time to digest. Thank you.

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
```diff
@@ -0,0 +1,232 @@
...
+    ConnectionMgmtConfig(
+            final HttpHost proxy,
+            final Timeout socketTimeout,
+            final Timeout connectionRequestTimeout,
+            final Timeout connectTimeout) {
+        super();
+        this.proxy = proxy;
+        this.socketTimeout = socketTimeout;
+        this.connectionRequestTimeout = connectionRequestTimeout;
+        this.connectTimeout = connectTimeout;
```
carterkozak, line 71: My understanding is that there are two sets of configuration options:
1. Options which apply to the connection pool as a whole
2. Options that may differ based on individual requests

I think the `connectTimeout` should generally apply to an entire pool of connections (within a route anyhow), while the `connectionRequestTimeout` is more likely to be configured on a per-request basis. Although if pooling is disabled, that may not be the case, but in the modern era of http/1.1 and greater it's probably reasonable to assume pooling is used.

It's not clear to me that all of these options should be configurable per-request, while others should. So I suppose my questions are:
Should this be split into two separate configurations for per-pool/route configuration, and per-request configuration? 
Should the division between `ConnectionMgmtConfig` and `RequestConfig` be different?

rschmitt: > Options which apply to the connection pool as a whole

It seems that the current convention is to specify those options directly on the connection manager, e.g. connection TTL, idle connection validation, connection pool queueing and concurrency policies, etc. So if we wanted to make `connectTimeout` specify behavior for the pool as a whole without the ability to overload `connectTimeout` on a per-request basis, I assume we would add that option to `PoolingHttpClientConnectionManagerBuilder` and similar classes.

rschmitt:
It seems that `RequestConfig` and `ConnectionMgmtConfig` are always passed together and both specify per-request options. In that case, why not declare `ConnectionMgmtConfig` _on_ `RequestConfig`? In addition to simplifying the diff, it will make it clear that `ConnectionMgmtConfig` are all request-scoped (so to speak).

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
michael-o: I suppor the config on the pool itself. Using Commons Pool or Tomcat JDBC Pool I specify how long it should block until a get a valid option from. This is an immutable config on the pool and not on the request. This should be completely opaque to the client code.

michael-o (changes requested):
Addtionally to the comments: I'd like to understand the followinig:
Why has the proxy host has been moved to connection mgmt? In my understanding an HTTP proxy does not operate on socket level, but on protocol level. This would make sense if we'd use SOCKS5. Can you explainß

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
ok2c: @carterkozak @rschmitt @michael-o All right. I will see if I can move `ConnectionMgmtConfig` to connection managers, though I am sure it is a matter of time till someone pops up on the mailing list and demands those settings to be made available on a per request basis.

ok2c:
> Addtionally to the comments: I'd like to understand the followinig:
> Why has the proxy host has been moved to connection mgmt? In my understanding an HTTP proxy does not operate on socket level, but on protocol level. This would make sense if we'd use SOCKS5. Can you explain?

HTTP proxy do essentially operate on the socket level once a tunnel has been established through it. However, I will deprecate proxy settings in `RequestConfig` in favor of `HttpRoutePlanner`. People should be using `HttpRoutePlanner` for custom routing strategies anyway.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -42,8 +42,6 @@
 @Contract(threading = ThreadingBehavior.IMMUTABLE)
 public class RequestConfig implements Cloneable {
 
-    private static final Timeout DEFAULT_CONNECTION_REQUEST_TIMEOUT = Timeout.ofMinutes(3);
-    private static final Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3);
     private static final TimeValue DEFAULT_CONN_KEEP_ALIVE = TimeValue.ofMinutes(3);
```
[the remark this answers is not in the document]
ok2c: @michael-o You are correct. This is a request level parameter. I will keep it in `RequestConfig`.

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
michael-o: @ok2c I think the challenge here is that the connect timeout should actually be or additionally be on a per-route basis because the conn manager handles connections to different servers w/o different attributes in contrast to a database conn pool which in most cases connects to a single host. WDYT?

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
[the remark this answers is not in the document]
michael-o: I actually meant `DEFAULT_CONN_KEEP_ALIVE`. Which did you mean?

michael-o:
> 
> 
> > Addtionally to the comments: I'd like to understand the followinig:
> > Why has the proxy host has been moved to connection mgmt? In my understanding an HTTP proxy does not operate on socket level, but on protocol level. This would make sense if we'd use SOCKS5. Can you explain?
> 
> HTTP proxy do essentially operate on the socket level once a tunnel has been established through it. However, I will deprecate proxy settings in `RequestConfig` in favor of `HttpRoutePlanner`. People should be using `HttpRoutePlanner` for custom routing strategies anyway.

Agreed, sounds better than moving to `ConnMgmtConfig`.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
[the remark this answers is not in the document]
ok2c: > I actually meant `DEFAULT_CONN_KEEP_ALIVE`. Which did you mean?

I was referring to DEFAULT_CONNECTION_REQUEST_TIMEOUT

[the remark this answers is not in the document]
michael-o: That's also true...

ok2c:
@rschmitt @carterkozak @rschmitt 
Here's the second take:
* new `ConnectionMgmtConfig` structure configurable on the connection manager level
* proxy setting in `RequestConfig` deprecated in favor of `HttpRoutePlanner`
* connect timeout setting in `RequestConfig` deprecated in favor of `ConnectionMgmtConfig`
* `validateAfterInactivity` setting deprecated in favor `ConnectionMgmtConfig`

httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/Executor.java
```diff
@@ -64,7 +65,9 @@
                         .useSystemProperties()
                         .setMaxConnPerRoute(100)
                         .setMaxConnTotal(200)
-                        .setValidateAfterInactivity(TimeValue.ofSeconds(10))
+                        .setDefaultConnectionMgmtConfig(ConnectionMgmtConfig.custom()
```
carterkozak, line 68: I would slightly prefer expanding "Mgmt" to "Management" in method and class names for readability, but I don't have a strong opinion if you feel otherwise.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -495,13 +513,58 @@ public void setDefaultSocketConfig(final SocketConfig defaultSocketConfig) {
...
+     */
+    public void setConnectionMgmtConfig(final HttpRoute route, final ConnectionMgmtConfig connectionMgmtConfig) {
+        connectionMgmtConfigMap.put(route, connectionMgmtConfig);
+    }
+
+    /**
+     * @since 5.2
+     */
+    public ConnectionMgmtConfig getConnectionMgmtConfig(final HttpRoute route) {
+        return connectionMgmtConfigMap.get(route);
```
carterkozak, line 555: Should this return the default configuration when no route-specific configuration exists?
`return connectionMgmtConfigMap.getOrDefault(route, defaultConnectionMgmtConfig);`

It's not clear if this is intended to expose the configuration that we use for a given route, or provide access to route-specific overrides.

I think the same argument can be made for the SocketConfig.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -113,10 +116,12 @@
     private final HttpClientConnectionOperator connectionOperator;
     private final ManagedConnPool<HttpRoute, ManagedHttpClientConnection> pool;
     private final HttpConnectionFactory<ManagedHttpClientConnection> connFactory;
+    private final ConcurrentMap<HttpRoute, SocketConfig> socketConfigMap;
+    private final ConcurrentMap<HttpRoute, ConnectionMgmtConfig> connectionMgmtConfigMap;
```
carterkozak, line 120: Should we worry about routes building up over time without a path to removal and garbage collection?

I wonder if we should provide an interface to optionally build a route-specific socket and ConnMgmtConfig the first time a route is used, which can be forgotten and recalculated later if all connections to that route are cleared (or even naively recalculated for each connection, allowing implementations to cache internally).

```java
interface RouteBasedSocketConfigProvider {
    SocketConfig socketConfigForRoute(HttpRoute route);
}
```

ok2c: @carterkozak Using a `Provider` interface for the same end is a great idea. Thank you! 
I will re-work the PR.

michael-o:
Going through...

httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/Executor.java
michael-o: @carterkozak I wonder why we need `Mgmt` at all. Do we have `RequestMmgtConfig`. `ConnectionMgmt..` makes sense when it has influence on the connection tool manager. Then it makes sense, but a full word is better to me too.

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
```diff
@@ -0,0 +1,202 @@
...
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+public class ConnectionMgmtConfig implements Cloneable {
+
+    private static final Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3);
+
+    public static final ConnectionMgmtConfig DEFAULT = new Builder().build();
+
+    private final Timeout socketTimeout;
+    private final Timeout connectTimeout;
+    private final TimeValue validateAfterInactivity;
```
michael-o, line 51: Can you sort them like in the constructor, please?

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
```diff
@@ -0,0 +1,232 @@
...
+         */
+        public Builder setSocketTimeout(final int soTimeout, final TimeUnit timeUnit) {
+            this.socketTimeout = Timeout.of(soTimeout, timeUnit);
+            return this;
+        }
+
+        /**
+         * Determines the default socket timeout value for I/O operations.
+         * <p>
+         * Default: 3 minutes
```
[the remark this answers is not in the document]
michael-o: This one still holds true for this class.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/ClientConfiguration.java
```diff
@@ -172,15 +173,17 @@ public Header parseHeader(final CharArrayBuffer buffer) {
...
-        connManager.setDefaultSocketConfig(socketConfig);
-        // Validate connections after 1 sec of inactivity
-        connManager.setValidateAfterInactivity(TimeValue.ofSeconds(10));
+        connManager.setDefaultSocketConfig(SocketConfig.custom()
+                .setTcpNoDelay(true)
+                .build());
+        // Validate connections after 10 sec of inactivity
+        connManager.setDefaultConnectionMgmtConfig(ConnectionMgmtConfig.custom()
+                .setConnectTimeout(Timeout.ofSeconds(30))
+                .setSocketTimeout(Timeout.ofSeconds(30))
```
michael-o, line 184: When I see the socket config above I am confused that the socket timeout doesn't happen there. General question to other reviewers: Should that be rather read `readTimeout`?

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -43,7 +43,6 @@
 public class RequestConfig implements Cloneable {
 
     private static final Timeout DEFAULT_CONNECTION_REQUEST_TIMEOUT = Timeout.ofMinutes(3);
-    private static final Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3);
     private static final TimeValue DEFAULT_CONN_KEEP_ALIVE = TimeValue.ofMinutes(3);
```
michael-o, line 46: Another stupid question: `DEFAULT_CONN_KEEP_ALIVE` applies to the lifetime of the connection. Why is is retained here and not moved to the new config class?

michael-o (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
ok2c: > 
> 
> Another stupid question: `DEFAULT_CONN_KEEP_ALIVE` applies to the lifetime of the connection. Why is is retained here and not moved to the new config class?

The `keep-alive` period is determined on a per request basis, generally based on `Keep-Alive` header.

ok2c:
@carterkozak @michael-o Would you like `ConnectionConfig` better than `ConnectionMgmtConfig`? `ConnectionManagementConfig` is just too unwieldy in my opinion.

michael-o:
> 
> 
> @carterkozak @michael-o Would you like `ConnectionConfig` better than `ConnectionMgmtConfig`? `ConnectionManagementConfig` is just too unwieldy in my opinion.

Since we have `RequestConfig` and `SocketConfig` it would feel at least consistent to me.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
michael-o: But this will affect the pool object after the request has been completed, doesn't it?

ok2c: @michael-o Yes, it would but this decision cannot be made by the connection management layer. It must be made by the HTTP protocol layer.

michael-o: So the overall idle time can change over time depending on this or the header value?

carterkozak:
ConnectionConfig sounds good to me.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
carterkozak: Based on the request and response I think there are three results:
1. Do not keep the connection alive (`Connection: close` response header)
2. keep the connection alive for a specific amount of time `Keep-Alive: timeout=30`
3. keep the connection alive for an indeterminate amount of time so the default value can be used (http/1.1 without specifying close)

If memory serves, we use `null` for [1] (connection is not kept alive) and return a value based on either a keep-alive timeout value [2] or the default [3] when configured. If we can update the abstraction to differentiate states [1] from [3] I think the default configuration makes sense in the connection management configuration.

ok2c: @carterkozak To me, this is clearly an HTTP protocol layer responsibility, but if you, guys, are absolutely convinced I can move this setting to `ConnectionConfig`.

carterkozak: I defer to your judgement, I could make a case for either design.

ok2c:
@carterkozak @michael-o Please do another pass.
The only big change compared to the previous change-set is the use of config resolvers instead of internal maps.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalH2ConnPool.java
```diff
@@ -0,0 +1,102 @@
...
+        connPool.close(closeMode);
+    }
+
+    public void close() {
+        connPool.close();
+    }
+
+    private ConnectionConfig resolveConnectionConfig(final HttpHost httpHost) {
+        final ConnectionConfig connectionConfig = connectionConfigResolver != null
+                ? connectionConfigResolver.resolve(httpHost) : null;
```
carterkozak, line 68: Potential race which results in an NPE
```suggestion
        final Resolver<HttpHost, ConnectionConfig> snapshot = connectionConfigResolver;
        final ConnectionConfig connectionConfig = snapshot != null ? snapshot.resolve(httpHost) : null;
```

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -243,6 +245,16 @@ private InternalConnectionEndpoint cast(final ConnectionEndpoint endpoint) {
...
     }
 
+    private ConnectionConfig resolveConnectionConfig(final HttpRoute route) {
+        final ConnectionConfig connectionConfig = connectionConfigResolver != null ? connectionConfigResolver.resolve(route) : null;
+        return connectionConfig != null ? connectionConfig : ConnectionConfig.DEFAULT;
+    }
+
+    private SocketConfig resolveSocketConfig(final HttpRoute route) {
+        final SocketConfig socketConfig = socketConfigResolver != null ? socketConfigResolver.resolve(route) : null;
+        return socketConfig != null ? socketConfig : SocketConfig.DEFAULT;
```
carterkozak, line 255: These two methods should use a single volatile read on the resolver as well.

carterkozak:
Aside from the minor threading issue, this looks good to me. Thanks!

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
ok2c: @carterkozak Corrected.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
```diff
@@ -179,6 +181,20 @@ public synchronized void setSocketConfig(final SocketConfig socketConfig) {
         this.socketConfig = socketConfig != null ? socketConfig : SocketConfig.DEFAULT;
     }
 
+    /**
+     * @since 5.2
+     */
+    public synchronized ConnectionConfig getConnectionMgmtConfig() {
```
michael-o, line 187: Mgmt leftover?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
```diff
@@ -179,6 +181,20 @@ public synchronized void setSocketConfig(final SocketConfig socketConfig) {
...
+        return connectionConfig;
+    }
+
+    /**
+     * @since 5.2
+     */
+    public synchronized void setConnectionMgmtConfig(final ConnectionConfig connectionConfig) {
```
michael-o, line 194: Mgmt leftover?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
```diff
@@ -221,6 +229,7 @@ private InternalConnectionEndpoint cast(final AsyncConnectionEndpoint endpoint)
             LOG.debug("{} endpoint lease request ({}) {}", id, requestTimeout, ConnPoolSupport.formatStats(route, state, pool));
         }
         final ComplexFuture<AsyncConnectionEndpoint> resultFuture = new ComplexFuture<>(callback);
+        final ConnectionConfig connMgmtConfig = resolveConnectionConfig(route);
```
michael-o, line 232: Mgmt leftover?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
```diff
@@ -242,7 +251,7 @@ void leaseCompleted(final PoolEntry<HttpRoute, ManagedAsyncClientConnection> poo
                     @Override
                     public void completed(final PoolEntry<HttpRoute, ManagedAsyncClientConnection> poolEntry) {
                         final ManagedAsyncClientConnection connection = poolEntry.getConnection();
-                        final TimeValue timeValue = PoolingAsyncClientConnectionManager.this.validateAfterInactivity;
+                        final TimeValue timeValue = connMgmtConfig != null ? connMgmtConfig.getValidateAfterInactivity() : null;
```
michael-o, line 254: Mgmt leftover?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
```diff
@@ -357,6 +365,9 @@ public void release(final AsyncConnectionEndpoint endpoint, final Object state,
             host = route.getTargetHost();
         }
         final InetSocketAddress localAddress = route.getLocalSocketAddress();
+        final ConnectionConfig connMgmtConfig = resolveConnectionConfig(route);
```
michael-o, line 368: Mgmt leftover?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
```diff
@@ -369,6 +380,10 @@ public void completed(final ManagedAsyncClientConnection connection) {
                             if (LOG.isDebugEnabled()) {
                                 LOG.debug("{} connected {}", ConnPoolSupport.getId(endpoint), ConnPoolSupport.getId(connection));
                             }
+                            final Timeout socketTimeout = connMgmtConfig.getSocketTimeout();
```
michael-o, line 383: Mgmt leftover?

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionMgmtConfig.java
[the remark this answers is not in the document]
michael-o: Still wrong.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
michael-o: I am accepting @ok2c's argumentation.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -43,7 +43,6 @@
 public class RequestConfig implements Cloneable {
 
     private static final Timeout DEFAULT_CONNECTION_REQUEST_TIMEOUT = Timeout.ofMinutes(3);
```
michael-o, line 45: Yet another stupid question: Since this is the time how long the pool at most should require to return a working connection, why is this a HTTP request level config?

michael-o (changes requested):
@ok2c I have found a few leftovers and a few questions. Please have a look.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
ok2c: @michael-o This really depends how one looks at it. This setting is used by each and every request and as such makes sense at the request level. However, indeed, the inner machinery is entirely in the connection manager's responsibility. 
I initially moved that parameter to `ConnectionConfig` but if I remember it correctly you questioned that decision. 
Basically as far as I am concerned it can be either way. Blonds or Brunettes, so to speak. However I tend to see it more of a request level setting.

ok2c:
> 
> 
> @ok2c I have found a few leftovers and a few questions. Please have a look.

Thank you for spotting the issues. Corrected. Please have a look and let me know if you are conceptually OK with the change. We can still move individual settings around in the course of 5.2 development, so this is definitely not the final state of things.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalH2ConnPool.java
```diff
@@ -0,0 +1,102 @@
...
+        final Resolver<HttpHost, ConnectionConfig> resolver = this.connectionConfigResolver;
+        final ConnectionConfig connectionConfig = resolver != null ? resolver.resolve(httpHost) : null;
+        return connectionConfig != null ? connectionConfig : ConnectionConfig.DEFAULT;
+    }
+
+    public Future<IOSession> getSession(
+            final HttpHost endpoint,
+            final Timeout connectTimeout,
+            final FutureCallback<IOSession> callback) {
+        final ConnectionConfig connMgmtConfig = resolveConnectionConfig(endpoint);
```
michael-o, line 76: yet another one

michael-o:
> 
> 
> > @ok2c I have found a few leftovers and a few questions. Please have a look.
> 
> Thank you for spotting the issues. Corrected. Please have a look and let me know if you are conceptually OK with the change. We can still move individual settings around in the course of 5.2 development, so this is definitely not the final state of things.

Did you push? I still can find a lot of `Mgmt` in the changeset.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
michael-o: Let's leave it as-is, but I still think that a connection lease timeout should be on a pool. See here: https://github.com/apache/httpcomponents-client/pull/296#discussion_r597967488

ok2c:
> Did you push? I still can find a lot of Mgmt in the changeset.

@michael-o I overlooked variable names. Corrected.

Do you want me to move `ConnectionRequestTimeout` to `ConnectionConfig`? I can do that but I suppose at some point we will have to deal with a bunch of angry people demanding those properties to be un-deprecated. `ConnectTimeout` would be a easier to argue about. `ConnectionRequestTimeout` is less so.

michael-o (approved):

michael-o:
> 
> 
> > Did you push? I still can find a lot of Mgmt in the changeset.
> 
> @michael-o I overlooked variable names. Corrected.
> 
> Do you want me to move `ConnectionRequestTimeout` to `ConnectionConfig`? I can do that but I suppose at some point we will have to deal with a bunch of angry people demanding those properties to be un-deprecated. `ConnectTimeout` would be a easier to argue about. `ConnectionRequestTimeout` is less so.

No, we should discuss this separately. It is basically a `connectionLeaseTimeout` from a connection manage. I am fine with the rest of the code....

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
```diff
@@ -0,0 +1,205 @@
...
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+public class ConnectionConfig implements Cloneable {
+
+    private static final Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3);
+
+    public static final ConnectionConfig DEFAULT = new Builder().build();
+
+    private final Timeout connectTimeout;
+    private final Timeout socketTimeout;
+    private final TimeValue validateAfterInactivity;
```
001mertiya, line 51: Hi @ok2c , If the plan was to move all connection-related configurations to ConnectionConfig, why is connectionRequestTimeout not included here in ConnectionConfigs?

ok2c: @001mertiya Yes, this is the plan. The `connectionRequestTimeout` however is a request level parameter that applies to every request execution. The `connectTimeout` is a connection level parameter that applies only when a new connection needs to be established.

001mertiya: Correct, but `connectionRequestTimeout` is the timeout used when requesting a connection from the connection pool. I think it should be part of `ConnectionConfig` or there should be a way to set it in `PoolingAsyncClientConnectionManager`, as it is a property of the connection pool.

ok2c: @001mertiya Again, this is a _request_ level parameter that applies to request execution and not to connection creation. There are enough reasons for users to configure connection request timeout on a per request basis.

001mertiya: Makes sense. I was thinking there should be a way to set it in `PoolingAsyncClientConnectionManager` since it’s a pool property. Also, what about having `socketTimeout` in `ConnectionConfig` and `responseTimeout` in `RequestConfig` - don’t they both serve the same purpose?

ok2c: @001mertiya They are the same. Socket timeout can be set at three levels: 1. transport level (socket for classic and channel for async) 2. connection level and 3.  request level. when the response timeout is not explicitly set, the connection socket timeout applies. When connection socket timeout is not set the transport socket timeout applies (which is usually always to some default value).  This way users can override their socket timeout setting either for individual connections or individual requests

001mertiya: Correct. Ideally, for responseTimeout, the client should throw a `RequestCancelledException` with the message: `Request cancelled by client after reaching responseTimeout specified in the RequestConfig` instead of a `SocketTimeoutException`. What are your thoughts?

ok2c: @001mertiya No, we cannot do that. `RequestCancelledException` is thrown when it is absolutely certain the request has not even been sent to the opposite endpoint and is absolutely safe to re-execute. With the socket timeout the server could have already processed the request but the response did not reach the client in time. We cannot assume the request has not been already acted upon and is safe to re-execute.

PR #302, opened 2021-04-09
HTTPCLIENT-2149: When no dNSName, match against CN

peterdettman:
https://issues.apache.org/jira/browse/HTTPCLIENT-2149

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
@@ -287,29 +287,7 @@ public final class DefaultHostnameVerifier implements HttpClientHostnameVerifier
     }
 
     static List<SubjectName> getSubjectAltNames(final X509Certificate cert) {
-        try {
-            final Collection<List<?>> entries = cert.getSubjectAlternativeNames();
-            if (entries == null) {
-                return Collections.emptyList();
-            }
-            final List<SubjectName> result = new ArrayList<>();
-            for (final List<?> entry : entries) {
-                final Integer type = entry.size() >= 2 ? (Integer) entry.get(0) : null;
-                if (type != null) {
-                    if (type == SubjectName.DNS || type == SubjectName.IP) {
-                        final Object o = entry.get(1);
-                        if (o instanceof String) {
-                            result.add(new SubjectName((String) o, type));
-                        } else if (o instanceof byte[]) {
-                            // TODO ASN.1 DER encoded form
-                        }
-                    }
-                }
-            }
-            return result;
-        } catch (final CertificateParsingException ignore) {
-            return Collections.emptyList();
-        }
+        return getSubjectAltNames(cert, -1);
     }
 
     static List<SubjectName> getSubjectAltNames(final X509Certificate cert, final int subjectName) {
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
```diff
@@ -114,18 +114,15 @@ public void verify(
                     matchIPv6Address(host, subjectAlts);
                     break;
                 default:
-                    matchDNSName(host, subjectAlts, this.publicSuffixMatcher);
+                    // In case there are no SubjectName.DNS entries, fallback to CN matching
```
ok2c, line 117: @peterdettman I am not sure this is what the RFC says. I believe that CN matching is to be used _only_ if there are no `dNSName` entries present.

ok2c: @peterdettman I only expected changes in handling of URIs specified as an IP address rather than a hostname.

peterdettman: Yes, matchDNSNames is only returning false if there were no dNSName at all. If there were any present that didn’t match, it would have already thrown an exception.

ok2c: @peterdettman Then what is the point in calling `#matchCN` here?

peterdettman: That is the fallback if there were no dNSName entries to even check against. RFC 2818 3.1: “If a subjectAltName extension of type dNSName is present, that MUST be used as the identity. Otherwise, the (most specific) Common Name field in the Subject field of the certificate MUST be used.”. Note that IP addresses are singled out further down: “In some cases, the URI is specified as an IP address rather than a hostname. In this case, the iPAddress subjectAltName must be present in the certificate and must exactly match the IP in the URI.”.

peterdettman: Perhaps the code could be clearer if we only collected either IP or DNS entries depending on the HostNameType?

peterdettman: For comparison, SunJSSE uses [HostnameChecker](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/security/util/HostnameChecker.java) for the equivalent checks when endpoint identification is set to HTTPS. Likewise our (BCJSSE provider) uses [HostnameUtil](https://github.com/bcgit/bc-java/blob/master/tls/src/main/java/org/bouncycastle/jsse/provider/HostnameUtil.java).

ok2c: > Perhaps the code could be clearer if we only collected either IP or DNS entries depending on the HostNameType?

@peterdettman This is what I was trying to hint at. I really do not understand why there should be multiple invocations of `#matchCN`. Please refactor the code to make the execution flow as well as the intent of your changes a bit easier to understand.

ok2c (changes requested):

peterdettman:
Have updated so that matchCN is no longer called when host is an IP address, and to clarify the handling for the DNS case.

ok2c:
@peterdettman This is so much better. One last thing. Is there a chance we could avoid having two almost identical `#getSubjectAltNames` methods?

peterdettman:
I've removed the redundancy. Note that `#getSubjectAltNames(X509Certificate)` now will return all names (that have String values), not just DNS, IP. For the only existing caller (`TlsSessionValidator` failure logging), that seems OK.

ok2c (approved):
@peterdettman Awesome.

PR #303, opened 2021-04-23
Extendable ClosableHttpClient

cdalexndr:
Allow extension of ClosableHttpClient.

Closes https://issues.apache.org/jira/browse/HTTPCLIENT-2154

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/CloseableHttpClient.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/CloseableHttpClient.java
@@ -56,23 +56,17 @@ public abstract class CloseableHttpClient implements HttpClient, ModalCloseable 
 
     private static final Logger LOG = LoggerFactory.getLogger(CloseableHttpClient.class);
 
-    protected abstract CloseableHttpResponse doExecute(HttpHost target, ClassicHttpRequest request,
-                                                     HttpContext context) throws IOException;
-
-    @Override
-    public CloseableHttpResponse execute(
+    public abstract CloseableHttpResponse execute(
             final HttpHost target,
             final ClassicHttpRequest request,
-            final HttpContext context) throws IOException {
-        return doExecute(target, request, context);
-    }
+            final HttpContext context) throws IOException ;
 
     @Override
     public CloseableHttpResponse execute(
             final ClassicHttpRequest request,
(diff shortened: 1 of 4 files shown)


ok2c:
@cdalexndr I am not sure I understand the rationale for removal of `#doExecute` method. How does this method prevent `ClosableHttpClient` extension?
Besides, we cannot just remove protected method without breaking API compatibility with previous 5.x releases.

cdalexndr:
The `HttpClientBuilder.build()` returns `CloseableHttpClient` instance, and I need to wrap it to provide custom functionality.
[HTTPCLIENT-2154](https://issues.apache.org/jira/browse/HTTPCLIENT-2154) contains example code: using a wrapper (decorator pattern) doesn't work (see error comment):
```
    class CustomCloseableHttpClient extends CloseableHttpClient {
        private CloseableHttpClient base;

        public CustomCloseableHttpClient( CloseableHttpClient httpClient ) {
            this.base= httpClient;
        }

        @Override
        protected CloseableHttpResponse doExecute( HttpHost httpHost, HttpRequest httpRequest, HttpContext httpContext ) throws IOException, ClientProtocolException {
            return base.doExecute( httpHost, httpRequest, httpContext ); //<-- error, cannot call protected method
        }
        ....
```

ok2c:
@cdalexndr 
```
public class DecoratedCloseableHttpClient extends CloseableHttpClient {

    private final CloseableHttpClient closeableHttpClient;

    public DecoratedCloseableHttpClient(CloseableHttpClient closeableHttpClient) {
        this.closeableHttpClient = closeableHttpClient;
    }

    @Override
    protected CloseableHttpResponse doExecute(HttpHost target, ClassicHttpRequest request, HttpContext context) throws IOException {
        return closeableHttpClient.execute(target, request, context);
    }

    @Override
    public void close(CloseMode closeMode) {
        closeableHttpClient.close(closeMode);
    }

    @Override
    public void close() throws IOException {
        closeableHttpClient.close();
    }
}
```
I still do not understand the problem.

ok2c:
@cdalexndr What is stopping you from using the public `#execute` method?

cdalexndr:
> @cdalexndr What is stopping you from using the public `#execute` method?

This can lead to side effects, but for my use case I guess it will work.

Ideally, `HttpClientBuilder.build()` should return an `interface` that doesn't have this issue.

PR #309, opened 2021-06-20
Connection lease request cancellation bug

ok2c:
@rschmitt Fundamentally, the core problem here is whose responsibility it is to release a pool entry back to the pool if the lease request simultaneously succeeds and gets cancelled. 

This is not a _final_ fix but merely a concept. I still need to figure the back way of dealing with the problem. I also need to fix the same defect in the async connection manager implementation.

What truly baffles me is how come this issue has never been reported with HttpClient 4.5 given it appears to be equally affected.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -28,7 +28,6 @@ package org.apache.hc.client5.http.impl.io;
 
 import java.io.IOException;
 import java.util.Set;
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeoutException;
@@ -270,9 +269,6 @@ public class PoolingHttpClientConnectionManager
                 final PoolEntry<HttpRoute, ManagedHttpClientConnection> poolEntry;
                 try {
                     poolEntry = leaseFuture.get(timeout.getDuration(), timeout.getTimeUnit());
-                    if (poolEntry == null || leaseFuture.isCancelled()) {
-                        throw new ExecutionException(new CancellationException("Operation cancelled"));
-                    }
                 } catch (final TimeoutException ex) {
                     leaseFuture.cancel(true);
                     throw ex;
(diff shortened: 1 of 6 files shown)


carterkozak:
déjà vu, I've run into this flavor of leak a number of times because it's incredibly difficult to handle `cancel()` on any `Future<? extends AutoCloseable>`.

> the core problem here is whose responsibility it is to release a pool entry back to the pool if the lease request simultaneously succeeds and gets cancelled.

This exactly matches my experience, it's nontrivial to determine the singular 'owner' for a given future instance.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -329,7 +331,22 @@ public synchronized ConnectionEndpoint get(
 
             @Override
             public boolean cancel() {
-                return leaseFuture.cancel(true);
+                if (cancelled.compareAndSet(false, true)) {
+                    final boolean result = leaseFuture.cancel(true);
+                    if (!result) {
+                        // The lease request has already completed successfully and cannot be cancelled
+                        try {
+                            final PoolEntry<HttpRoute, ManagedHttpClientConnection> poolEntry = leaseFuture.get(1, TimeUnit.MILLISECONDS);
```
rschmitt, line 339: I'd prefer to see `leaseFuture.get(0, TimeUnit.MILLISECONDS)` here, by way of asserting that this call should not block

rschmitt:
This proposal passes all of my regression tests. Would it allow us to also simplify the cancellation checks in classes like `InternalExecRuntime`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
ok2c: @rschmitt This call should never block given at this point that the future is complete but you are absolutely right. Will do so.

ok2c:
> This proposal passes all of my regression tests. Would it allow us to also simplify the cancellation checks in classes like `InternalExecRuntime`?

Great. Please allow me a few more day to work on the test coverage and also fix the same defect in the async code. I am also going to review and clean up `InternalExecRuntime` at the same time.

ok2c:
@rschmitt I came up with what I beleive is a better fix. I have also added test cases to verify correctness of new code.

Essentially I was going about it all wrong. I was trying to ensure that request execution gets aborted as early as possible and by doing so I was creating more points of potential thread race conditions. Now I simplified things by letting the request execution procced to reach a point where it is easier to avoid race conditions and to ensure correct resource deallocation. Several chunks of ugly code is also now gone from `InternalExecRuntime`.

Please review.

rschmitt (approved):
LGTM, this is roughly what I expected. When I get back from vacation I'll make sure to run our internal regression tests against these changes, just to be sure.

PR #311, opened 2021-08-22
Don't retry a request for NoRouteToHostException

jaikiran:
As discussed in the mailing list discussion here[1], the commit here includes `java.net.NoRouteToHostException` to the set of exception types, for which request retries are disabled by default. The commit also has a test case which reproduces the issue and verifies the fix.

I had a look at the `java.net.SocketException` hieararchy. It has `BindException`, `ConnectException`, `NoRouteToHostException` and `PortUnreachableException` as its subclasses. The `BindException` is not relevant while connecting as a client. The `ConnectException` is already included in the default set of non-retriable exception types in this `DefaultHttpRequestRetryHandler` class. The `PortUnreachableException` is for ICMP port unreachable cases, so it isn't relevant in this flow. So that leaves only `NoRouteToHostException` which was missing from this set of non-retriable exception type. Adding this should thus make it behave similar like the `ConnectException` and cover the possible exception types.

P.S: I haven't checked if a similar change is needed in 5.x versions of this library.

[1] https://mail-archives.apache.org/mod_mbox/hc-httpclient-users/202108.mbox/%3C3362ac32-1ced-9aad-36d6-2e0e000168ff%40apache.org%3E

### DIFF
--- a/httpclient/src/main/java/org/apache/http/impl/client/DefaultHttpRequestRetryHandler.java
+++ b/httpclient/src/main/java/org/apache/http/impl/client/DefaultHttpRequestRetryHandler.java
@@ -30,6 +30,7 @@ package org.apache.http.impl.client;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.net.ConnectException;
+import java.net.NoRouteToHostException;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


ok2c:
@jaikiran Ported to `5.1.x` and `master`.

jaikiran:
@ok2c, Thank you for the help in reviewing and porting this.

jaikiran:
Hello @ok2c, would it be possible to do a release of 4.5.x version some time in near future? It looks like the workaround that one of the projects added to get past this issue is no longer applicable in Java 18 https://github.com/quarkusio/quarkus/pull/19559#issuecomment-932987301. So it would be good to have a official release of 4.5.x with this fix.

PR #314, opened 2021-10-17
Minor fix.

arturobernalg:
* Fix javadoc typo.
* Use java array declarations instead of "c" way.


arturobernalg:
HI @ok2c @michael-o 
Fair enough. Reverted the change of the iterator.
TY

PR #324, opened 2021-11-11
Fix typo method and class. Deprecate the old ones.

arturobernalg:


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheUpdateHandler.java
```diff
@@ -65,6 +65,25 @@
     /**
      * Creates a cache entry for the given request, origin response message and response content.
      */
+    public HttpCacheEntry createCacheEntry(
```
ok2c, line 68: @arturobernalg `CacheUpdateHandler` is not a part of public API. Deprecation is not needed  here.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExecBase.java
```diff
@@ -139,6 +139,23 @@ public long getCacheUpdates() {
         return cacheUpdates.get();
     }
 
+    /**
```
ok2c, line 142: @arturobernalg Same here. This is not a public method.

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/IgnoreCompleteExceptonFutureCallback.java
```diff
@@ -30,6 +30,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
```
ok2c, line 33: @arturobernalg Please make this class extend `IgnoreCompleteExceptionFutureCallback`.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/IgnoreCompleteExceptonFutureCallback.java
arturobernalg: > @arturobernalg Please make this class extend `IgnoreCompleteExceptionFutureCallback`.

Fixed al the remarks.
TY

ok2c (approved):

PR #326, opened 2021-11-14
Use Args for argument check.

arturobernalg:


michael-o (approved):

httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/HttpByteArrayCacheEntrySerializerTestUtils.java
```diff
@@ -320,9 +321,7 @@ private static int readFully(final InputStream src, final byte[] dest) throws IO
      *                     number of bytes have been read
      */
     static byte[] readFullyStrict(final InputStream src, final long length) throws IOException {
-        if (length > Integer.MAX_VALUE) {
-            throw new IllegalArgumentException(String.format("Length %d is too large to fit in an array", length));
-        }
+        Args.check(length <= Integer.MAX_VALUE, String.format("Length %d is too large to fit in an array", length));
```
carterkozak, line 324: This `String.format` is very expensive, and now it's executed in both the valid path and invalid path, where previously the work was avoided unless it was necessary. I think this should probably be reverted.

Also note that refactoring into the following will yield more efficient code (but to be clear, it doesn't really matter much if it's not executed in most cases):
```java
"Length " + length + " is too large to fit in an array"
```

on java 8 this produces:
```java
new StringBuilder().append("Length ").append(length).append(" is too large to fit in an array").build()
```

and once we target java 11 bytecode will take advantage of [jep-280](https://openjdk.java.net/jeps/280) `StringConcatFactory.makeConcat`.

ok2c: @carterkozak Ah, Jesus. I will roll back the damn commit.

garydgregory: Instead of:
```
Args.check(length <= Integer.MAX_VALUE, String.format("Length %d is too large to fit in an array", length));
```
we should support:
```
Args.check(length <= Integer.MAX_VALUE, () -> String.format("Length %d is too large to fit in an array", length));
```
In the style of `Objects.requireNonNull()` with a `Supplier<String>`

carterkozak: Boxing + varargs allocation isn't ideal, but it beats string allocation and string.format.
Lambda allocation can take a while to optimize away depending on the jvm, especially since it's not static -- the length value must be captured from the call-site. Still beats String.format though.

It's hard to beat a conditional :-)

ok2c:
> It's hard to beat a conditional :-)

Precisely. Varargs still entails an extra array allocation.

ok2c:
This change-set has been backed out.

httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/HttpByteArrayCacheEntrySerializerTestUtils.java
arturobernalg: HI @carterkozak  thank you for the review, I really appreciate it. 
I din´t see that. My bad. 
sorry guys.

PR #330, opened 2021-11-24
HTTPCLIENT-2189 - Cookie and Cache APIs to use Java time primitives.

arturobernalg:
Change all the remain classes to use Time. API´s.
Remarks:

- There are tests with name `Deprecated` that indicate that are using a all method (deprecate).  They should be removed when you decide to stop supporting to old API´s.  This way it is easier to delete all the old code when the time comes.
- I had to add support to the java time package used in Cache entity Serializer/deserializer.
- I tried to keep the same of all variable and method, but in some point was´t possible because the conflict name. I've used xxxInstant, e.g `getExpiryInstant`

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/cookie/BasicCookieStore.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/cookie/BasicCookieStore.java
@@ -29,6 +29,7 @@ package org.apache.hc.client5.http.cookie;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.Iterator;
@@ -83,7 +84,7 @@ public class BasicCookieStore implements CookieStore, Serializable {
             try {
                 // first remove any old cookie that is equivalent
                 cookies.remove(cookie);
-                if (!cookie.isExpired(new Date())) {
+                if (!cookie.isExpired(Instant.now())) {
                     cookies.add(cookie);
                 }
             } finally {
@@ -136,6 +137,7 @@ public class BasicCookieStore implements CookieStore, Serializable {
      * @see Cookie#isExpired(Date)
      */
     @Override
+    @SuppressWarnings("deprecation")
     public boolean clearExpired(final Date date) {
         if (date == null) {
             return false;
@@ -145,6 +147,34 @@ public class BasicCookieStore implements CookieStore, Serializable {
             boolean removed = false;
             for (final Iterator<Cookie> it = cookies.iterator(); it.hasNext(); ) {
                 if (it.next().isExpired(date)) {
+                    it.remove();
+                    removed = true;
+                }
+            }
+            return removed;
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    /**
+     * Removes all of {@link Cookie cookies} in this HTTP state that have expired by the specified
+     * {@link Instant date}.
+     *
+     * @return true if any cookies were purged.
+     * @see Cookie#isExpired(Instant)
+     * @since 5.2
+     */
+    @Override
+    public boolean clearExpired(final Instant instant) {
+        if (instant == null) {
+            return false;
+        }
+        lock.writeLock().lock();
+        try {
+            boolean removed = false;
+            for (final Iterator<Cookie> it = cookies.iterator(); it.hasNext(); ) {
+                if (it.next().isExpired(instant)) {
                     it.remove();
                     removed = true;
                 }
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
@@ -27,6 +27,7 @@
 
 package org.apache.hc.client5.http.cookie;
 
+import java.time.Instant;
 import java.util.Date;
 
 /**
@@ -77,8 +78,25 @@ public interface Cookie {
      * considered immutable. Changing it (e.g. using setTime()) could result
      * in undefined behaviour. Do so at your peril. </p>
      * @return Expiration {@link Date}, or {@code null}.
+     * @deprecated Use {{@link #getExpiryInstant()}}
      */
+    @Deprecated
     Date getExpiryDate();
+
+    /**
+     * Returns the expiration {@link Instant} of the cookie, or {@code null} if none exists.
+     * <p><strong>Note:</strong> the object returned by this method is
+     * considered immutable. Changing it (e.g. using setTime()) could result in undefined behaviour.
+     * Do so at your peril. </p>
+     *
+     * @return Expiration {@link Instant}, or {@code null}.
+     * @since 5.2
+     */
+    @SuppressWarnings("deprecated")
+    default Instant getExpiryInstant() {
+        final Date date = getExpiryDate();
+        return date != null ? Instant.ofEpochMilli(date.getTime()) : null;
+    }
 
     /**
      * Returns {@code false} if the cookie should be discarded at the end
@@ -119,13 +137,34 @@ public interface Cookie {
      * @param date Current time
      *
      * @return {@code true} if the cookie has expired.
+     * @deprecated Use {{@link #isExpired(Instant)}}
      */
+    @Deprecated
     boolean isExpired(final Date date);
+
+    /**
+     * Returns true if this cookie has expired.
+     *
+     * @param date Current time
+     * @return {@code true} if the cookie has expired.
+     * @since 5.2
+     */
+    @SuppressWarnings("deprecation")
+    default boolean isExpired(final Instant date) {
+        return isExpired(date != null ? new Date(date.toEpochMilli()) : null);
+    }
+
+    /**
+     * Returns creation time of the cookie.
+     * @deprecated Use {@link #getCreationInstant()}
+     */
+    @Deprecated
+    Date getCreationDate();
 
     /**
      * Returns creation time of the cookie.
      */
-    Date getCreationDate();
+    default Instant getCreationInstant() { return null;  }
 
     /**
      * Checks whether this Cookie has been marked as {@code httpOnly}.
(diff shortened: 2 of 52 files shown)


arturobernalg:
@michael-o 
Do you mind take a look. All the HttpClient API was migrated to Java 8 Time. Now I need to created the log of the method that I had to duplicated.
TY

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/BasicCookieStore.java
```diff
@@ -155,6 +156,34 @@ public boolean clearExpired(final Date date) {
...
     }
 
+    /**
+     * Removes all of {@link Cookie cookies} in this HTTP state
+     * that have expired by the specified {@link java.util.Date date}.
+     *
+     * @return true if any cookies were purged.
+     *
+     * @see Cookie#isExpired(Instant)
+     */
```
ok2c, line 166: @arturobernalg Please add `@since` tag.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
```diff
@@ -77,9 +78,22 @@
...
+    /**
+     * Returns the expiration {@link Instant} of the cookie, or {@code null}
+     * if none exists.
+     * <p><strong>Note:</strong> the object returned by this method is
+     * considered immutable. Changing it (e.g. using setTime()) could result
+     * in undefined behaviour. Do so at your peril. </p>
+     * @return Expiration {@link Instant}, or {@code null}.
+     * @since 5.2
+     */
+    default Instant getExpiryInstant() { return null; }
```
ok2c, line 95: @arturobernalg Please convert `Date` returned by deprecated `getExpiryDate()` to `Instance` here. Suppress deprecation warnings.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
```diff
@@ -119,14 +133,31 @@
...
+    @Deprecated
     boolean isExpired(final Date date);
 
+    /**
+     * Returns true if this cookie has expired.
+     * @param date Current time
+     *
+     * @return {@code true} if the cookie has expired.
+     */
+    default boolean isExpired(final Instant date) { return false; }
```
ok2c, line 147: @arturobernalg Use deprecated method here and suppress deprecation warnings.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
```diff
@@ -119,14 +133,31 @@
...
      * Returns creation time of the cookie.
+     * @deprecated Use {@link #getCreationInstant()}
      */
+    @Deprecated
     Date getCreationDate();
 
+    /**
+     * Returns creation time of the cookie.
+     */
+    default Instant getCreationInstant() { return null;  }
```
ok2c, line 159: @arturobernalg Use deprecated method here and suppress deprecation warnings.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/CookieStore.java
```diff
@@ -58,9 +59,19 @@
...
+    @Deprecated
     boolean clearExpired(Date date);
 
+    /**
+     * Removes all of {@link Cookie}s in this store that have expired by
+     * the specified {@link Instant}.
+     *
+     * @return true if any cookies were purged.
+     */
+    default boolean clearExpired(Instant date) { return false; }
```
ok2c, line 73: @arturobernalg Use deprecated method here and suppress deprecation warnings.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/SetCookie.java
```diff
@@ -48,10 +49,25 @@
...
+     * <p><strong>Note:</strong> the object returned by this method is considered
+     * immutable. Changing it (e.g. using setTime()) could result in undefined
+     * behaviour. Do so at your peril.</p>
+     *
+     * @param expiryDate the {@link Instant} after which this cookie is no longer valid.
+     *
+     * @see Cookie#getExpiryInstant()
+     *
+     */
+    default void setExpiryDate (Instant expiryDate) {}
```
ok2c, line 68: @arturobernalg Use deprecated method here and suppress deprecation warnings.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/cookie/LaxMaxAgeHandler.java
```diff
@@ -68,8 +68,8 @@ public void parse(final SetCookie cookie, final String value) throws MalformedCo
             } catch (final NumberFormatException e) {
                 return;
             }
-            final Date expiryDate = age >= 0 ? new Date(System.currentTimeMillis() + age * 1000L) :
-                    new Date(Long.MIN_VALUE);
+            final Instant expiryDate = age >= 0 ? Instant.now().plusSeconds(age * 1000L) :
```
ok2c, line 71: @arturobernalg This looks wrong. Should not it be `Instant.now().plusSeconds(age)` or `Instant.now().plusMillis(age * 1000L)`?

ok2c (changes requested):

arturobernalg:
Hi @ok2c 
Fixed all the remarks.
TY

ok2c:
@arturobernalg 

1. The `Cookie` interface is now coupled with `DateUtils` which is ugly. Please expand those utility methods and remove dependency on `DateUtils`. A bit of code duplication is justified in this particular case.
2. Is  `@SuppressWarnings("DeprecatedIsStillUsed")` better than shorter @SuppressWarnings("deprecation")? Just asking.
3. There are still several places where deprecated code is being used and causes deprecation warning at compile time. We need to get rid of those. I also propose that test cases for deprecated methods get dropped. No need to keep them if new methods pass the same test. Ideally there should be no deprecation warnings.

arturobernalg:
> 1. coupled



> @arturobernalg
> 
> 1. The `Cookie` interface is now coupled with `DateUtils` which is ugly. Please expand those utility methods and remove dependency on `DateUtils`. A bit of code duplication is justified in this particular case.
> 2. Is  `@SuppressWarnings("DeprecatedIsStillUsed")` better than shorter @SuppressWarnings("deprecation")? Just asking.
> 3. There are still several places where deprecated code is being used and causes deprecation warning at compile time. We need to get rid of those. I also propose that test cases for deprecated methods get dropped. No need to keep them if new methods pass the same test. Ideally there should be no deprecation warnings.

HI @ok2c 
All the remark fixed.
About using `@SuppressWarnings("DeprecatedIsStillUsed")`  instead of `@SuppressWarnings("deprecation")` it just because the IDE show me a warning of redundant with `deprecation `

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
```diff
@@ -77,9 +78,26 @@
      * considered immutable. Changing it (e.g. using setTime()) could result
      * in undefined behaviour. Do so at your peril. </p>
      * @return Expiration {@link Date}, or {@code null}.
+     * @deprecated Use {{@link #getExpiryInstant()}}
      */
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    @Deprecated
     Date getExpiryDate();
```
ok2c, line 85: @arturobernalg This looks wrong. Why are you suppressing deprecated method itself and not its usage where required for backward compatibility? Please correct it.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
```diff
@@ -77,9 +78,26 @@
...
+     * Returns the expiration {@link Instant} of the cookie, or {@code null} if none exists.
+     * <p><strong>Note:</strong> the object returned by this method is
+     * considered immutable. Changing it (e.g. using setTime()) could result in undefined behaviour.
+     * Do so at your peril. </p>
+     *
+     * @return Expiration {@link Instant}, or {@code null}.
+     * @since 5.2
+     */
+    default Instant getExpiryInstant() {
+        final Date date = getExpiryDate();
```
ok2c, line 97: @arturobernalg This is where you should be suppressing deprecation warnings.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
arturobernalg: Hi @ok2c 
The annotations @SuppressWarnings support method.  through this ways no need annotated where it is used.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
arturobernalg: idem above. but I can changed.  Please let me what do you prefer.

httpclient5/src/main/java/org/apache/hc/client5/http/cookie/Cookie.java
ok2c: > The annotations @SuppressWarnings support method. through this ways no need annotated where it is used.

@arturobernalg This is not a matter of taste or preference. Consumers of the library must be warned about API deprecation. The _only_ exception is the internal use of deprecated methods required for backward compatibility.

arturobernalg: Changed.
ty

ok2c: @arturobernalg I still see a lot of `@SuppressWarnings("DeprecatedIsStillUsed")` over deprecated methods. This is wrong. Please suppress the warning where the deprecated is used for backward compatibility, not the declaration itself.

arturobernalg: Solved.

ok2c (approved):
@arturobernalg Looks good to me now. Good work.

arturobernalg:
> @arturobernalg Looks good to me now. Good work.

Thank you @ok2c
I'm happy to help
TY

ok2c:
@arturobernalg Please close the associated JIRA.

arturobernalg:
> @arturobernalg Please close the associated JIRA.

done.
TY

PR #332, opened 2021-12-10
Expose method to check the cache is still active.

arturobernalg:
And reuse code for `this.active.compareAndSet(true, false);` validations.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ManagedHttpCacheStorage.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ManagedHttpCacheStorage.java
@@ -91,7 +91,7 @@ public class ManagedHttpCacheStorage implements HttpCacheStorage, Closeable {
     }
 
     private void ensureValidState() {
-        if (!this.active.get()) {
+        if (!isActive()) {
             throw new IllegalStateException("Cache has been shut down");
         }
     }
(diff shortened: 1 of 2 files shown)


SmallerCoder:
邮件已收到！我会尽快阅读！谢谢！

httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheStorage.java
```diff
@@ -94,4 +94,14 @@ void updateEntry(
      */
     Map<String, HttpCacheEntry> getEntries(Collection<String> keys) throws ResourceIOException;
 
+    /**
+     * Check if the cache is still active and has not shut down.
+     *
+     * @return {@code true} if the cache is active, otherwise return {@code false}.
+     * @since 5.2
+     */
+    default boolean isActive() {
```
ok2c, line 103: @arturobernalg Why adding this method to `HttpCacheStorage` when it clearly applies to `ManagedHttpCacheStorage` only and it makes no sense for all other implementations?

ok2c (changes requested):

httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheStorage.java
arturobernalg: > ManagedHttpCacheStorage

that make sense. Didn't see the other implementations. My bad.
Changed only for `ManagedHttpCacheStorage `

PR #333, opened 2021-12-10
Add private constructor so that no instances can be created

arturobernalg:
Private constructors prevent a class from being explicitly instantiated by its callers. Java Part 4 - Enforce noninstantiability


ok2c:
@arturobernalg This is a terrible idea to break binary compatibility just to hide a constructor. Who cares if someone creates an instance of this class?

arturobernalg:
> @arturobernalg This is a terrible idea to break binary compatibility just to hide a constructor. Who cares if someone creates an instance of this class?

Of course depend. Depend if we really want to avoid any change in those constants. could also apply to `HttpStatus`, for example.
But again, depend. In any case, I close the PR.
TY

PR #334, opened 2021-12-11
Method to check if the connection manager has been shut down and is closed or not.

arturobernalg:


ok2c:
@arturobernalg What about the async connection manager? What is the use case for this extra interface method? Do we really need it?

arturobernalg:
> @arturobernalg What about the async connection manager? What is the use case for this extra interface method? Do we really need it?

HI @ok2c 
Agree that we need the same implementation on async connection manager. IMO its logical and necessary have a method in connections manager class a checker of the connections state.
TY

ok2c:
> IMO its logical and necessary have a method in connections manager class a checker of the connections state.

@arturobernalg Why? What would be the use case?

arturobernalg:
> > IMO its logical and necessary have a method in connections manager class a checker of the connections state.
> 
> @arturobernalg Why? What would be the use case?

it seemed logic have a method to check if the connections is still valid. Make it private?

ok2c:
@arturobernalg you see, usually one closes the pool and then just throws it away. I do not see any use for this method outside the unit test. That is why I would not change the public APIs.

arturobernalg:
> . That is why I would not change the public APIs.

HI @ok2c 
I see,  now I understand your point.  no need it make it public. WDYT private?  I  changed it private method, feel free   close the PR if you believe there is no need it.
TY

ok2c:
@arturobernalg Feel free to keep them public in implementation classes or make them package private. Just make sure you make symmetric changes to the async conn manager as well.

arturobernalg:
> @arturobernalg Feel free to keep them public in implementation classes or make them package private. Just make sure you make symmetric changes to the async conn manager as well.

Hi @ok2c 
Just to double check. When you say async conn manager, you mean to `PoolingAsyncClientConnectionManager.java`

ok2c:
> Just to double check. When you say async conn manager, you mean to `PoolingAsyncClientConnectionManager.java`

@arturobernalg Yes.

arturobernalg:
> > Just to double check. When you say async conn manager, you mean to `PoolingAsyncClientConnectionManager.java`
> 
> @arturobernalg Yes.

Great @ok2c 
changes are applied
TY

PR #335, opened 2021-12-11
Avoid redundant validation of '\"'. Its checked before.

arturobernalg:
No need check again if the Chart is `\` or not.

`
else if (c == '\"') {
                foundEnd = true;
                offs++;
} else if (c != '\"' && !isControl(c)) {
`


ok2c:
@arturobernalg Not useful. Double-quote char has a special meaning in the context of the method  besides being a control char.

arturobernalg:
> @arturobernalg Not useful. Double-quote char has a special meaning in the context of the method besides being a control char.

HI @ok2c 
what I meant was the condition seem duplicated. We already check if the value is '\"'.  At the line 221.
Condition 'c != '\"'' is always 'true.' when reach line 224 
TY

PR #336, opened 2021-12-11
Safely remove unnecessary unboxing. No need it after Java 5 and later.

arturobernalg:
Sorry for so many stupid changes. I'm checking the cache module as was suggest in the mailing list.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AbstractSerializingAsyncCacheStorage.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AbstractSerializingAsyncCacheStorage.java
@@ -187,8 +187,8 @@ public abstract class AbstractSerializingAsyncCacheStorage<T, CAS> implements Ht
 
                                 @Override
                                 public void completed(final Boolean result) {
-                                    if (result.booleanValue()) {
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
looks good to me.

garydgregory:
-1 for me: unboxing makes it obvious where wrappers are used and forced one to consider when NPEs might be thrown.

arturobernalg:
> -1 for me: unboxing makes it obvious where wrappers are used and forces one to consider when NPEs might be thrown.

HI @garydgregory 
I thing your 

> unboxing

This Unboxing has't prevent the PNE.  At lest this one. The NPE will  produce before of this.

ok2c:
@arturobernalg Generally, not useful and applied to a single class, not consistently across the entire code base. 
Generally all stylistic changes are  _not_ useful unless they are backed by a style check or a commonly agreed upon style-guide applied consistently throughout the code base. Please find something more useful to work on. 

Agree with @garydgregory about autoboxing hiding possible NPEs.

arturobernalg:
Would be better validate the result. Not spect a NPE

` Args.notNull(result, "Completed Result");`

ok2c:
@arturobernalg Still not useful, unless done consistently throughout the entire code base.

PR #340, opened 2021-12-17
use isDebugEnabled avoid extra computations.

arturobernalg:


arturobernalg:
only when we need populate values.

ok2c:
@arturobernalg There is already `LOG.isDebugEnabled()` check around that debug log. Please take a look.

arturobernalg:
True.  my bad.

PR #343, opened 2022-01-05
Httpclient-2194 async retries not including body

JasonMathison:
https://issues.apache.org/jira/browse/HTTPCLIENT-2194

We found that our async retries were failing because there was no body being set.  Releasing the entityProducer resets the starting point to 0 and resolved the issue for us.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncHttpRequestRetryExec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncHttpRequestRetryExec.java
@@ -128,6 +128,9 @@ public final class AsyncHttpRequestRetryExec implements AsyncExecChainHandler {
             public void completed() {
                 if (state.retrying) {
                     scope.execCount.incrementAndGet();
+                    if (entityProducer != null && entityProducer.isRepeatable()) {
+                       entityProducer.releaseResources();
+                    }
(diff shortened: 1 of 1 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncHttpRequestRetryExec.java
```diff
@@ -128,6 +128,9 @@ public void handleInformationResponse(final HttpResponse response) throws HttpEx
             public void completed() {
                 if (state.retrying) {
                     scope.execCount.incrementAndGet();
+                    if (entityProducer != null && entityProducer.isRepeatable()) {
```
ok2c, line 131: @JasonMathison Damn. Good catch. However, I propose `AsyncEntityProducer#isRepeatable` check be dropped here. At this point if the entity producer is not null we should always call `AsyncEntityProducer#releaseResources`.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncHttpRequestRetryExec.java
JasonMathison: PR updated.

PR #347, opened 2022-01-13
Add new Constructs with the default configuration options.

arturobernalg:


ok2c:
@arturobernalg Why? Not useful.

arturobernalg:
> @arturobernalg Why? Not useful.

what can I say. seem logical to me.  We can create a EhcacheHttpCacheStorage with a default CacheConfig. Why don't give the way to do it automatically 

TY

PR #351, opened 2022-02-18
Async HttpClient builders to support custom IOSessionListener 

cda007:
Add the option to set an IOSessionListener on HttpAsyncClientBuilder

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/CircularRedirectException.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/CircularRedirectException.java
@@ -39,7 +39,7 @@ public class CircularRedirectException extends RedirectException {
      * Creates a new CircularRedirectException with a {@code null} detail message.
      */
     public CircularRedirectException() {
-        super(); 
+        super();
     }
 
     /**
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/CircularRedirectException.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/CircularRedirectException.java
@@ -39,7 +39,7 @@ public class CircularRedirectException extends RedirectException {
      * Creates a new CircularRedirectException with a {@code null} detail message.
      */
     public CircularRedirectException() {
-        super();
+        super(); 
     }
 
     /**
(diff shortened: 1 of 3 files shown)


SmallerCoder:
邮件已收到！我会尽快阅读！谢谢！

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
```diff
@@ -318,7 +320,15 @@ public final HttpAsyncClientBuilder setIOReactorConfig(final IOReactorConfig ioR
         this.ioReactorConfig = ioReactorConfig;
         return this;
     }
-
```
ok2c, line 321: @cda007 Could you please add `@since 5.2` tag?

ok2c (changes requested):
@cda007 Could you please also make symmetric changes to `H2AsyncClientBuilder`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
cda007: done

ok2c (approved):

cda007:
Hi Oleg,

Sorry for this. As I told you I am not familiar with this process and with git. Is there any specific pre-commit hook that would avoid to commit code not respecting rules ?

Regards,
Christophe

ok2c:
@cda007 This has nothing to do with git. This is straight Maven. `mvn clean verify` has got to pass for you locally with your changes. Style rules like no trailing blanks are enforced by the Maven style-check plugin.

httpclient5/src/main/java/org/apache/hc/client5/http/CircularRedirectException.java
```diff
@@ -39,7 +39,7 @@
      * Creates a new CircularRedirectException with a {@code null} detail message.
      */
     public CircularRedirectException() {
-        super();
+        super();
```
michael-o, line 42: Redundant space

michael-o (changes requested):

cda007:
Fixed. I use now to maven verification tool, should not occur anymore

PR #353, opened 2022-03-12
HTTPCLIENT-2209: Pass HttpContext to AsyncClientConnectionOperator

reta:
Pass HttpContext to AsyncClientConnectionOperator, change is non-breaking by introducing two default methods:

```
    default void upgrade(ManagedAsyncClientConnection conn, HttpHost host, Object attachment, HttpContext context) {
        upgrade(conn, host, attachment);
    }

    default Future<ManagedAsyncClientConnection> connect(
            ConnectionInitiator connectionInitiator,
            HttpHost host,
            SocketAddress localAddress,
            Timeout connectTimeout,
            Object attachment,
            HttpContext context,
            FutureCallback<ManagedAsyncClientConnection> callback) {
        return connect(connectionInitiator, host, localAddress, connectTimeout,
            attachment, callback);
    }
```

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/DefaultAsyncClientConnectionOperator.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/DefaultAsyncClientConnectionOperator.java
@@ -78,7 +78,7 @@ final class DefaultAsyncClientConnectionOperator implements AsyncClientConnectio
             final Object attachment,
             final FutureCallback<ManagedAsyncClientConnection> callback) {
         return connect(connectionInitiator, host, localAddress, connectTimeout,
-            attachment, callback, null);
+            attachment, null, callback);
     }
 
     @Override
@@ -88,8 +88,8 @@ final class DefaultAsyncClientConnectionOperator implements AsyncClientConnectio
             final SocketAddress localAddress,
             final Timeout connectTimeout,
             final Object attachment,
-            final FutureCallback<ManagedAsyncClientConnection> callback,
-            final HttpContext context) {
+            final HttpContext context,
+            final FutureCallback<ManagedAsyncClientConnection> callback) {
         Args.notNull(connectionInitiator, "Connection initiator");
         Args.notNull(host, "Host");
         final ComplexFuture<ManagedAsyncClientConnection> future = new ComplexFuture<>(callback);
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/nio/AsyncClientConnectionOperator.java
```diff
@@ -86,18 +113,35 @@
...
      * @since 5.2
      */
     default void upgrade(
             ManagedAsyncClientConnection conn,
             HttpHost host,
             Object attachment,
-            FutureCallback<ManagedAsyncClientConnection> callback) {
-        upgrade(conn, host, attachment);
+            FutureCallback<ManagedAsyncClientConnection> callback,
+            HttpContext context) {
```
reta, line 125: @ok2c I have changed the signature of this method since it has been introduce in 5.2 as well, hope it is OK, thanks!

reta:
@ok2c would very appreciate if you could take a look, thank you.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/DefaultAsyncClientConnectionOperator.java
```diff
@@ -141,15 +155,25 @@ public void upgrade(
...
+        upgrade(connection, host, attachment, null, context);
+    }
+
+    @Override
+    public void upgrade(
+            final ManagedAsyncClientConnection connection,
+            final HttpHost host,
+            final Object attachment,
+            final FutureCallback<ManagedAsyncClientConnection> callback,
+            final HttpContext context) {
```
ok2c, line 176: @reta For the sake of API consistency, could you please keep the callback as the last parameter? Otherwise looks good to me.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/nio/AsyncClientConnectionOperator.java
ok2c: @reta Sure thing. The connection operators are considered internal APIs, so technically we can do with them whatever we please.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/DefaultAsyncClientConnectionOperator.java
reta: 👍 Sure thing, thank you!

PR #354, opened 2022-03-15
StringBody: just write the bytes to the output stream

slackhappy:
StringBody's `writeTo` method looks like it was inspired by [FileBody's](https://github.com/apache/httpcomponents-client/blob/6a487ba686bafed9c642069e6046337af1bfd3a6/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FileBody.java#L80) writeTo:  

It creates an intermediate 4k buffer to transfer the content to the OutputStream.

however, this isn't necessary, since the `content` itself is already an array of bytes.

The content can be directly written to the output.

This was a significant source of allocations in a project I'm working on:
- for small form values, the 4k buffer was a significant over allocation
- for large form values, transferring the data using the buffer was a lot of work

```
Started [alloc] profiling
--- Execution profile ---
Total samples       : 155058
skipped             : 35740 (23.05%)

Frame buffer usage  : 10.567%

--- 1214394624 bytes (26.18%), 16734 samples
  [ 0] byte[]
  [ 1] org.apache.http.entity.mime.content.StringBody.writeTo
  [ 2] org.apache.http.entity.mime.AbstractMultipartForm.doWriteTo
  [ 3] org.apache.http.entity.mime.AbstractMultipartForm.writeTo
  [ 4] org.apache.http.entity.mime.MultipartFormEntity.writeTo
[SNIP]
```

I removed both the 4k buffer and the flush, similar to how [ByteArrayBody](https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/ByteArrayBody.java#L91) implements `writeTo`  under the assumption that, were those features an intended part of the specification, then `ByteArrayBody` would have implemented them as well.


httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/StringBody.java
```diff
@@ -70,13 +70,7 @@ public Reader getReader() {
     @Override
     public void writeTo(final OutputStream out) throws IOException {
         Args.notNull(out, "Output stream");
-        final InputStream in = new ByteArrayInputStream(this.content);
```
carterkozak, line 73: I think the `InputStream` import can be removed.

carterkozak (commented):
Great find, thank you for the contribution!

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/StringBody.java
slackhappy: oh good catch! will fix

carterkozak (approved):

ok2c:
@slackhappy Also merged to 5.1.x. Thank you for this contribution.

PR #355, opened 2022-03-22
Fix infinite recursion in SSLConnectionSocketFactory

rschmitt:


aarzare-boku:
Hi @rschmitt 
I faced an issue with my webclient where connection to an external host hangs forever. There are no exceptions thrown by httpclient.  I have set all required timeouts but still this happened. I was wondering where this infinite recursion issue could be the cause ?

OR 

How do I reproduce "infinite recursion in SSLConnectionSocketFactory" ?

rschmitt:
An infinite recursion would definitely throw `StackOverflowError`, not hang forever.

PR #358, opened 2022-04-06
Create daemon threads in InternalAbstractHttpAsyncClient

rhernandez35:
This thread pool is currently preventing applications with unclosed clients from exiting. Anyone who needs a client to finish a call before shutting down can continue to block on a non-daemon thread, but now someone needing an immediate shutdown is able to actually trigger one.


garydgregory:
Sounds reasonable.

PR #369, opened 2022-05-24
HTTPCLIENT-2218: Use Java 8 Base64 classes

j3graham:
Provides a minimal version of the commons-codec Base-64 classes for use with httpcomonents-client.

Code updates to use the new utility are also included, keeping the needed changes as small as possible.


httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthTokenExtractor.java
```diff
@@ -49,9 +47,9 @@ public String extract(final String challengeResponse) throws HttpException {
                 final String s = challengeResponse.substring(i + 1).trim();
                 try {
                     final byte[] credsRaw = s.getBytes(StandardCharsets.US_ASCII);
-                    final BinaryDecoder codec = new Base64();
+                    final Base64 codec = new Base64();
                     return new String(codec.decode(credsRaw), StandardCharsets.US_ASCII);
-                } catch (final DecoderException ex) {
```
j3graham, line 54: The DecoderException was caught because BinaryDecoder declared it, however the commons-codec Base64 implementation never actually threw it. Leaving handling here as a hint on how bad encoding could be handlerd

ok2c (commented):
@j3graham 
1. Could you please also drop dependency on Commons Codec? 
2. Is there any reason for deletion of `codeql-analysis.yml`?

ok2c:
@j3graham The PR actually looks pretty good. Why is it closed?

j3graham:
I opened a new one - https://github.com/apache/httpcomponents-client/pull/370

Closed this due to accidentally pulling in changes like the codeql one.

j3graham:
New one also drops the commons-codec dependency for production (my own test code obsessively compares the behaviour of the two)

PR #371, opened 2022-06-10
HTTPCLIENT-2221 Closing a classic response/entity allows connection reuse

carterkozak:
[HTTPCLIENT-2221](https://issues.apache.org/jira/browse/HTTPCLIENT-2221)

Previously, a partially consumed response body closed via
CloseableHttpResponse.close or HttpEntity.close would fully consume
remaining bytes (via close), however it would not release the
connection for reuse.
If CloseableHttpResponse.close was called, it would follow the close
with a discard/disconnect, however if only the entity was closed,
the connection would remain in a checked-out (leaked) state.

Now, we take advantage of the fact that closing a response stream
on any reusable connection is required to drain bytes on closure.
Failures are detected by writeTo and the stream returned by
getContent, so we can be confident that we will not return a
broken connection to the pool.

### DIFF
--- a/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestConnectionReuse.java
+++ b/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestConnectionReuse.java
@@ -84,6 +84,8 @@ public class TestConnectionReuse extends LocalServerTestBase {
             }
         }
 
+        // Expect leased connections to be returned
+        Assertions.assertEquals(0, this.connManager.getTotalStats().getLeased());
         // Expect some connection in the pool
         Assertions.assertTrue(this.connManager.getTotalStats().getAvailable() > 0);
     }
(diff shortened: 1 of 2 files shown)


httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestConnectionReuse.java
```diff
@@ -197,8 +204,10 @@ public void testReuseOfAbortedConnections() throws Exception {
             }
         }
 
-        // Expect zero connections in the pool
-        Assertions.assertEquals(0, this.connManager.getTotalStats().getAvailable());
+        // Expect leased connections to be returned
+        Assertions.assertEquals(0, this.connManager.getTotalStats().getLeased());
+        // Expect some connections in the pool
+        Assertions.assertTrue(this.connManager.getTotalStats().getAvailable() > 0);
```
carterkozak, line 210: Previously this test expected connections to be closed, however the response was fully drained and ready for reuse.

PR #376, opened 2022-06-29
Bumping versions of slf4j and log4j.

tsaarni:
Bumping to latest log4j binding as to remove CVE-2021-44832.
SLF4J v1.7.25 verson is from March 2017 and it would be nice to keep up with more recent versions, just to be save.


tsaarni:
xref https://github.com/apache/httpcomponents-core/pull/353

PR #378, opened 2022-07-05
Avoid duplicate redundant objects and use Singleton instead.

arturobernalg:


arturobernalg:
> So those are all stateless, thus threadsafe, by default?

Yes, otherwise didn’t have much sense the `STATELESS ` annotations.

PR #399, opened 2022-12-06
Bearer auth support

ok2c:
@michael-o Please do another pass.

### DIFF
--- a/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/Authenticator.java
+++ b/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/Authenticator.java
@@ -31,6 +31,13 @@ import org.apache.hc.core5.net.URIAuthority;
 
 public interface Authenticator {
 
+    /**
+     * @since 5.3
+     */
+    default AuthResult perform(URIAuthority authority, String requestUri, String credentials) {
+        return new AuthResult(authenticate(authority, requestUri, credentials));
+    }
+
     boolean authenticate(URIAuthority authority, String requestUri, String credentials);
 
     String getRealm(URIAuthority authority, String requestUri);
--- a/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
+++ b/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
@@ -27,12 +27,31 @@
 
 package org.apache.hc.client5.testing.auth;
 
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
 import org.apache.hc.client5.http.auth.StandardAuthScheme;
 import org.apache.hc.client5.http.utils.Base64;
+import org.apache.hc.core5.util.Args;
 
 public class BasicAuthenticationHandler extends AbstractAuthenticationHandler {
+
+    private final Charset charset;
+
+    /**
+     * @since 5.3
+     */
+    public BasicAuthenticationHandler(final Charset charset) {
+        this.charset = Args.notNull(charset, "Charset");
+    }
+
+    /**
+     * @deprecated Use {@link #BasicAuthenticationHandler(Charset)}
+     */
+    @Deprecated
+    public BasicAuthenticationHandler() {
+        this(StandardCharsets.US_ASCII);
+    }
 
     @Override
     String getSchemeName() {
@@ -43,7 +62,7 @@ public class BasicAuthenticationHandler extends AbstractAuthenticationHandler {
     String decodeChallenge(final String challenge) throws IllegalArgumentException {
         final byte[] bytes = challenge.getBytes(StandardCharsets.US_ASCII);
         final Base64 codec = new Base64();
-        return new String(codec.decode(bytes), StandardCharsets.US_ASCII);
+        return new String(codec.decode(bytes), charset);
     }
 
 }
--- a/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
+++ b/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
@@ -31,34 +31,19 @@ import java.nio.charset.StandardCharsets;
 
 import org.apache.hc.client5.http.auth.StandardAuthScheme;
 import org.apache.hc.client5.http.utils.Base64;
-import org.apache.hc.core5.http.HttpException;
-import org.apache.hc.core5.http.ProtocolException;
 
-/**
- * @deprecated Use {@link BasicAuthenticationHandler}.
- */
-@Deprecated
-public class BasicAuthTokenExtractor {
+public class BasicAuthenticationHandler extends AbstractAuthenticationHandler {
 
-    public String extract(final String challengeResponse) throws HttpException {
-        if (challengeResponse != null) {
-            final int i = challengeResponse.indexOf(' ');
-            if (i == -1) {
-                throw new ProtocolException("Invalid challenge response: " + challengeResponse);
-            }
-            final String schemeName = challengeResponse.substring(0, i);
-            if (schemeName.equalsIgnoreCase(StandardAuthScheme.BASIC)) {
-                final String s = challengeResponse.substring(i + 1).trim();
-                try {
-                    final byte[] credsRaw = s.getBytes(StandardCharsets.US_ASCII);
-                    final Base64 codec = new Base64();
-                    return new String(codec.decode(credsRaw), StandardCharsets.US_ASCII);
-                } catch (final IllegalArgumentException ex) {
-                    throw new ProtocolException("Malformed Basic credentials");
-                }
-            }
-        }
-        return null;
+    @Override
+    String getSchemeName() {
+        return StandardAuthScheme.BASIC;
+    }
+
+    @Override
+    String decodeChallenge(final String challenge) throws IllegalArgumentException {
+        final byte[] bytes = challenge.getBytes(StandardCharsets.US_ASCII);
+        final Base64 codec = new Base64();
+        return new String(codec.decode(bytes), StandardCharsets.US_ASCII);
     }
 
 }
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
@@ -1,0 +1,90 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.auth;
+
+import java.io.Serializable;
+import java.security.Principal;
+import java.util.Objects;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.util.Args;
+
+/**
+ * Opaque token {@link Credentials} usually representing a set of claims, often encrypted
+ * or signed. The JWT (JSON Web Token) is among most widely used tokens used at the time
+ * of writing.
+ *
+ * @since 5.3
+ */
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+public class BearerToken implements Credentials, Serializable {
+
+    private final String token;
+
+    public BearerToken(final String token) {
+        super();
+        this.token = Args.notBlank(token, "Token");
+    }
+
+    @Override
+    public Principal getUserPrincipal() {
+        return null;
+    }
+
+    /**
+     * @deprecated Do not use.
+     */
+    @Deprecated
+    @Override
+    public char[] getPassword() {
+        return null;
+    }
+
+    public String getToken() {
+        return token;
+    }
+
+    @Override
+    public int hashCode() {
+        return token.hashCode();
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o instanceof BearerToken) {
+            final BearerToken that = (BearerToken) o;
+            return Objects.equals(this.token, that.token);
+        }
+        return false;
+    }
+
+}
+
--- a/httpclient5-cache/pom.xml
+++ b/httpclient5-cache/pom.xml
@@ -28,7 +28,7 @@
   <parent>
     <groupId>org.apache.httpcomponents.client5</groupId>
     <artifactId>httpclient5-parent</artifactId>
-    <version>5.2.1-SNAPSHOT</version>
+    <version>5.2.2-SNAPSHOT</version>
   </parent>
   <artifactId>httpclient5-cache</artifactId>
   <name>Apache HttpClient Cache</name>
--- a/httpclient5-fluent/pom.xml
+++ b/httpclient5-fluent/pom.xml
@@ -28,7 +28,7 @@
   <parent>
     <groupId>org.apache.httpcomponents.client5</groupId>
     <artifactId>httpclient5-parent</artifactId>
-    <version>5.2.1-SNAPSHOT</version>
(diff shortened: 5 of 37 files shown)


michael-o:
Will take a look and test this week.

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/TokenBearerAuthenticationHandler.java
```diff
@@ -0,0 +1,44 @@
...
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.testing.auth;
+
+import org.apache.hc.client5.http.auth.StandardAuthScheme;
+
+public class TokenBearerAuthenticationHandler extends AbstractAuthenticationHandler {
```
michael-o, line 32: TokenBearer? Just Bearer. Since we don't do SpnegoTokenAuth...

michael-o (changes requested):
Don't add runtime exceptions to method signatures. They belong into Javadoc. See Effective Java from Joshua Bloch.

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    String getSchemeName() {
+        return StandardAuthScheme.BASIC;
+    }
+
+    @Override
+    String decodeChallenge(final String challenge) throws IllegalArgumentException {
+        final byte[] bytes = challenge.getBytes(StandardCharsets.US_ASCII);
+        final Base64 codec = new Base64();
+        return new String(codec.decode(bytes), StandardCharsets.US_ASCII);
```
michael-o, line 46: You assume that the bytes are always ASCII? What about UTF-8 these days?

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/Authenticator.java
```diff
@@ -31,6 +31,13 @@
 
 public interface Authenticator {
 
+    /**
+     * @since 5.3
+     */
+    default AuthResult auth(URIAuthority authority, String requestUri, String credentials) {
```
michael-o, line 37: Maybe `perform` is a better name? `auth` and `authenticate` seem to be to similar.

ok2c: @michael-o Agreed.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * @param userPrincipal the user principal
+     * @param token the token
+     *
+     * @see BasicUserPrincipal
+     * @see NTUserPrincipal
+     */
+    public BearerToken(final Principal userPrincipal, final String token) {
+        super();
+        this.principal = Args.notNull(userPrincipal, "User principal");
```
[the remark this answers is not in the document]
ok2c: @michael-o Why would not principal make sense here? Credentials are always associated with a particular security principal of some sort. Whether or not we can parse out the principal out of the token itself is a whole different story.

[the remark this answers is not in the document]
michael-o: This might be or not. The RFC does not indicate whether a subject is associated with. It solely talks about resource access.
Now let's look at this from client's perspective: I cannot obtain the principal since the token is supposed to be opaque to me and I have received it from my OIDC provider. Even with client credentials there is no guarantee that the final subject will correspond to the client_id? Should I supply a bogus principal?

michael-o:
> > Don't add runtime exceptions to method signatures. They belong into Javadoc. See Effective Java from Joshua Bloch.
> 
> @michael-o I must say I disagree with Joshua Bloch in this particular instance. There is no negative runtime impact from declaration of runtime exceptions in the throws clause. Essentially it is a matter of taste rather anything else. But so be it. I'll remove it.

I agree wth you, correct, there is no negative impact, but runtime exceptions denote a irrecoverable state mostly due to programming errors: shit in, shit out.

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
```diff
@@ -27,13 +27,27 @@
...
+    /**
+     * @since 5.3
+     */
+    public BasicAuthenticationHandler(final Charset charset) {
+        this.charset = charset != null ? charset : StandardCharsets.UTF_8;
+    }
+
+    public BasicAuthenticationHandler() {
+        this(StandardCharsets.US_ASCII);
+    }
```
michael-o, line 49: Is this for backwards compat? Since it looks inconsistent with no-arg and `null` arg. At least from a client perspective it is suprising.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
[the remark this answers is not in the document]
ok2c: @michael-o No matter what perspective, HTTP requests are always executed on someone's behalf and tokens is not something you find lying on the ground. They are acquired for a specific principal. It is bogus only if you make it bogus.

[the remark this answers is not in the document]
michael-o: I understand that, but what do you expect people to supply if they cannot deliver that?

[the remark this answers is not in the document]
ok2c: @michael-o What do you mean? They do not know what they are doing? HTTP requests are no farts. They do not happen once of a sudden and for no reason. They are always executed by some agent for a specific means. There is always at least the application with an application id of some sort.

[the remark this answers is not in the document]
michael-o: Yes, but my understanding is that the token is quite abstract compared to Basic.

michael-o:
Two more questions left, I will try this in production.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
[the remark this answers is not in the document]
ok2c: @michael-o It is, but despite being abstract it still represents a principal. Would you rather like it to be null?

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
ok2c: > Is this for backwards compat? 

@michael-o Yes, it is.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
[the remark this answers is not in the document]
michael-o: From my point it can be `null` since we don't have it [here](https://github.com/apache/httpcomponents-client/blob/5.3.x/httpclient5/src/main/java/org/apache/hc/client5/http/auth/KerberosCredentials.java) as well. (here it could ge source with the `GSSName`, but that is a different discussion)

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/auth/BasicAuthenticationHandler.java
michael-o: I see, the for consistency reasons the one-arg constructor should not allow null, encoding must be explicit then. Then the client can explicitly pass `UTF-8` and the rest will always be `US-ASCII`. We could also consider to use always `UTF-8` since ASCII input will always work.

ok2c:
@michael-o I have a feeling we are talking blondes vs brunettes here, but have it the way you like it. Please review

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
```diff
@@ -0,0 +1,120 @@
...
+        this(userName != null ? new BasicUserPrincipal(userName) : null, token);
+    }
+
+    @Override
+    public Principal getUserPrincipal() {
+        return this.principal;
+    }
+
+    public String getUserName() {
+        return this.principal.getName();
```
michael-o, line 81: This will cause a NPE when `principal` is null.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o instanceof BearerToken) {
+            final BearerToken that = (BearerToken) o;
+            return Objects.equals(this.principal, that.principal);
```
michael-o, line 109: Do you really consider both equal when the access tokens differ, but principal matches?

michael-o:
> @michael-o I have a feeling we are talking blondes vs brunettes here, but have it the way you like it. Please review

I like both, I am not picky :-D

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
ok2c: @michael-o I do. This is why I think null principal is sully, but so be it.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
```diff
@@ -0,0 +1,120 @@
...
+    private final String token;
+
+    /**
+     * The constructor with the user principal and token arguments.
+     *
+     * @param userPrincipal the user principal
+     * @param token the token
+     *
+     * @see BasicUserPrincipal
+     * @see NTUserPrincipal
```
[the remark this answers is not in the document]
michael-o: The `NTUserPrincipal` does not apply here.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
```diff
@@ -0,0 +1,120 @@
...
+        return null;
+    }
+
+    public String getToken() {
+        return token;
+    }
+
+    @Override
+    public int hashCode() {
+        return this.principal.hashCode();
```
michael-o, line 99: This one as well, no?

ok2c:
> @ok2c Can you explain why you use the `realm` parameter actively with Bearer altough rfc6749 does not mention it once? I only see these: https://datatracker.ietf.org/doc/html/rfc6749#section-11-2-2

Please see rfc6750, section 3.
https://www.rfc-editor.org/rfc/rfc6750

michael-o:
> > @ok2c Can you explain why you use the `realm` parameter actively with Bearer altough rfc6749 does not mention it once? I only see these: https://datatracker.ietf.org/doc/html/rfc6749#section-11-2-2
> 
> @michael-o Please see rfc6750, section 3. https://datatracker.ietf.org/doc/html/rfc6750#section-3

Indeed, thank you very much. This makes sense now!

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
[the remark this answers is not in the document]
ok2c: @michael-o Why?

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
michael-o: I can imagine the following: two clients (principals) have been granted to the same token to access a resource on behalf of the resource owner. How should equals behave here?

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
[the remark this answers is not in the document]
michael-o: Because the `NTUserPrincipal` consists of two components. This assumption cannot be done here.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
```diff
@@ -0,0 +1,95 @@
...
+        if (o instanceof BearerToken) {
+            final BearerToken that = (BearerToken) o;
+            return Objects.equals(this.token, that.token);
+        }
+        return false;
+    }
+
+    @Override
+    public String toString() {
+        return this.token;
```
michael-o, line 91: This is reasonable? Other credentials don't expose the secret via `toString()`.

michael-o (commented):
Much better now, tomorrow is live testing.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/BearerToken.java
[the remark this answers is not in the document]
ok2c: > This assumption cannot be done here.

@michael-o Why is that? An NT principal is just a composite of domain and user name. Anyhow, I removed the principal from `BearerToken`. So, forget it.

ok2c:
@michael-o what is it exactly that you want me to change now?

michael-o:
> @michael-o what is it exactly that you want me to change now?

Only the `toString()` is questionable, the rest is fine. Let me give it a shot tomorrow.

michael-o:
OK, I gave it a shot. OIDC provider is Ping Identity and the target host uses Spring Security. The good news, it works with no, invalid and valid tokens. The following needs to be clarified:
* The client has no chance to access response params from https://datatracker.ietf.org/doc/html/rfc6750#section-3
* In logging I see: "[main] DEBUG org.apache.hc.client5.http.impl.auth.BearerScheme - invalid_token (Invalid token)". I think this misses a bit context that this actually comes from the target server. For my taste, it is too sparse.
* Optional `error_uri` isn't logged for the developer. I don't know how valuable this information is, at least Spring Security refers to RFC 6750.

ok2c:
> The client has no chance to access response params from https://datatracker.ietf.org/doc/html/rfc6750#section-3

@michael-o What do you mean exactly? Do you mean the client or the caller? The client has access to the response parameters. if you want the caller to be able to have access to them as well, you need to tell me how that API should look like. At any rate, this should be a separate change-set and not a blocker for this PR.

>  think this misses a bit context that this actually comes from the target server. For my taste, it is too sparse.

What additional details would you like to have? Could you please specify how exactly the log entry should look like for your taste?

> Optional error_uri isn't logged for the developer. 

That I will do.

michael-o:
> > The client has no chance to access response params from https://datatracker.ietf.org/doc/html/rfc6750#section-3
> 
> @michael-o What do you mean exactly? Do you mean the client or the caller? The client has access to the response parameters. if you want the caller to be able to have access to them as well, you need to tell me how that API should look like. At any rate, this should be a separate change-set and not a blocker for this PR.

Caller. Yes, separate PR.
 
> > think this misses a bit context that this actually comes from the target server. For my taste, it is too sparse.
> 
> What additional details would you like to have? Could you please specify how exactly the log entry should look like for your taste?

I need to crunch on this...

ok2c:
@michael-o Out of curiosity, how long do you intend to block this PR?

michael-o:
> @michael-o Out of curiosity, how long do you intend to block this PR?

??? There are still unresolved issues:
* `toString()` unnecessarily exposes the token
* `error_uri` is not passed

ok2c:
> > @michael-o Out of curiosity, how long do you intend to block this PR?
> 
> ??? There are still unresolved issues:
> 

Issues?

>     * `toString()` unnecessarily exposes the token
> 

Why is this an issue, let alone a blocker?

>     * `error_uri` is not passed

Why is this an issue? Is there a MUST or SHOULD requirement? This is merely logging, which I am happy to tweak to your personal liking, but I still do not know _how_ exactly you want log entries to look like.

michael-o:
> > > @michael-o Out of curiosity, how long do you intend to block this PR?
> > 
> > 
> > ??? There are still unresolved issues:
> 
> Issues?
> 
> > ```
> > * `toString()` unnecessarily exposes the token
> > ```
> 
> Why is this an issue, let alone a blocker?

I bet there will be clowns complaining about this. I never wrote that this is a blocker.

> > ```
> > * `error_uri` is not passed
> > ```
> 
> Why is this an issue? Is there a MUST or SHOULD requirement? This is merely logging, which I am happy to tweak to your personal liking, but I still do not know _how_ exactly you want log entries to look like.

Look at your answer here: https://github.com/apache/httpcomponents-client/pull/399#issuecomment-1355338254

You wrote: I will do.

ok2c:
> > Why is this an issue, let alone a blocker?
> 
> I bet there will be clowns complaining about this. I never wrote that this is a blocker.
>

OK. This I can understand.
 
> Look at your answer here: [#399 (comment)](https://github.com/apache/httpcomponents-client/pull/399#issuecomment-1355338254)
> 
> You wrote: I will do.

Sure, I will, but I need to know _how_ it should look like.

michael-o:
> > > Why is this an issue, let alone a blocker?
> > 
> > 
> > I bet there will be clowns complaining about this. I never wrote that this is a blocker.
> 
> OK. This I can understand.
> 
> > Look at your answer here: [#399 (comment)](https://github.com/apache/httpcomponents-client/pull/399#issuecomment-1355338254)
> > You wrote: I will do.
> 
> Sure, I will, but I need to know _how_ it should look like.

I believe it can be added here: https://github.com/apache/httpcomponents-client/pull/399/files#diff-7a1d150c72f302b07cf402857b31bc209747308f250ef752f38bdc51472ffe48R101-R110

Maybe: `{error} ({error_description}; `{error_uri}`)

michael-o (approved):

PR #405, opened 2023-01-10
refactor: refactor bad smell InnerClassMayBeStatic

MartinWitt:
# Repairing Code Style Issues

## InnerClassMayBeStatic
Inner classes that do not reference their enclosing instances can be made static.
This prevents a common cause of memory leaks and uses less memory per instance of the class.

# Repairing Code Style Issues
* InnerClassMayBeStatic (3)


ok2c:
@MartinWitt This is a reasonable thing. However code style policies should be enforced through the style check [1] and not randomly. Would you be willing to adjust the style check definitions as well? 

[1] https://github.com/apache/httpcomponents-stylecheck

MartinWitt:
To my best knowledge, this rule isn't part of checkstyle, See https://checkstyle.sourceforge.io/config_coding.html. This bad smell was discovered by [Qodana ](https://www.jetbrains.com/qodana/) with the rule `InnerClassMayBeStatic`.

PR #411, opened 2023-02-06
Update Apache Commons Codec from 1.11 to 1.15.

alicejli:
There was a vulnerability found with commons-codec (https://sca.analysiscenter.veracode.com/vulnerability-database/security/sca/vulnerability/sid-22742/summary) that was fixed in v1.13.  
The latest version is 1.15 so I've updated the dependency to the latest.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
     <maven.compiler.target>1.6</maven.compiler.target>
     <httpcore.version>4.4.16</httpcore.version>
     <commons-logging.version>1.2</commons-logging.version>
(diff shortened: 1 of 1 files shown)


suztomo:
You talked about Java 6 compatibility. Can you write down findings here?

ok2c:
@alicejli We cannot upgrade HttpClient 4.5.x to the latest Commons Codec version due to Java 1.6 compatibility. Commons folks did not consider the security issue severe enough to warrant a release of Commons Codec compatible with Java 1.6.

alicejli:
Thank you for clarifying - that makes sense.

PR #414, opened 2023-02-23
Hygiene - remove unused private fields

jkmcl:


ok2c:
@jkmcl It cannot be done until 6.0.

PR #418, opened 2023-02-28
Fix issue with LLv6 literals in URLs

arturobernalg:
This pull request addresses issue LLv6 literals in URLs, where LLv6 literals in URLs are not properly handled by HttpClient due to a limitation in the Java standard method InetAddress.getAllByName(host). Specifically, when constructing a URL with an LLv6 literal, the % sign that prefixes the ZoneID must be quoted as %25 according to RFC 6874. However, HttpClient does not support quoted host literals.

To work around this limitation, this pull request modifies the SystemDefaultDnsResolver class to strip the IPv6 zone identifier from the host string before calling InetAddress.getAllByName(host). This allows HttpClient to handle LLv6 literals in URLs that conform to RFC 6874.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/SystemDefaultDnsResolver.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/SystemDefaultDnsResolver.java
@@ -40,7 +40,23 @@ public class SystemDefaultDnsResolver implements DnsResolver {
 
     @Override
     public InetAddress[] resolve(final String host) throws UnknownHostException {
-        return InetAddress.getAllByName(host);
+        try {
+            // Try resolving using the default resolver
+            return InetAddress.getAllByName(host);
+        } catch (final UnknownHostException e) {
+            // If default resolver fails, try stripping the IPv6 zone ID and resolving again
+            String strippedHost = null;
+            if (host.charAt(0) == '[') {
+                final int i = host.lastIndexOf('%');
+                if (i != -1) {
+                    strippedHost = host.substring(0, i) + "]";
+                }
+            }
+            if (strippedHost != null) {
+                return InetAddress.getAllByName(strippedHost);
+            }
+            throw e;
+        }
     }
 
     @Override
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/SystemDefaultDnsResolverTest.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/SystemDefaultDnsResolverTest.java
@@ -1,0 +1,67 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http;
+
+import org.junit.jupiter.api.Test;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SystemDefaultDnsResolverTest {
+
+
+    @Test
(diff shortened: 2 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/SystemDefaultDnsResolver.java
```diff
@@ -40,7 +40,14 @@ public class SystemDefaultDnsResolver implements DnsResolver {
 
     @Override
     public InetAddress[] resolve(final String host) throws UnknownHostException {
-        return InetAddress.getAllByName(host);
+        try {
```
ok2c, line 43: @arturobernalg This approach has a substantial drawback: DNS lookup failures completely unrelated to LLv6 will not generate two DNS queries. For some applications it may be not a big deal but for some it may. What is the problem with always calling `#stripsIPv6ZoneId` prior to calling `InetAddress#getAllByName`?

arturobernalg: The reason why we don't want to call stripsIPv6ZoneId all the time is that it involves additional processing and may not be necessary for hostnames that are not LLv6. This additional processing can cause a slight delay in the overall execution time of the resolve method.
In addition, if we always call stripsIPv6ZoneId before calling InetAddress.getAllByName, it means that we will be making two DNS queries for every hostname, even those that are not LLv6. This can result in unnecessary network traffic and may cause performance issues in certain environments.
Therefore, the approach of only calling stripsIPv6ZoneId when a UnknownHostException is caught and the hostname is an LLv6 address is a more targeted and efficient solution.

ok2c: @arturobernalg Repeating the same DNS query with the same input makes no sense and a lot of folks will be upset due to unnecessary execution costs. Please make sure the second query gets executed only if anything got stripped and the input is not the same

arturobernalg: @ok2c That's a valid point.
Changed.
TY

ok2c: @arturobernalg I suggest `#stripsIPv6ZoneId` be inlined as it is not being used anywhere else and extra hostname comparison be eliminated.

httpclient5/src/test/java/org/apache/hc/client5/http/SystemDefaultDnsResolverTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SystemDefaultDnsResolverTest {
+
+    @Test
+    void resolve() throws UnknownHostException {
+        final SystemDefaultDnsResolver resolver = SystemDefaultDnsResolver.INSTANCE;
+
+        final InetAddress[] result1 = resolver.resolve("example.com");
```
ok2c, line 43: @arturobernalg While extra test coverage is nice, in this particular case these tests introduce dependency on DNS infrastructure with public network access. These tests may fail when run in a private network where 'example.com' does not resolve. Please re-write the tests that do not require public network access or remote them.

ok2c (changes requested):

arturobernalg:
HI @ok2c 
Even though I don't fully agree with putting logic in the catch block, I made the change. I believe it's better to keep it separate."
TY

httpclient5/src/main/java/org/apache/hc/client5/http/SystemDefaultDnsResolver.java
```diff
@@ -40,7 +40,23 @@ public class SystemDefaultDnsResolver implements DnsResolver {
...
+        } catch (final UnknownHostException e) {
+            // If default resolver fails, try stripping the IPv6 zone ID and resolving again
+            String strippedHost = host;
+            if (host.charAt(0) == '[') {
+                final int i = host.lastIndexOf('%');
+                if (i != -1) {
+                    strippedHost = host.substring(0, i) + "]";
+                }
+            }
+            if (!strippedHost.equals(host)) {
```
ok2c, line 55: @arturobernalg The whole point of inlining the method was to avoid this extra comparison! Please use a boolean variable instead or set `strippedHost` by default.

ok2c: @arturobernalg In fact if you use null to signify a case when the original host is unchanged (not stripped) and no second DNS is not needed you can put this code into a separate method as before, if you like it better that way.

ok2c: @arturobernalg How about this?
```
String strippedHost = null;
if (host.charAt(0) == '[') {
  final int i = host.lastIndexOf('%');
  if (i != -1) {
    strippedHost = host.substring(0, i) + "]";
  }
}
if (strippedHost != null) {
  return InetAddress.getAllByName(strippedHost);
}
throw e;
```

arturobernalg: @ok2c , there is no significant difference between the two code snippets you provided. Both snippets attempt to strip the IPv6 zone ID from the host name if it is present and retry the resolution if the default resolver fails.
But I any case, I made the change

ok2c: @arturobernalg There is a biiiig difference in terms of performance between `if (strippedHost != null)` and `if (!strippedHost.equals(host))`

ok2c (approved):

PR #419, opened 2023-03-04
Fix calculation of freshness lifetime in HttpResponse

arturobernalg:
This pull request adds a fix for the calculateFreshnessLifetime method in the HttpResponseCache class. The method calculates the freshness lifetime of a response based on the headers in the response and follows the algorithm for calculating the freshness lifetime described in RFC 7234, section 4.2.1.

The method takes into account the s-maxage, max-age, Expires, and Date headers as follows:

- If the s-maxage directive is present in the Cache-Control header of the response, its value is used as the freshness lifetime for shared caches, which typically serve multiple users or clients.

- If the max-age directive is present in the Cache-Control header of the response, its value is used as the freshness lifetime for private caches, which serve a single user or client.

- If the Expires header is present in the response, its value is used as the expiration time of the response. The freshness lifetime is calculated as the difference between the expiration time and the time specified in the Date header of the response.

- If none of the above headers are present or if the calculated freshness lifetime is invalid, a default value of 5 minutes is returned.

The existing implementation had a bug where it did not handle the case where the max-age directive was present in the Cache-Control header but the s-maxage directive was not. In this case, the method would incorrectly return the default freshness lifetime of 5 minutes instead of using the max-age value as the freshness lifetime. This pull request fixes the bug by adding a check for the s-maxage directive and returning the max-age value if the s-maxage directive is not present.

### Motivation
The bug in the calculateFreshnessLifetime method could cause incorrect caching behavior, which could lead to degraded performance and unexpected results for clients. By fixing this bug, we ensure that the method correctly calculates the freshness lifetime of responses and that clients can rely on the cache to serve responses efficiently and accurately.

### Changes Made
To fix the bug, we added a check for the s-maxage directive in the Cache-Control header and updated the return value of the method to use the max-age value if the s-maxage directive is not present.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HeaderConstants.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HeaderConstants.java
@@ -39,6 +39,7 @@ public class HeaderConstants {
     public static final String PUT_METHOD = "PUT";
     public static final String DELETE_METHOD = "DELETE";
     public static final String TRACE_METHOD = "TRACE";
+    public static final String POST_METHOD = "POST";
 
     public static final String LAST_MODIFIED = "Last-Modified";
     public static final String IF_MATCH = "If-Match";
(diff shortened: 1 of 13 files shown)


arturobernalg:
closed in favor of https://github.com/apache/httpcomponents-client/pull/420 to 5.3.X branch

PR #420, opened 2023-03-04
Fix calculation of freshness lifetime in HttpResponse

arturobernalg:
This pull request adds a fix for the calculateFreshnessLifetime method in the HttpResponseCache class. The method calculates the freshness lifetime of a response based on the headers in the response and follows the algorithm for calculating the freshness lifetime described in RFC 7234, section 4.2.1.

The method takes into account the s-maxage, max-age, Expires, and Date headers as follows:

- If the s-maxage directive is present in the Cache-Control header of the response, its value is used as the freshness lifetime for shared caches, which typically serve multiple users or clients.

- If the max-age directive is present in the Cache-Control header of the response, its value is used as the freshness lifetime for private caches, which serve a single user or client.

- If the Expires header is present in the response, its value is used as the expiration time of the response. The freshness lifetime is calculated as the difference between the expiration time and the time specified in the Date header of the response.

- If none of the above headers are present or if the calculated freshness lifetime is invalid, a default value of 5 minutes is returned.

The existing implementation had a bug where it did not handle the case where the max-age directive was present in the Cache-Control header but the s-maxage directive was not. In this case, the method would incorrectly return the default freshness lifetime of 5 minutes instead of using the max-age value as the freshness lifetime. This pull request fixes the bug by adding a check for the s-maxage directive and returning the max-age value if the s-maxage directive is not present.

Motivation

The bug in the calculateFreshnessLifetime method could cause incorrect caching behavior, which could lead to degraded performance and unexpected results for clients. By fixing this bug, we ensure that the method correctly calculates the freshness lifetime of responses and that clients can rely on the cache to serve responses efficiently and accurately.

Changes Made

To fix the bug, we added a check for the s-maxage directive in the Cache-Control header and updated the return value of the method to use the max-age value if the s-maxage directive is not present.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
@@ -1,0 +1,191 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl.cache;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.http.FormattedHeader;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.util.Args;
+import org.apache.hc.core5.util.CharArrayBuffer;
+import org.apache.hc.core5.util.Tokenizer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.BitSet;
+
+
+/**
+ * A parser for the HTTP Cache-Control header that can be used to extract information about caching directives.
+ * <p>
+ * This class is thread-safe and has a singleton instance ({@link #INSTANCE}).
+ * </p>
+ * <p>
+ * The {@link #parse(Header)} method takes an HTTP header and returns a {@link CacheControl} object containing
+ * the relevant caching directives. The header can be either a {@link FormattedHeader} object, which contains a
+ * pre-parsed {@link CharArrayBuffer}, or a plain {@link Header} object, in which case the value will be parsed and
+ * stored in a new {@link CharArrayBuffer}.
+ * </p>
+ * <p>
+ * This parser only supports two directives: "max-age" and "s-maxage". If either of these directives are present in the
+ * header, their values will be parsed and stored in the {@link CacheControl} object. If both directives are
+ * present, the value of "s-maxage" takes precedence.
+ * </p>
+ * <p>
+ * Example usage:
+ * </p>
+ * <pre>
+ * Header header = new BasicHeader("Cache-Control", "max-age=60, s-maxage=120");
+ * CacheControlHeaderParser parser = CacheControlHeaderParser.INSTANCE;
+ * CacheControlHeader cacheControlHeader = parser.parse(header);
+ * long maxAgeSeconds = cacheControlHeader.getMaxAgeSeconds(); // returns 120
+ * </pre>
+ */
+@Internal
+@Contract(threading = ThreadingBehavior.SAFE)
+class CacheControlHeaderParser {
+
+    /**
+     * The character used to indicate a parameter's value in the Cache-Control header.
+     */
+    private final static char EQUAL_CHAR = '=';
+
+    /**
+     * The singleton instance of this parser.
+     */
+    public static final CacheControlHeaderParser INSTANCE = new CacheControlHeaderParser();
+
+    /**
+     * The logger for this class.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(CacheControlHeaderParser.class);
+
+
+    /**
+     * The delimiter character used to separate values within a caching directive.
+     */
+    private static final char VALUE_DELIMITER = '=';
+
+    /**
+     * The set of characters that can delimit a token in the header.
+     */
+    private static final BitSet TOKEN_DELIMS = Tokenizer.INIT_BITSET(VALUE_DELIMITER, ',');
+
+    /**
+     * The set of characters that can delimit a value in the header.
+     */
+    private static final BitSet VALUE_DELIMS = Tokenizer.INIT_BITSET(VALUE_DELIMITER, ',');
+
+    /**
+     * The token parser used to extract values from the header.
+     */
+    private final Tokenizer tokenParser;
+
+    /**
+     * Constructs a new instance of this parser.
+     */
+    protected CacheControlHeaderParser() {
+        super();
+        this.tokenParser = Tokenizer.INSTANCE;
+    }
+
+    /**
+     * Parses the specified header and returns a new {@link CacheControl} instance containing the relevant caching
+     * <p>
+     * directives.
+     *
+     * <p>The returned {@link CacheControl} instance will contain the values for "max-age" and "s-maxage" caching
+     * directives parsed from the input header. If the input header does not contain any caching directives or if the
+     * <p>
+     * directives are malformed, the returned {@link CacheControl} instance will have default values for "max-age" and
+     * <p>
+     * "s-maxage" (-1).</p>
+     *
+     * @param header the header to parse, cannot be {@code null}
+     * @return a new {@link CacheControl} instance containing the relevant caching directives parsed from the header
+     * @throws IllegalArgumentException if the input header is {@code null}
+     */
+    public final CacheControl parse(final Header header) {
+        Args.notNull(header, "Header");
+
+        long maxAge = -1;
+        long sharedMaxAge = -1;
+
+        //final CacheControl cacheControl = new CacheControl();
+        final CharArrayBuffer buffer;
+        final Tokenizer.Cursor cursor;
+        if (header instanceof FormattedHeader) {
+            buffer = ((FormattedHeader) header).getBuffer();
+            cursor = new Tokenizer.Cursor(((FormattedHeader) header).getValuePos(), buffer.length());
+        } else {
+            final String s = header.getValue();
+            if (s == null) {
+                return new CacheControl(maxAge, sharedMaxAge);
+            }
+            buffer = new CharArrayBuffer(s.length());
+            buffer.append(s);
+            cursor = new Tokenizer.Cursor(0, buffer.length());
+        }
+
+        while (!cursor.atEnd()) {
+            final String name = tokenParser.parseToken(buffer, cursor, TOKEN_DELIMS);
+            if (name.isEmpty()) {
+                return new CacheControl(maxAge, sharedMaxAge);
+            }
+            if (cursor.atEnd()) {
+                return new CacheControl(maxAge, sharedMaxAge);
+            }
+            final int valueDelim = buffer.charAt(cursor.getPos());
+            cursor.updatePos(cursor.getPos() + 1);
+            if (valueDelim != EQUAL_CHAR) {
+                continue;
+            }
+            final String value = tokenParser.parseValue(buffer, cursor, VALUE_DELIMS);
+
+            if (!cursor.atEnd()) {
+                cursor.updatePos(cursor.getPos() + 1);
+            }
+
+            try {
+                if (name.equalsIgnoreCase("s-maxage")) {
+                    sharedMaxAge = Long.parseLong(value);
+                } else if (name.equalsIgnoreCase("max-age")) {
+                    maxAge = Long.parseLong(value);
+                }
+            } catch (final NumberFormatException e) {
+                // skip malformed directive
+                if (LOG.isWarnEnabled()) {
+                    LOG.warn("Header {} was malformed: {}", name, value);
+                }
+            }
+        }
+        return new CacheControl(maxAge, sharedMaxAge);
+    }
+}
+
+
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
@@ -26,7 +26,9 @@
  */
 package org.apache.hc.client5.http.impl.cache;
 
+import java.time.Duration;
 import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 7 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -50,6 +51,29 @@
...
+     * The default freshness duration for a cached object, in seconds.
+     *
+     * <p>This constant is used to set the default value for the freshness lifetime of a cached object.
+     * When a new object is added to the cache, it will be assigned this duration if no other duration
+     * is specified.</p>
+     *
+     * <p>By default, this value is set to 300 seconds (5 minutes). Applications can customize this
+     * value as needed.</p>
+     */
+    private static final long DEFAULT_FRESHNESS_DURATION = 300L;
```
ok2c, line 64: @arturobernalg This is actually the case where the use of `java.time.Duration` instead of plain `long` might be beneficial.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -320,4 +354,87 @@ private boolean from1_0Origin(final HttpResponse response) {
...
+     * </ul>
+     *
+     * <p>
+     * Note that caching is a complex topic and cache control directives may interact with each other in non-trivial ways.
+     * This method provides a basic implementation of the freshness lifetime calculation algorithm and may not be suitable
+     * for all use cases. Developers should consult the HTTP caching specifications for more information and consider
+     * implementing additional caching mechanisms as needed.
+     * </p>
+     *
+     * @param response the HTTP response for which to calculate the freshness lifetime
```
ok2c, line 383: @arturobernalg We stopped referencing specific RFCs in javadocs. Those URLs tend to change and also need to be updated every time the code base gets revised against a newer set of RFCs.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -320,4 +354,87 @@ private boolean from1_0Origin(final HttpResponse response) {
...
+            return DEFAULT_FRESHNESS_DURATION; // 5 minutes
+        }
+
+        final String cacheControlValue = cacheControl.getValue();
+        final String[] headerValues = cacheControlValue.split(",");
+
+        long maxAge = -1;
+        long sharedMaxAge = -1;
+
+        for (final String headerValue : headerValues) {
```
ok2c, line 407: @arturobernalg Please use `org.apache.hc.core5.util.Tokenizer` based parsers from HttpCore to parse header elements, see `AuthChallengeParser` for an example.
Please also consider what is going to happen if your code has to parse something like `max-age=boom`. Malformed header values must not cause a runtime exception.

arturobernalg:
HI @ok2c 
All remarks changed and Refactor calculateFreshnessLifetime() to use Tokenizer for parsing Cache-Control header values.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
ok2c: @arturobernalg This is even worse then before. You should do all the parsing with `Tokenizer` and there should be no `split`, `trim` or `substring` and there should be almost no intermediate garbage. Please see how `Tokenizer`is being used to parse other complex headers, like `Set-Cookie` for instance. Please take a look at `RFC6265CookieSpec#parse`.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -320,4 +360,100 @@ private boolean from1_0Origin(final HttpResponse response) {
...
+        if (sharedMaxAge != -1) {
+            return Duration.ofSeconds(sharedMaxAge);
+        } else if (maxAge != -1) {
+            return Duration.ofSeconds(maxAge);
+        }
+
+        // Check if Expires is present and use its value minus the value of the Date header
+        Instant expiresInstant = null;
+        Instant dateInstant = null;
+        if (response.containsHeader(HttpHeaders.EXPIRES)) {
```
ok2c, line 443: @arturobernalg Please do not do that. Please do not use two method calls where a single method call and a null check should perfectly suffice.

ok2c (changes requested):

arturobernalg:
@ok2c  WDYT?

arturobernalg:
> @arturobernalg That is much better, but you can still improve it further. Please note that your method executes the same logic twice inside the loop. Please optimize it. You may also want to create a separate test case just for the parser. Once this is done I will suggest further improvements.

HI @ok2c 
what about now.
TY

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -0,0 +1,211 @@
...
+            }
+            if (cursor.atEnd()) {
+                if (LOG.isWarnEnabled()) {
+                    LOG.warn("Invalid 'Cache-Control' header value: '{}'. '{}' must be followed by a valid integer value.", header.getValue(), name);
+                }
+                return cacheControlHeader;
+            }
+            final int valueDelim = buffer.charAt(cursor.getPos());
+            cursor.updatePos(cursor.getPos() + 1);
+            if (valueDelim != EQUAL_CHAR) {
```
arturobernalg, line 166: @ok2c 
This is the only part that I'm not sure about. I don't think these conditions will ever happen. Otherwise, they would be controlled by the previous one.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -0,0 +1,211 @@
...
+        final CacheControlHeader cacheControlHeader = new CacheControlHeader();
+        final CharArrayBuffer buffer;
+        final Tokenizer.Cursor cursor;
+        if (header instanceof FormattedHeader) {
+            buffer = ((FormattedHeader) header).getBuffer();
+            cursor = new Tokenizer.Cursor(((FormattedHeader) header).getValuePos(), buffer.length());
+        } else {
+            final String s = header.getValue();
+            if (s == null) {
+                if (LOG.isWarnEnabled()) {
```
ok2c, line 143: @arturobernalg No need for a warning here.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -0,0 +1,211 @@
...
+                return cacheControlHeader;
+            }
+            buffer = new CharArrayBuffer(s.length());
+            buffer.append(s);
+            cursor = new Tokenizer.Cursor(0, buffer.length());
+        }
+
+        while (!cursor.atEnd()) {
+            final String name = tokenParser.parseToken(buffer, cursor, TOKEN_DELIMS);
+            if (name.isEmpty()) {
```
ok2c, line 155: @arturobernalg Make sure your parser will not choke on `Cache-Control: , = blah,` value. Please add this test case.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -0,0 +1,211 @@
...
+                continue;
+            }
+            if (cursor.atEnd()) {
+                if (LOG.isWarnEnabled()) {
```
ok2c, line 159: @arturobernalg Is this really so? What about `Cache-Control: no-cache` and similar? There can be several directives that do not have a parameter. Please remove or rewrite this warning.

ok2c (changes requested):

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
arturobernalg: its already cover here,
we really not need that validations
`   @Test
    public void testParseCookieEmptyValue() {
        final Header header = new BasicHeader("Cache-Control", "max-age=;");
        assertThrows(MalformedCookieException.class, () -> parser.parse(header));
    }`

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
arturobernalg: remove it

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
arturobernalg: handle by while (!cursor.atEnd())

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
arturobernalg: remove it

arturobernalg:
all remarks solved

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
ok2c: @arturobernalg Are you sure. Anyway I cannot see a test case for it.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -0,0 +1,205 @@
...
+            if (name.isEmpty()) {
+                continue;
+            }
+            if (cursor.atEnd()) {
+                return cacheControlHeader;
+            }
+            final int valueDelim = buffer.charAt(cursor.getPos());
+            cursor.updatePos(cursor.getPos() + 1);
+            if (valueDelim != EQUAL_CHAR) {
+                if (LOG.isWarnEnabled()) {
```
ok2c, line 161: @arturobernalg Do we really need this warning? Is the header really invalid or is it a directive without a parameter?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -0,0 +1,205 @@
...
+     * Sets the value of the specified Cache-Control header field in the given CacheControlHeader object.
+     * Returns true if the header field is not recognized, or if the value is malformed and throws a MalformedCookieException.
+     *
+     * @param cacheControlHeader the CacheControlHeader object to set the header field in
+     * @param name               the name of the Cache-Control header field to set
+     * @param value              the value of the Cache-Control header field to set
+     * @return true if the header field is not recognized
+     * @throws MalformedCookieException if the value is malformed and cannot be parsed as a long integer
+     */
+    private boolean setHeader(final CacheControlHeader cacheControlHeader, final String name, final String value) throws MalformedCookieException {
```
ok2c, line 189: @arturobernalg `MalformedCookieException` is clearly wrong here. This header has nothing to do with cookies. I also think the right thing to do here is to ignore values that cannot be interpreted as valid instead of throwing an exception. A log message at the DEBUG priority might be all right.

Another thing. Please  rename `CacheControlHeader` to `CacheControl` and make it nice and *immutable*. For that you may need to inline this method and create `CacheControl` instance at the last moment when exiting the `#parse` method.

arturobernalg:
@ok2c done?  the only thing that I didn't now what you mean was -->For that you may need to inline this method and create CacheControl instance at the last moment when exiting the #parse method.
but in anywise, seems everything ok.

ok2c:
> @ok2c done? 

@arturobernalg  No, not quite. You added `@Immutable` to `CacheControl` but it is not immutable. Its instance variables are not final. You need to make sure the instances of `CacheControl` cannot be mutated after construction.

arturobernalg:
> > @ok2c done?
> 
> @arturobernalg No, not quite. You added `@Immutable` to `CacheControl` but it is not immutable. Its instance variables are not final. You need to make sure the instances of `CacheControl` cannot be mutated after construction.

HI @ok2c

You are correct, adding @Immutable to CacheControl does not make it immutable by itself. I have made the necessary changes to make CacheControl immutable.

I declared the instance variables maxAge and sharedMaxAge as final, removed their setter methods, and modified the constructor to accept the values for these variables and initialize them with these values.

With these changes, instances of CacheControl cannot be mutated after construction, making it truly immutable. Please let me know if you have any further concerns or feedback.

ok2c (approved):
@arturobernalg Looks good now. I will commit your changes with some minor tweaks later today.

ok2c (changes requested):
@arturobernalg I asked you to look at this test case, did not I? What do you think is going to happen?
```
    @Test
    public void testGarbage() {
        final Header header = new BasicHeader("Cache-Control", ",,= blah,");
        final CacheControl cacheControl = parser.parse(header);
        assertEquals(0L, cacheControl.getMaxAge());
    }
```

ok2c:
@arturobernalg Please also remove `PARAM_DELIMITER` from the `CacheControlHeaderParser`. It is not applicable to the cache control directives. They do not have parameters.

arturobernalg:
@ok2c 

Tackled infinite loop caused by garbage input in testGarbage() test case
Removed unnecessary PARAM_DELIMITER from CacheControlHeaderParser

ok2c (approved):

ok2c:
@arturobernalg There is a simpler, nicer fix. Wanna try to find it?

Bonus question: what should the value of max-age be with the header like `Cache-Control: no-cache, max-age=123`?

arturobernalg:
> @arturobernalg There is a simpler, nicer fix. Wanna try to find it?
> 
> Bonus question: what should the value of max-age be with the header like `Cache-Control: no-cache, max-age=123`?

I just rt

> @arturobernalg There is a simpler, nicer fix. Wanna try to find it?
> 
> Bonus question: what should the value of max-age be with the header like `Cache-Control: no-cache, max-age=123`?
For what I understand a cache MUST NOT use a response to satisfy a subsequent request without successful revalidation with the origin server if the Cache-Control header field includes the "no-cache" directive. However, it MAY use the response to satisfy a subsequent request if the Cache-Control header field contains a "max-age" directive with a non-zero value.

In the given header "Cache-Control: no-cache, max-age=123", the "no-cache" directive takes precedence over the "max-age" directive, but the presence of "max-age" means that the response can be used to satisfy a subsequent request after 123 seconds have elapsed, assuming successful revalidation with the origin server. So the value of max-age should be 123 seconds.

Regarding the simple fix, we can return the CacheControl object since we have already extracted the relevant caching directives from the Cache-Control header by this point.

ok2c:
@arturobernalg I am going to merge the PR as is and make some small changes to it. The current handling of with different (conflicting) cache control directives should be fixed, however. In the simplest form the `max-age` value should be ignored if the `no-cache` directive is present.

ok2c:
@arturobernalg I now understood your response better. It turns out that now the same `Cache-Control` header gets parsed twice, once by `#isExplicitlyNonCacheable` and again by `#calculateFreshnessLifetime`, which is bad and must be fixed. The parsing code should be extracted from `#calculateFreshnessLifetime` and enhanced to include the main cache control directive that the `#isExplicitlyNonCacheable` could use to make its decision:

```
        if (isExplicitlyNonCacheable(response)) {
            LOG.debug("Response is explicitly non-cacheable");
            return false;
        }

        // calculate freshness lifetime
        final Duration freshnessLifetime = calculateFreshnessLifetime(response);
        if (freshnessLifetime.isNegative() || freshnessLifetime.isZero()) {
            if (LOG.isDebugEnabled()) {
                LOG.debug("Freshness lifetime is invalid");
            }
            return false;
        }
```

arturobernalg:
> LOG.debug

@ok2c  I'll take a look

PR #422, opened 2023-03-11
 Implement the stale-if-error request cache directive from RFC 5861

arturobernalg:
This Pull request implements the stale-if-error request cache directive from RFC 5861 in the BasicHttpCache class. The stale-if-error directive allows clients to receive a stale cache entry when a request results in an error, rather than an error response.

To implement this, we added the cacheConfig and cacheValidityPolicy fields to the BasicHttpCache class and modified the getCacheEntry method to check if the staleIfError directive is enabled and return the stale cache entry if it is. Additionally, we updated the CacheConfig class to add a staleIfErrorEnabled boolean field to control this behavior.

Here's a summary of the changes:

Added cacheConfig and cacheValidityPolicy fields to BasicHttpCache
Modified getCacheEntry to check for the staleIfError directive and return the stale cache entry if enabled
Added staleIfErrorEnabled boolean field to CacheConfig
Please let me know if there are any further changes or modifications required.


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExecBase.java
```diff
@@ -370,4 +372,33 @@ void storeRequestIfModifiedSinceFor304Response(final HttpRequest request, final
...
+     *
+     * @param entry the cached HTTP response entry to check
+     * @return {@code true} if a stale response can be served in case of an error status code, {@code false} otherwise
+     */
+    boolean isStaleIfErrorEnabled(final HttpCacheEntry entry) {
+        // Check if the stale-while-revalidate extension is enabled
+        if (cacheConfig.isStaleIfErrorEnabled()) {
+            // Check if the cached response has an error status code
+            final int statusCode = entry.getStatus();
+            if (statusCode >= HttpStatus.SC_INTERNAL_SERVER_ERROR && statusCode <= HttpStatus.SC_GATEWAY_TIMEOUT) {
```
ok2c, line 390: @arturobernalg Looks mostly good. 
1. Could you please move this logic to `ResponseCachingPolicy`? It belongs there.
2. There are multiple places in `ResponseCachingPolicy`where `Cache-Control` header gets parsed. This looks wasteful. Could you please see if  the `CacheControl` could be produced once and re-used by multiple policy methods? You may need to pull the parsing logic from the `ResponseCachingPolicy` into the caching exec interceptor (both classic and async).

ok2c (changes requested):

ok2c (approved):
@arturobernalg Looks good. I can merge this change-set unless you also want to add handling of I/O errors in the same PR.

arturobernalg:
> @arturobernalg Looks good. I can merge this change-set unless you also want to add handling of I/O errors in the same PR.

@ok2c 
Could you please clarify which part of the codebase you are referring to for this change?

ok2c (approved):

ok2c:
> > @arturobernalg Looks good. I can merge this change-set unless you also want to add handling of I/O errors in the same PR.
> 
> @ok2c Could you please clarify which part of the codebase you are referring to for this change?

```
   The stale-if-error HTTP Cache-Control extension allows a cache to
   return a stale response when an error -- e.g., a 500 Internal Server
   Error, a network segment, or DNS failure -- is encountered, rather
   than returning a "hard" error.  This improves availability.
```
The change-set introduces `stale-if-error` support for HTTP status 5xx but not for network failures. You do not have to do it. I am going to merge the PR anyway. I am just wondering if you were planning to work on it as well.

arturobernalg:
> > > @arturobernalg Looks good. I can merge this change-set unless you also want to add handling of I/O errors in the same PR.
> > 
> > 
> > @ok2c Could you please clarify which part of the codebase you are referring to for this change?
> 
> ```
>    The stale-if-error HTTP Cache-Control extension allows a cache to
>    return a stale response when an error -- e.g., a 500 Internal Server
>    Error, a network segment, or DNS failure -- is encountered, rather
>    than returning a "hard" error.  This improves availability.
> ```
> 
> The change-set introduces `stale-if-error` support for HTTP status 5xx but not for network failures. You do not have to do it. I am going to merge the PR anyway. I am just wondering if you were planning to work on it as well.

I see. i'll take a look this afternoon. I'll do the change in another PR
TY

PR #423, opened 2023-03-11
Enhances ContentResponseHandler to be fault-tolerant 

arturobernalg:
This pull request enhances ContentResponseHandler to be fault-tolerant by adding the ability to handle responses with null or empty bodies without throwing exceptions. If the response was successful (a 2xx status code), the Content object is returned. If no response body exists, Content.NO_CONTENT is returned.

This implementation of ContentResponseHandler overrides its methods to handle the response entity and transform it into the actual response object. This is designed to be efficient and use minimal memory.

This enhancement to ContentResponseHandler is intended to be used in scenarios where the response body may be null or empty, but the caller still needs to determine whether the response was successful or not. For example, when testing an HTTP API, the response body may be empty but the status code can still indicate whether the request was successful or not.

### DIFF
--- a/httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/ContentResponseHandler.java
+++ b/httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/ContentResponseHandler.java
@@ -28,7 +28,9 @@ package org.apache.hc.client5.http.fluent;
 
 import java.io.IOException;
 
+import org.apache.hc.client5.http.HttpResponseException;
 import org.apache.hc.client5.http.impl.classic.AbstractHttpClientResponseHandler;
+import org.apache.hc.core5.http.ClassicHttpResponse;
 import org.apache.hc.core5.http.ContentType;
 import org.apache.hc.core5.http.HttpEntity;
 import org.apache.hc.core5.http.io.entity.EntityUtils;
@@ -39,10 +41,15 @@ import org.apache.hc.core5.http.io.entity.EntityUtils;
  * to {@link Content} instances.
  *
  * @see Content
- *
  * @since 4.4
  */
 public class ContentResponseHandler extends AbstractHttpClientResponseHandler<Content> {
+
+
+    /**
+     * The maximum length of the exception message, to avoid excessive memory usage.
+     */
+    private static final int MAX_MESSAGE_LENGTH = 256;
 
     @Override
     public Content handleEntity(final HttpEntity entity) throws IOException {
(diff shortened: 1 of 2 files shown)


arturobernalg:
> @arturobernalg Please do it simpler and just enhance `ContentResponseHandler` instead of creating a whole new handler.

done. add it into `ContentResponseHandler`

httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/ContentResponseHandler.java
```diff
@@ -44,11 +46,63 @@
...
+     * @throws HttpResponseException if the response was unsuccessful (a >= 300 status code).
+     * @throws IOException           if an I/O error occurs.
+     */
+    @Override
+    public Content handleResponse(final ClassicHttpResponse response) throws IOException {
+        final int statusCode = response.getCode();
+        final HttpEntity entity = response.getEntity();
+        final Content content = handleEntity(entity);
+        if (statusCode >= 300) {
+            final StringBuilder messageBuilder = new StringBuilder(response.getReasonPhrase());
```
ok2c, line 83: @arturobernalg I would not abuse the reason phase and would add an extra new attribute to the `HttpResponseException` class.

httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/ContentResponseHandler.java
```diff
@@ -44,11 +46,63 @@
...
+     * @return a {@link Content} object that encapsulates the response body, or {@link Content#NO_CONTENT} if the
+     * response body is null or has zero length.
+     * @throws IOException if an I/O error occurs.
+     */
     @Override
     public Content handleEntity(final HttpEntity entity) throws IOException {
-        return entity != null ?
-                new Content(EntityUtils.toByteArray(entity), ContentType.parse(entity.getContentType())) :
-                Content.NO_CONTENT;
+        if (entity != null && entity.getContentLength() > 0) {
```
ok2c, line 60: @arturobernalg Is this reasonable? It is legal for an entity to have its content length as -1 if it is chunk coded.

httpclient5-fluent/src/main/java/org/apache/hc/client5/http/fluent/ContentResponseHandler.java
```diff
@@ -44,11 +46,63 @@
...
+    @Override
+    public Content handleResponse(final ClassicHttpResponse response) throws IOException {
+        final int statusCode = response.getCode();
+        final HttpEntity entity = response.getEntity();
+        final Content content = handleEntity(entity);
+        if (statusCode >= 300) {
+            final StringBuilder messageBuilder = new StringBuilder(response.getReasonPhrase());
+            if (content != null && content.getType() != null) {
+                messageBuilder.append(", body was ");
+                truncateString(content.asString(), messageBuilder);
```
ok2c, line 86: @arturobernalg The truncation logic looks flawed. What is the point of truncating the message if it has already been fully loaded in memory? Use `EntityUtils#toByteArray` with the content max limit instead.

ok2c (changes requested):

ok2c (approved):
@arturobernalg Looks good otherwise.

PR #428, opened 2023-03-24
Implement HappyEyeballsV2AsyncClientConnectionOperator

arturobernalg:
This pull request adds a new implementation of the Happy Eyeballs algorithm for asynchronous client connections in Apache's HttpClient component. The new implementation, HappyEyeballsV2AsyncClientConnectionOperator, builds on the previous version by improving support for IPv6 and providing more fine-grained control over connection timing.

The changes include:
- Addition of HappyEyeballsV2AsyncClientConnectionOperator class and related test classes
- Modifications to existing code to use the new operator where appropriate
- Documentation updates
- Allow activate HappyEyeballsV2Async via Fast Fallback var in PoolingAsyncClientConnectionManagerBuilder.java

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
@@ -1,0 +1,745 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.http.impl.nio;
+
+import org.apache.hc.client5.http.DnsResolver;
+import org.apache.hc.client5.http.SchemePortResolver;
+import org.apache.hc.client5.http.SystemDefaultDnsResolver;
+import org.apache.hc.client5.http.impl.ConnPoolSupport;
+import org.apache.hc.client5.http.impl.DefaultSchemePortResolver;
+import org.apache.hc.client5.http.nio.AsyncClientConnectionOperator;
+import org.apache.hc.client5.http.nio.ManagedAsyncClientConnection;
+import org.apache.hc.core5.concurrent.FutureCallback;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.config.Lookup;
+import org.apache.hc.core5.http.nio.ssl.TlsStrategy;
+import org.apache.hc.core5.http.protocol.HttpContext;
+import org.apache.hc.core5.io.CloseMode;
+import org.apache.hc.core5.reactor.ConnectionInitiator;
+import org.apache.hc.core5.reactor.ssl.TransportSecurityLayer;
+import org.apache.hc.core5.util.Args;
+import org.apache.hc.core5.util.Timeout;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.ConnectException;
+import java.net.Inet4Address;
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.SocketAddress;
+import java.net.UnknownHostException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * The {@link AsyncClientConnectionOperator} implementation that uses Happy Eyeballs V2 algorithm to connect
+ * to the target server. Happy Eyeballs V2 (HEV2) algorithm is used to connect to the target server by concurrently
+ * attempting to establish multiple connections to different IP addresses. The first connection to complete
+ * successfully is selected and the others are closed. If all connections fail, the last error is rethrown.
+ * The algorithm also applies a configurable delay before subsequent connection attempts. HEV2 was introduced
+ * as a means to mitigate the latency issues caused by IPv4 and IPv6 co-existence in the Internet. HEV2 is defined
+ * in RFC 8305.
+ *
+ * <p>
+ * This connection operator maintains a connection pool for each unique route (combination of target host and
+ * target port) and selects the next connection from the pool to establish a new connection or reuse an
+ * existing connection. The connection pool uses a First-In-First-Out (FIFO) queue and has a configurable limit
+ * on the maximum number of connections that can be kept alive in the pool. Once the maximum number of connections
+ * has been reached, the oldest connection in the pool is closed to make room for a new one.
+ * </p>
+ *
+ * <p>
+ * This class is thread-safe and can be used in a multi-threaded environment.
+ * </p>
+ *
+ * <p>
+ * The HEV2 algorithm is configurable through the following parameters:
+ * <ul>
+ *   <li>{@code dualStackEnabled}: Whether to enable dual-stack connectivity. When set to {@code true},
+ *   the operator attempts to connect to both IPv4 and IPv6 addresses concurrently. When set to {@code false},
+ *   only IPv4 or IPv6 addresses are attempted depending on the address type of the target server.</li>
+ *   <li>{@code maxAttempts}: The maximum number of connection attempts to be made before failing. If all
+ *   attempts fail, the last error is rethrown.</li>
+ *   <li>{@code delay}: The delay (in milliseconds) to apply before subsequent connection attempts.</li>
+ *   <li>{@code connectTimeout}: The connection timeout (in milliseconds) for each attempt.</li>
+ * </ul>
+ * </p>
+ *
+ *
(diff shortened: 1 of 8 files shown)


ok2c:
@arturobernalg Looks good to me. @rschmitt Would you also like to review?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
```diff
@@ -0,0 +1,680 @@
...
+                            }
+                            attempt.cancel(true);
+                        }
+                    });
+
+            // Introduce a delay before executing the next connection attempt
+            if (i < addresses.size() - 1) {
+                try {
+                    final Duration delay = calculateDelay(i);
+                    delay.wait();
```
rschmitt, line 484: What thread is this code running in? This will block for up to 2,000 milliseconds.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperatorBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+
+        connectionOperator = Args.notNull(connectionOperator, "Connection operator");
+        final DnsResolver dnsResolverToUse = dnsResolver != null ? dnsResolver : SystemDefaultDnsResolver.INSTANCE;
+        final Timeout timeoutToUse = timeout != null ? timeout : Timeout.ofMilliseconds(250);
+        final Timeout resolutionDelayToUse = resolutionDelay != null ? resolutionDelay : Timeout.ofMilliseconds(50);
+        final Timeout minimumConnectionAttemptDelayToUse = minimumConnectionAttemptDelay != null ? minimumConnectionAttemptDelay : Timeout.ofMilliseconds(100);
+        final Timeout maximumConnectionAttemptDelayToUse = maximumConnectionAttemptDelay != null ? maximumConnectionAttemptDelay : Timeout.ofSeconds(2);
+        final Timeout connectionAttemptDelayToUse = connectionAttemptDelay != null ? connectionAttemptDelay : Timeout.ofMilliseconds(250);
+        final int firstAddressFamilyCountToUse = Args.positive(firstAddressFamilyCount, "First Address Family");
+        final HappyEyeballsV2AsyncClientConnectionOperator.AddressFamily addressFamilyToUse = addressFamily;
```
rschmitt, line 254: These defaults are all duplicated in the `HappyEyeballsV2AsyncClientConnectionOperator` constructor.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/InetAddressComparator.java
```diff
@@ -0,0 +1,89 @@
...
+            return 1;
+        }
+        // Compare IPv6 addresses first
+        if (adr1 instanceof Inet6Address && adr2 instanceof Inet4Address) {
+            return 1;
+        }
+        if (adr1 instanceof Inet4Address && adr2 instanceof Inet6Address) {
+            return -1;
+        }
+        // Compare based on address bytes
```
rschmitt, line 76: This appears to have nothing to do with the [address sorting algorithm](http://biplane.com.au/blog/?p=141) specified by [RFC 6724](https://www.rfc-editor.org/rfc/rfc6724#section-6), which is [mandated](https://www.rfc-editor.org/rfc/rfc8305#section-4) by RFC 8305. Furthermore, sorting addresses in this way is a bad idea, since it will defeat [round-robin DNS](https://en.wikipedia.org/wiki/Round-robin_DNS), an important load balancing technique.

httpclient5/src/test/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperatorTest.java
```diff
@@ -0,0 +1,309 @@
...
+                        1,
+                        HappyEyeballsV2AsyncClientConnectionOperator.AddressFamily.IPv4);
+
+        connectionInitiator = mock(ConnectionInitiator.class);
+        socketAddress = mock(SocketAddress.class);
+
+
+    }
+
+    @DisplayName("Test that application prioritizes IPv6 over IPv4 when both are available")
```
rschmitt, line 99: Again, the sorting rules are more complicated than this. For example, native IPv4 addresses are preferred to IPv6 addresses that use an encapsulation/tunneling transport like 6to4 or Teredo. IPv6 addresses will also be considered unusable, and therefore deprioritized, if no IPv6 _source_ address is available.

rschmitt (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
```diff
@@ -0,0 +1,680 @@
...
+     * @param host the host name to resolve DNS for
+     * @return a CompletableFuture that will be completed with an array of InetAddress objects representing the IP addresses
+     */
+    private CompletableFuture<InetAddress[]> resolveDnsAsync(final String host) {
+        final CompletableFuture<InetAddress[]> dnsFuture = new CompletableFuture<>();
+        CompletableFuture.runAsync(() -> {
+            try {
+                final InetAddress[] inetAddresses = dnsResolver.resolve(host);
+                // Introduce a delay before resolving AAAA records after receiving A records
+                resolution_delay.sleep();
```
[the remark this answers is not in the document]
arturobernalg: Instead of using Thread.sleep(), i could use non-blocking methods(ScheduledExecutorService) to delay the execution of the next connection attempt.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
arturobernalg: The RFC recommends introducing a delay before starting the next connection attempt, but it doesn't mandate a specific implementation. The RFC suggests a default delay of 250 milliseconds, but also notes that a more nuanced implementation can use historical RTT data to influence the delay. Therefore, it is up to the implementation to decide whether to introduce a delay and how to calculate that delay.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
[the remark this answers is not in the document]
rschmitt: Yes, in async code delays should always be implemented by yielding the current thread and scheduling a continuation, rather than calling `Thread.sleep`. Ideally, the DNS lookup would also be made as an async call, but to my knowledge there's really nothing we can do about that, and in practical terms DNS is very fast. I'm not even sure Netty has an async implementation of DNS lookups.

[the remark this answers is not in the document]
rhernandez35: Netty does, but it's an entirely bespoke implementation that doesn't touch the JDK's built-int resolver. The blocking here will hopefully become less of a problem for this client once Loom is available.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
```diff
@@ -0,0 +1,697 @@
...
+            try {
+                final InetAddress[] inetAddresses = dnsResolver.resolve(host);
+                dnsFuture.complete(inetAddresses);
+
+                // Introduce a delay before resolving AAAA records after receiving A records
+                if (inetAddresses.length > 0) {
+                    scheduler.schedule(() -> {
+                        try {
+                            final InetAddress[] inet6Addresses = dnsResolver.resolve(host);
+                            dnsFuture.complete(inet6Addresses);
```
rhernandez35, line 415: You can't complete `dnsFuture` twice. The `complete` call on line 408 will always put `dnsFuture` into a finished state.

rhernandez35: And please don't use `obtrudeValue` to work around `complete` being one-and-done

arturobernalg:
Hi @rschmitt 
just wanted to let you know that I have implemented Destination Address Selection and improved the handling of async connections in HappyEyeballsV2AsyncClientConnectionOperator. 
I have also added a new class to validate the HappyEyeballsV2 rules: HappyEyeballsV2RulesTest. Thought you might find it useful. 

Let me know if you have any feedback. Thanks!

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/InetAddressComparator.java
```diff
@@ -0,0 +1,470 @@
...
+     */
+    private int getScope(final InetAddress address) {
+        if (address instanceof Inet6Address) {
+            final Inet6Address ipv6Addr = (Inet6Address) address;
+            final int scope = ipv6Addr.getScopeId();
+            if (scope > 0) {
+                return scope;
+            }
+        }
+        return -1;
```
rschmitt, line 291: Isn't `127.0.0.1/8` supposed to have special treatment?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
```diff
@@ -0,0 +1,698 @@
...
+                    for (final InetAddress inetAddress : inetAddresses) {
+                        if (inetAddress instanceof Inet4Address) {
+                            ipv4Addresses.add(inetAddress);
+                        } else if (inetAddress instanceof Inet6Address) {
+                            ipv6Addresses.add(inetAddress);
+                        }
+                    }
+
+                    // Sort the array of addresses using the custom Comparator
+                    Arrays.sort(inetAddresses, InetAddressComparator.INSTANCE);
```
rschmitt, line 369: Is this a [stable sort](https://www.geeksforgeeks.org/stable-and-unstable-sorting-algorithms/)? It's important to not change the order of IP addresses that are equal under the comparison rules.

rschmitt (changes requested):
I think a lot of the destination address sorting rules need to be reimplemented in terms of local address availability. It's also important to make sure we are using a stable sort. Finally, the `compareTo` method can't make any network calls.

These rules are complex enough that they really need dedicated unit testing. Generative testing may be a good idea, particularly to ensure that the sort is stable.

arturobernalg:
Hi @rschmitt I've made a commit addressing the feedback you provided on the pull request. I believe I've covered all the points. There isn't a specific JUnit test for the rules; there's only an example test class. Currently, I'm looking into how to approach the testing since the class is internal and not exposed at the moment. 
Please feel free to provide any further comments or suggestions.

ok2c:
@rschmitt @rhernandez35 Folks,, is there any chance you could do another review sometime soon? @arturobernalg Please in the meantime change the target branch to `5.4.x`.

PR #430, opened 2023-04-01
Add support for proxy failover

arturobernalg:
This pull request adds support for proxy failover in the Apache HttpClient. The current implementation of the ProxySelectorRoutePlanner selects a single proxy from the list returned by the java.net.ProxySelector API. If the client can't open a socket to the proxy, other proxies in the list are ignored and the request fails. This is not optimal in network configurations where proxies can go up and down for maintenance.

To address this, I have implemented a new FailoverProxySelector class that selects a list of proxies for the given URI by delegating to the next available ProxySelector in the list, until a non-empty proxy list is returned. Once a non-empty proxy list is returned, it becomes the active ProxySelector and is used for subsequent calls to select(). The active ProxySelector is remembered and used for future requests until it fails, at which point the next ProxySelector is used.

This implementation is more efficient than the previous solution, which created a disposable socket for every request. It also provides better support for network configurations where proxies can go up and down for maintenance.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/DistributedProxySelector.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/DistributedProxySelector.java
@@ -1,0 +1,179 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl.routing;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.Proxy;
+import java.net.ProxySelector;
+import java.net.SocketAddress;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A DistributedProxySelector is a custom {@link ProxySelector} implementation that
+ * delegates proxy selection to a list of underlying ProxySelectors in a
+ * distributed manner. It ensures that proxy selection is load-balanced
+ * across the available ProxySelectors, and provides thread safety by
+ * maintaining separate states for each thread.
+ *
+ * <p>The DistributedProxySelector class maintains a list of ProxySelectors,
+ * a {@link ThreadLocal} variable for the current {@link ProxySelector}, and an {@link AtomicInteger}
+ * to keep track of the shared index across all threads. When the select()
+ * method is called, it delegates the proxy selection to the current
+ * ProxySelector or the next available one in the list if the current one
+ * returns an empty proxy list. Any exceptions that occur during proxy
+ * selection are caught and ignored, and the next ProxySelector is tried.
+ *
+ * <p>The connectFailed() method notifies the active {@link ProxySelector} of a
+ * connection failure, allowing the underlying ProxySelector to handle
+ * connection failures according to its own logic.
+ *
+ * @since 5.3
+ */
+@Contract(threading = ThreadingBehavior.SAFE)
+public class DistributedProxySelector extends ProxySelector {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DistributedProxySelector.class);
+
+    /**
+     * A list of {@link ProxySelector} instances to be used by the DistributedProxySelector
+     * for selecting proxies.
+     */
+    private final List<ProxySelector> selectors;
+
+    /**
+     * A {@link ThreadLocal} variable holding the current {@link ProxySelector} for each thread,
+     * ensuring thread safety when accessing the current {@link ProxySelector}.
+     */
+    private final ThreadLocal<ProxySelector> currentSelector;
+
+    /**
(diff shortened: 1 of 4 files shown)


ok2c (changes requested):
@arturobernalg Could you please also update `HttpClientBuilder` and make sure both `HttpClientBuilder` and `HttpAsyncClientBuilder` behave consistently?

arturobernalg:
> @arturobernalg Could you please also update `HttpClientBuilder` and make sure both `HttpClientBuilder` and `HttpAsyncClientBuilder` behave consistently?

@ok2c I update the HttpClientBuilder and unified the logic. 
TY

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/FailoverProxySelector.java
```diff
@@ -0,0 +1,123 @@
...
+ * one fails.
+ *
+ * @since 5.3
+ */
+public class FailoverProxySelector extends ProxySelector {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FailoverProxySelector.class);
+
+    private final List<ProxySelector> selectors;
+    private int currentIndex = 0;
```
ok2c, line 63: @arturobernalg I think `currentIndex` should be volatile or represented as an `AtomicInteger`.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/FailoverProxySelector.java
arturobernalg: Yes, you are correct. If multiple threads access the currentIndex variable in the FailoverProxySelector class, then it should be made thread-safe.
Changed.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/FailoverProxySelector.java
```diff
@@ -0,0 +1,127 @@
...
+     * Selects a list of proxies for the given URI by delegating to the next
+     * available ProxySelector in the list, until a non-empty proxy list is
+     * returned. Once a non-empty proxy list is returned, it becomes the active
+     * ProxySelector and is used for subsequent calls to select().
+     *
+     * @param uri the URI to select a proxy for.
+     * @return a list of proxies for the given URI.
+     */
+    @Override
+    public List<Proxy> select(final URI uri) {
```
ok2c, line 92: @arturobernalg What is exactly the intent of this implementation? Pick a backup selector in case the primary selector fails to produce a result or distribute the load randomly across multiple selectors? I am not quite sure I understand the logic of this method

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/FailoverProxySelector.java
```diff
@@ -0,0 +1,127 @@
...
+     * Notifies the active ProxySelector that a connection to the given URI
+     * using the given proxy has failed.
+     *
+     * @param uri the URI that failed to connect.
+     * @param sa  the SocketAddress of the proxy that failed to connect.
+     * @param ioe the IOException that resulted from the failed connection.
+     */
+    @Override
+    public void connectFailed(final URI uri, final SocketAddress sa, final IOException ioe) {
+        final ProxySelector selector = selectors.get(currentIndex.get());
```
ok2c, line 124: @arturobernalg This approach can produce unexpected results in the selector is being used concurrently by multiple threads. The current index can easily change from another thread while the `#select` method is still being called. As much as I hate `ThreadLocal`s I suppose this implementation must maintain a separate state per each thread of execution.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/FailoverProxySelector.java
arturobernalg: Is to select a list of proxies for the given URI by delegating to the next available ProxySelector in the list, until a non-empty proxy list is returned. It does not necessarily pick a backup selector in case the primary selector fails to produce a result, but rather rotates through the list of selectors until a non-empty proxy list is obtained

ok2c: @arturobernalg In this case `FailoverProxySelector` may be not the best choice of the name. I personally think `spraying` or  distributing` would likely be a better term.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/FailoverProxySelector.java
ok2c: @arturobernalg Better, but we can make it better still. 
1. Use `ThreadLocal` to store the currently chosen selector
2. Unset the `ThreadLocal` once the `#select` call has been completed successfully or unsuccessfully. 
3. Use `AtomicInteger` to store the shared current index

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/DistributedProxySelector.java
```diff
@@ -0,0 +1,137 @@
...
+ * redundant proxy selection calls.
+ *
+ * @since 5.3
+ */
+public class DistributedProxySelector extends ProxySelector {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DistributedProxySelector.class);
+
+    private final List<ProxySelector> selectors;
+    private final ThreadLocal<Integer> currentIndex;
```
ok2c, line 69: @arturobernalg 
1. You can stick `ProxySelector` into `ThreadLocal`. It will make the logic more readable.
2. Do not forget to remove it in `#connectFailed`

arturobernalg: @ok2c 
I have now incorporated this change into the DistributedProxySelector class, which has made the logic more readable and ensures thread safety.

The updated implementation now utilizes ThreadLocal<ProxySelector> for the currentSelector, allowing each thread to have its own instance of the ProxySelector. Additionally, I have updated the connectFailed method to remove the currentSelector from the ThreadLocal variable after handling the connection failure.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/DistributedProxySelector.java
```diff
@@ -0,0 +1,183 @@
...
+        for (int i = 0; i < selectors.size(); i++) {
+            if (selector == null) {
+                selector = nextSelector();
+                currentSelector.set(selector);
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Current selector is null, selecting next proxy selector for URI {}: {}", uri, selector);
+                }
+            }
+            try {
+                result = selector.select(uri);
```
ok2c, line 130: @arturobernalg Almost there. Please store the current selector only for the duration of the `result = selector.select(uri)` call and unset it immediately after (or in case of a failure in `#connectFailed`). The main idea is to use thread local only as short as possible and unset it as soon as possible to avoid thread local leaks.

arturobernalg: done

ok2c (approved):
@arturobernalg Very good.

PR #431, opened 2023-04-01
Enable IDN conversion for scheme port resolver

arturobernalg:
**Summary**

This pull request adds support for IDN conversion in the schemePortResolver for HTTP client requests. When enableIDNConversion is set to true and schemePortResolver is null, a new DefaultSchemePortResolver object is created with a custom implementation that uses the IDN.toASCII method to convert hostnames to ASCII format. This is compliant with RFC 5891 and helps to avoid issues with non-ASCII hostnames in HTTP requests.

**Changes Made**

Added null check for schemePortResolver and creation of new DefaultSchemePortResolver object if it is null and enableIDNConversion is true.
Implemented custom resolve method in DefaultSchemePortResolver object to perform IDN conversion using IDN.toASCII method.
Testing

Created unit tests to verify that IDN conversion works as expected.
Ran existing test suite to ensure that no existing functionality is broken.
Notes

This pull request does not modify any existing behavior when enableIDNConversion is false or schemePortResolver is already set.
The changes made in this pull request are compliant with RFC 5891 and help to avoid issues with non-ASCII hostnames in HTTP requests.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
@@ -28,6 +28,7 @@
 package org.apache.hc.client5.http.impl.classic;
 
 import java.io.Closeable;
+import java.net.IDN;
 import java.net.ProxySelector;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -82,6 +83,8 @@ import org.apache.hc.client5.http.protocol.RequestClientConnControl;
 import org.apache.hc.client5.http.protocol.RequestDefaultHeaders;
 import org.apache.hc.client5.http.protocol.RequestExpectContinue;
 import org.apache.hc.client5.http.protocol.ResponseProcessCookies;
+import org.apache.hc.client5.http.psl.PublicSuffixMatcher;
+import org.apache.hc.client5.http.psl.PublicSuffixMatcherLoader;
 import org.apache.hc.client5.http.routing.HttpRoutePlanner;
 import org.apache.hc.core5.annotation.Internal;
 import org.apache.hc.core5.http.ConnectionReuseStrategy;
@@ -217,7 +220,6 @@ public class HttpClientBuilder {
     private boolean evictExpiredConnections;
     private boolean evictIdleConnections;
     private TimeValue maxIdleTime;
-
     private boolean systemProperties;
     private boolean redirectHandlingDisabled;
     private boolean automaticRetriesDisabled;
@@ -226,6 +228,7 @@ public class HttpClientBuilder {
     private boolean authCachingDisabled;
     private boolean connectionStateDisabled;
     private boolean defaultUserAgentDisabled;
+    private boolean enableIDNConversion;
 
     private List<Closeable> closeables;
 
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
```diff
@@ -751,6 +766,26 @@ public CloseableHttpClient build() {
...
         }
 
+        if (enableIDNConversion && schemePortResolver == null) {
+            // Load the public suffix list
+            final PublicSuffixMatcher publicSuffixMatcher = PublicSuffixMatcherLoader.getDefault();
+
+            schemePortResolver = new DefaultSchemePortResolver() {
+                @Override
+                public int resolve(final HttpHost host) {
+                    final String hostname = IDN.toASCII(host.getHostName());
```
ok2c, line 776: @arturobernalg I am not sure I understand why host name conversion is necessary here. The purpose of the `SchemePortResolver` is to provide the default port for a protocol scheme: `http` -> 80, `https` -> 443, `myhttp` -> 8888. Could you please give an example where the proposed host name translation is necessary?

arturobernalg: @ok2c 
The purpose of converting the hostname to its ASCII representation using the IDN.toASCII method is to handle domain names that contain non-ASCII characters which may not be supported by some components in the networking stack. By converting the hostname to its ASCII representation, now we ensure that it can be correctly resolved and used by these components.

ok2c: @arturobernalg I understand what the IDN encoding is for. I do not understand what it has to do with the process of the scheme port resolution.

arturobernalg: @ok2c 
Here's an example of how this conversion could be necessary: let's say you have a URI with an IDN hostname like http://www.öbb.at/, and you are using the SchemePortResolver to determine the default port for the http scheme. If the getHostName method returns the hostname as www.öbb.at, some components in the networking stack may not be able to handle it correctly, and you may get errors or unexpected behavior. By converting the hostname to its ASCII representation (www.xn--bb-eka.at) using IDN.toASCII, you ensure that it can be correctly resolved and used by these components.

ok2c: @arturobernalg Look. I understand what the IDN encoding is for and why it is necessary but please understand what I am trying to say here. The hostname encoding makes no difference on the result of the  `SchemePortResolver`. It produces the same result for both versions because it does not take the host name into consideration. At all. 
```
http://www.öbb.at -> 80
http://www.xn--bb-eka.at -> 80
```
The IDN conversion should take place in `HttpHost` and nowhere else. The `SchemePortResolver` is simply the wrong place.

ok2c (changes requested):

arturobernalg:
No need it anymore. 

https://github.com/apache/httpcomponents-core/pull/394

PR #432, opened 2023-04-07
Refactored some of the implementation and design smells

smit2509:


ok2c:
@smit2509 What is the point of all these changes? Why do you think they are useful and how? Not to mention that these changes break API compatibility for no reason of what so ever. I think I could accept 032549fa820bcbf1676271d245bec9ed1b7bd450 but no other commits.

PR #437, opened 2023-04-24
Improve AIMDBackoffManager with atomic references, thread-safety, and parameter checks

arturobernalg:
In this commit, we have made several improvements to the AIMDBackoffManager class:

* Replaced volatile variables with AtomicReference and AtomicInteger to ensure better thread-safety and atomic operations.
* Updated the @Contract annotation to reflect the new thread-safe behavior of the class.
 * Added parameter validation checks to public methods for better error handling and prevention of incorrect usage.
* Improved the code readability and organization by separating sections of the class with blank lines and consistent indentation.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/AIMDBackoffManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/AIMDBackoffManager.java
@@ -26,15 +26,21 @@
  */
 package org.apache.hc.client5.http.impl.classic;
 
-import java.util.HashMap;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
-
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.hc.client5.http.HttpRoute;
 import org.apache.hc.client5.http.classic.BackoffManager;
-import org.apache.hc.core5.annotation.Experimental;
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/AIMDBackoffManager.java
```diff
@@ -56,79 +62,134 @@
...
-                || now - lastBackoff < coolDown.toMilliseconds()) {
-                return;
+        final int curr = connPerRoute.getMaxPerRoute(route);
+        final int max = (curr >= cap.get()) ? cap.get() : curr + 1;
+        final Instant now = Instant.now();
+
+        lastRouteProbes.compute(route, (r, lastProbe) -> {
+            if (lastProbe == null || now.isAfter(lastProbe.plus(Duration.ofMillis(coolDown.get().toMilliseconds())))) {
+                final Instant lastBackoff = lastRouteBackoffs.get(r);
+                if (lastBackoff == null || now.isAfter(lastBackoff.plus(Duration.ofMillis(coolDown.get().toMilliseconds())))) {
```
ok2c, line 172: @arturobernalg It is a nitpick but one can avoid conversion to the `Duration` here. I think `lastProbe.plus(coolDown.get().toMilliseconds(), ChronoUnit.MILLIS)` should work the same way here with one less intermediate object on the heap.

arturobernalg: Sounds reasonable.
changed.

PR #444, opened 2023-05-06
HTTPCLIENT-2272  - Add AbstractBackoff, LinearBackoffManager, and ExponentialBackoffManager classes

arturobernalg:
- Create AbstractBackoff class as a base for different backoff strategies
- Add LinearBackoffManager class for implementing linear backoff algorithm in networking and communication systems. This class provides a thread-safe and configurable implementation of linear backoff, allowing for gradual adjustment of maximum connection pool sizes for a given route based on traffic and other factors. The class supports customizable cool-down periods and increment values for each route, and uses a ConcurrentHashMap to ensure thread safety and accurate tracking of backoff and probe attempts
- Implement ExponentialBackoffManager for connection pool control.
This commit introduces the ExponentialBackoffManager class, which manages the connection pool control for HTTP routes based on the Exponential Backoff algorithm. This implementation helps improve connection handling and stability in case of connection failures or network issues.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/AIMDBackoffManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/AIMDBackoffManager.java
@@ -26,21 +26,11 @@
  */
 package org.apache.hc.client5.http.impl.classic;
 
-import java.time.Instant;
-import java.time.temporal.ChronoUnit;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 import org.apache.hc.client5.http.HttpRoute;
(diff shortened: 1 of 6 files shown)


ok2c:
@arturobernalg So much better. Almost there. Let's reduce the public API footprint a bit, though. Generally there should be no protected instance variables. It will be very difficult to change the internal structure of a class in the future if it has protected instance variables. Please make those variables private and provide protected getters for them. 

And there are still test case failures.

arturobernalg:
> @arturobernalg So much better. Almost there. Let's reduce the public API footprint a bit, though. Generally there should be no protected instance variables. It will be very difficult to change the internal structure of a class in the future if it has protected instance variables. Please make those variables private and provide protected getters for them.
> 
> And there are still test case failures.
@ok2c 
 I have made the previously protected instance variables private and provided protected getters for them. This will help maintain encapsulation and allow for easier changes to the internal structure of the class in the future.
Regarding the test case failures, I have reviewed and made adjustments to the tests to minimize timing-related issues. Should be fine now.

PR #446, opened 2023-05-11
HTTPCLIENT-2273: Enhance "no-cache" directive handling

arturobernalg:
This pull request addresses ticket [HTTPCLIENT-2273](https://issues.apache.org/jira/browse/HTTPCLIENT-2273), which identifies an issue with the Apache HttpComponents client's handling of the "no-cache" directive when used with specific header fields, as outlined in RFC 7234.

Changes made to address this issue include:

- Updating the caching module to correctly identify and handle "no-cache" directives with specific header fields. This was done by modifying the `responseContainsNoCacheDirective` method.

-  Modifying the `handleCacheHit` method to ensure that a cached entry is revalidated with the origin server when a "no-cache" directive with specific header fields is present in the response.

- Ensuring that the rest of the response is still cacheable when the specified header fields are present, as long as the response complies with other caching requirements.

These changes allow an origin server to prevent the reuse of certain header fields in a response, while still allowing caching of the rest of the response. Corresponding unit tests have been updated to reflect these changes.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControl.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControl.java
@@ -31,6 +31,9 @@ import org.apache.hc.core5.annotation.Contract;
 import org.apache.hc.core5.annotation.Internal;
 import org.apache.hc.core5.annotation.ThreadingBehavior;
 
+import java.util.Collections;
+import java.util.Set;
+
 /**
  * Represents the values of the Cache-Control header in an HTTP response, which indicate whether and for how long
  * the response can be cached by the client and intermediary proxies.
@@ -92,16 +95,22 @@ final class CacheControl {
      * The number of seconds that a stale response is considered fresh for the purpose
      * of serving a response while a revalidation request is made to the origin server.
      */
-    private final long stale_while_revalidate;
+    private final long staleWhileRevalidate;
+
+    /**
+     * A set of field names specified in the "no-cache" directive of the Cache-Control header.
+     */
+    private final Set<String> noCacheFields;
 
 
     /**
      * Creates a new instance of {@code CacheControl} with default values.
      * The default values are: max-age=-1, shared-max-age=-1, must-revalidate=false, no-cache=false,
-     * no-store=false, private=false, proxy-revalidate=false, public=false and stale_while_revalidate=-1.
+     * no-store=false, private=false, proxy-revalidate=false, public=false, stale_while_revalidate=-1,
+     * and noCacheFields as an empty set.
      */
     public CacheControl() {
-        this(-1, -1, false, false, false, false, false, false,-1);
+        this(-1, -1, false, false, false, false, false, false,-1, Collections.emptySet());
     }
 
     /**
@@ -116,10 +125,12 @@ final class CacheControl {
      * @param cachePrivate    The private value from the Cache-Control header.
      * @param proxyRevalidate The proxy-revalidate value from the Cache-Control header.
      * @param cachePublic     The public value from the Cache-Control header.
+     * @param noCacheFields        The set of field names specified in the "no-cache" directive of the Cache-Control header.
      */
     public CacheControl(final long maxAge, final long sharedMaxAge, final boolean mustRevalidate, final boolean noCache, final boolean noStore,
                         final boolean cachePrivate, final boolean proxyRevalidate, final boolean cachePublic,
-                        final long stale_while_revalidate) {
+                        final long staleWhileRevalidate,
+                        final Set<String> noCacheFields) {
         this.maxAge = maxAge;
         this.sharedMaxAge = sharedMaxAge;
         this.noCache = noCache;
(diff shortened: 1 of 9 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControl.java
```diff
@@ -129,6 +140,7 @@ public CacheControl(final long maxAge, final long sharedMaxAge, final boolean mu
         this.proxyRevalidate = proxyRevalidate;
         this.cachePublic = cachePublic;
         this.stale_while_revalidate = stale_while_revalidate;
+        this.noCacheFields = noCacheFields;
```
ok2c, line 143: @arturobernalg Please make this set unmodifiable

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -162,6 +169,19 @@ public final CacheControl parse(final Header header) {
                 mustRevalidate = true;
             } else if (name.equalsIgnoreCase(HeaderConstants.CACHE_CONTROL_NO_CACHE)) {
                 noCache = true;
+                if (value != null) {
+                    final Tokenizer.Cursor valueCursor = new Tokenizer.Cursor(cursor.getPos() - value.length() -1, buffer.length());
```
ok2c, line 173: @ There is a way to parse the value in a single pass without creating an extra `Tokenizer.Cursor`. It is nice to have, though.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -525,4 +544,51 @@ private CacheControl parseCacheControlHeader(final MessageHeaders messageHeaders
...
+            // If no-cache directive is present and has no field names
+            if (responseCacheControl.isNoCache() && noCacheFields.isEmpty()) {
+                LOG.debug("No-cache directive present without field names. Revalidation required.");
+                return true;
+            }
+
+            // If no-cache directive is present with field names
+            if (responseCacheControl.isNoCache()) {
+                final HeaderGroup headerGroup = new HeaderGroup();
+                headerGroup.setHeaders(entry.getHeaders());
```
ok2c, line 576: @arturobernalg This looks completely unnecessary. `HttpCacheEntry`has `#getFirstHeader` method already.

ok2c (changes requested):

ok2c:
@arturobernalg 

1. While reviewing your code I realized that `CacheControlHeaderParser` in its present form was wrong. It assumes that there can be only a single header `Cache-Control` header, which is wrong. There can be multiple headers, all of which must be parsed and taken into account. I also would like the parser to be able to parse `Cache-Control` with values in one pass with a single cursor.

Headers
```
Cache-Control: this, that
```
and
```
Cache-Control: this
Cache-Control: that
```
must be treated as equivalent and must produce the same `Cache-Control` structure.

2. I found out that `Cache-Control` get parsed over and over again multiple times in the course of a request execution. This is wasteful and conceptually wrong. 

I am going to address these two issues, once your PR is merged.

arturobernalg:
> @arturobernalg
> 
> 1. While reviewing your code I realized that `CacheControlHeaderParser` in its present form was wrong. It assumes that there can be only a single header `Cache-Control` header, which is wrong. There can be multiple headers, all of which must be parsed and taken into account.
> 
> Headers
> 
> ```
> Cache-Control: this, that
> ```
> 
> and
> 
> ```
> Cache-Control: this
> Cache-Control: that
> ```
> 
> must be treated as equivalent and must produce the same `Cache-Control` structure. I also would like the parser to be able to parse `Cache-Control` with values in one pass with a single cursor.
> 
> 2. I found out that `Cache-Control` get parsed over and over again multiple times in the course of a request execution. This is wasteful and conceptually wrong.
> 
> I am going to address these two issues, once your PR is merged.

@ok2c , you're right. I didn't realize that. My apologies. I have made the necessary changes to resolve the issue. After merging the HTTPCLIENT-2273 PR, I will update the branch and fix the conflict. You can find the PR here: https://github.com/apache/httpcomponents-client/pull/447.

Regarding point 2, I would also like to address it, but I need some guidance or instructions to proceed."

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
arturobernalg: I made some changes to avoid creating an additional Tokenizer.Cursor, but I'm not completely certain if it's what you're looking for..

ok2c: @arturobernalg My apologies. My suggestion was completely unfounded. Your initial implementation was perfectly fine. I restored it in your commit. Sorry about that.

arturobernalg: > @arturobernalg My apologies. My suggestion was completely unfounded. Your initial implementation was perfectly fine. I restored it in your commit. Sorry about that.

np @ok2c

PR #447, opened 2023-05-13
Refactor CacheControl parser to handle multiple headers.

arturobernalg:
This commit refactors the CacheControl parsing logic to handle multiple "Cache-Control" headers. The previous implementation treated each header independently, returning an array of CacheControl objects. This caused issues when headers had directives that should be combined into a single CacheControl object. The updated implementation combines all directives from all headers into a single CacheControl object, ensuring accurate representation of the caching directives.


arturobernalg:
What I'm not yet certain about is how to avoid parsing multiple times during a request's lifecycle. It seems to be a complex change because it involves modifying several classes/methods.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -113,28 +113,14 @@ protected CacheControlHeaderParser() {
...
-        } else {
-            final String s = header.getValue();
-            if (s == null) {
-                return new CacheControl();
-            }
-            buffer = new CharArrayBuffer(s.length());
-            buffer.append(s);
-            cursor = new Tokenizer.Cursor(0, buffer.length());
-        }
+    public final CacheControl parse(final Header[] headers) {
```
ok2c, line 120: @arturobernalg Please use `Iterator<Header>` instead. One can then use `Iterator<Header> MessageHeaders#headerIterator(String name)` without creating an intermediate array of `Header`s.

ok2c (changes requested):

PR #449, opened 2023-05-15
Cache-Control improvements

ok2c:
@arturobernalg I have started working on Cache-Control improvements throughout the entire HTTP cache code. I added `RequestCacheControl` to represent request cache control directives similar to `CacheControl`. Please take a look and let me know if you find anything disagreeable.  As the next step I will try to make sure the request and response `Cache-Control` headers in the request and the response get parted only once instead of multiple times in various places.

We may want to rename `CacheControl` to `ResponseCacheControl` for consistency. Please let me know what you think.


arturobernalg:
> @arturobernalg I have started working on Cache-Control improvements throughout the entire HTTP cache code. I added `RequestCacheControl` to represent request cache control directives similar to `CacheControl`. Please take a look and let me know if you find anything disagreeable. As the next step I will try to make sure the request and response `Cache-Control` headers in the request and the response get parted only once instead of multiple times in various places.
> 
> We may want to rename `CacheControl` to `ResponseCacheControl` for consistency. Please let me know what you think.

@ok2c 

Regarding renaming 'CacheControl' to 'ResponseCacheControl', I'm in agreement with this proposal. It brings clarity and consistency to our codebase, thereby enhancing its comprehensibility.
Have we considered the possibility of creating an abstract class  for common directives?
Overall, it looks good.
By the way, the use of 'BiConsumer' is a nice touch.

ok2c:
@arturobernalg Could you please review this PR and let me know if you find anything disagreeable?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachedResponseSuitabilityChecker.java
```diff
@@ -69,84 +70,51 @@ class CachedResponseSuitabilityChecker {
...
-                        }
-                    } catch (final NumberFormatException nfe) {
-                        // err on the side of preserving semantic transparency
-                        maxStale = 0;
-                    }
-                }
-            }
-        }
-        return maxStale;
+        final ResponseCacheControl cacheControl = CacheControlHeaderParser.INSTANCE.parse(entry);
```
arturobernalg, line 99: I might be mistaken or perhaps missing something, but don't we already have the ResponseCacheControl? It seems like we just need to pass it through the 'isFreshEnough' method.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExec.java
```diff
@@ -290,38 +293,38 @@ private ClassicHttpResponse handleCacheHit(
 
         if (requestContainsNoCacheDirective(request)) {
```
arturobernalg, line 294: We can The method 'requestContainsNoCacheDirective' also parses the request, but we have already done that. We should simply pass the RequestCacheControl. Also, it should be RequestCacheControl, not ResponseCacheControl.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ConditionalRequestBuilder.java
```diff
@@ -58,26 +56,17 @@
...
-            final HeaderElement elt = it.next();
-            if (HeaderConstants.CACHE_CONTROL_MUST_REVALIDATE.equalsIgnoreCase(elt.getName())
-                    || HeaderConstants.CACHE_CONTROL_PROXY_REVALIDATE.equalsIgnoreCase(elt.getName())) {
-                mustRevalidate = true;
-                break;
-            }
-        }
-        if (mustRevalidate) {
-            newRequest.addHeader(HeaderConstants.CACHE_CONTROL, HeaderConstants.CACHE_CONTROL_MAX_AGE + "=0");
+        final ResponseCacheControl cacheControl = CacheControlHeaderParser.INSTANCE.parse(cacheEntry);
```
arturobernalg, line 67: Similarly, since we already possess the ResponseCacheControl, we should simply pass it as a parameter

arturobernalg (commented):
@ok2c In general, the changes appear solid; I've just added two minor comments.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachedResponseSuitabilityChecker.java
ok2c: @arturobernalg No, you are not mistaken. I am just getting old. Corrected.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExec.java
ok2c: @ok2c Corrected.

ok2c:
@arturobernalg Please do another pass.

arturobernalg:
> @arturobernalg Please do another pass.

@ok2c 
This change-set looks good to me.

PR #454, opened 2023-06-11
HTTPCLIENT-2277: do not store hop-by-hop and connection specific headers in cache

ok2c:
As per [RFC 9111 3.1](https://datatracker.ietf.org/doc/html/rfc9111#name-storing-header-and-trailer-) the cache now filters out hop-by-hop and connection specific headers prior to committing them to the storage. It also no longer stores the original `Content-Length` which is now gets generated based on the underlying cache entry resource.

@arturobernalg Please review.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/CacheHeaderSupport.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/CacheHeaderSupport.java
@@ -1,0 +1,93 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


arturobernalg (approved):
@ok2c 
Jut two remarks,

-  I noticed that the Upgrade and Trailer headers are currently missing from the HOP_BY_HOP Set. These headers are typically considered "hop-by-hop" headers according to the HTTP/1.1 specification, and may be necessary depending on the intended use of this class. 
- Given that the primary operation we are using in this case is contains, which checks if a certain header is in the "hop-by-hop" set, we might get better performance by using a HashSet. A HashSet has an average time complexity of O(1) for these operations which makes it more efficient for larger data sets.


In any case, the change-set looks good.

ok2c:
> @ok2c Just two remarks,
> 
>     * I noticed that the Upgrade and Trailer headers are currently missing from the HOP_BY_HOP Set. 

@arturobernalg Good catch. I overlooked `Upgrade`. Corrected. But for some reason `Trailer` is no longer mentioned in RFC 9110, Section 7.6.1. I do not know why. So is `Host` but I think it should be. This is the problem with mixing up conformance with different specification revisions. 

> 
>     * Given that the primary operation we are using in this case is contains, which checks if a certain header is in the "hop-by-hop" set, we might get better performance by using a HashSet. A HashSet has an average time complexity of O(1) for these operations which makes it more efficient for larger data sets.

This is true. The problem is that `HashSet` operations are case-sensitive. Therefore each lookup operation would require a normalized input, which in practical terms implies creation of an extra copy of the header name. I believe this is more expensive than having a `TreeSet` with just a few elements in it.

PR #457, opened 2023-06-17
Check if is Suitable Variant

arturobernalg:


ok2c:
@arturobernalg This is all unnecessary in my opinion. With the current design there will always be just one cache entry matching a given key. Please wait until I get my changes ready for review.

PR #458, opened 2023-06-18
HTTPCLIENT-2277, HTTPCLIENT-1347: Revision of the variant handling by the HTTP cache implementations

ok2c:
Revision of the variant handling by the HTTP cache implementations

* Cache entries now can be of two distinct types: root entries containing a map of known representation variants of the same resource and resource entries containing a resource potentially sharable by multiple resource entries.The same entry cannot have both: a variant map and a resource (fixes HTTPCLIENT-1347)
* Cache entry factory class added to the public APIs
* Cleanup of internal cache implementations


ok2c:
@arturobernalg Please review.

arturobernalg:
@ok2c the change-set looks good

PR #460, opened 2023-06-21
HTTPCLIENT-2277: Revision and optimization of cache key generation

ok2c:
Presently a lot of cache operations are sub-optimal and redundant. The same cache key may be generated and the same cache lookup performed over and over again in different places in the course of a single request execution

I am optimizing the cache code starting with the most fundamental bit: the cache key generator

Key generator changes:
* improved normalization of variant keys
* potentially less intermediate garbage and a better performance of key computation

@arturobernalg Please review.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/BasicHttpAsyncCache.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/BasicHttpAsyncCache.java
@@ -28,6 +28,7 @@ package org.apache.hc.client5.http.impl.cache;
 
 import java.time.Instant;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
```diff
@@ -94,18 +96,67 @@ public String generateKey(final HttpHost host, final HttpRequest request) {
...
+    /**
+     * Returns all variant names contained in {@literal VARY} headers of the given message.
+     *
+     * @since 5.3
+     */
+    public static List<String> variantNames(final MessageHeaders message) {
+        if (message == null) {
+            return null;
+        }
+        if (!message.containsHeader(HttpHeaders.VARY)) {
```
arturobernalg, line 108: Could we possibly omit the check `if (!message.containsHeader(HttpHeaders.VARY))`? 
If I understand correctly, if there are no `VARY` headers present, the loop `for (final Iterator<Header> it = message.headerIterator(HttpHeaders.VARY); it.hasNext(); )` wouldn't iterate, making the explicit check redundant. Am I missing something?"

arturobernalg (approved):
@ok2c, the changes-set look good overall. I just have two minor remarks/questions that I'd like to raise.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/BasicHttpAsyncCache.java
```diff
@@ -90,11 +91,15 @@ public BasicHttpAsyncCache(final ResourceFactory resourceFactory, final HttpAsyn
 
     @Override
     public String generateKey(final HttpHost host, final HttpRequest request, final HttpCacheEntry cacheEntry) {
-        if (cacheEntry == null) {
-            return cacheKeyGenerator.generateKey(host, request);
-        } else {
-            return cacheKeyGenerator.generateKey(host, request, cacheEntry);
+        final String root = cacheKeyGenerator.generateKey(host, request);
```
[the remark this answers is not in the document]
ok2c: @arturobernalg I am aware. I am working on a major refactoring of the caching logic and those duplicate bits will go away. Ignore it for now and bear with me.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
ok2c: @arturobernalg It is a very valid point. It is a trade-off to save one intermediate object allocation if no `Vary` headers are present, but may be it is not worth it. I will get rid of it.

PR #461, opened 2023-06-22
HTTPCLIENT-2277: Implement must-understand directive according to RFC9111

arturobernalg:
This commit introduces changes to handle the 'must-understand' directive in accordance with the updated HTTP cache-related RFC 9111. The logic ensures that the cache only stores responses with status codes it understands when the 'must-understand' directive is present.

This implementation adheres to the following RFC guidance:

- A cache that understands and conforms to the requirements of a response's status code may cache it when the 'must-understand' directive is present.
- If the 'no-store' directive is present along with the 'must-understand' directive, the cache can ignore the 'no-store' directive if it understands the status code's caching requirements.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
@@ -307,10 +307,18 @@ class ResponseCachingPolicy {
             }
             return false;
         }
-        if (cacheControl.isNoStore()) {
+
+        if (cacheControl.isMustUnderstand() && cacheControl.isNoStore() && !understoodStatusCode(response.getCode())) {
+            // must-understand cache directive overrides no-store
+            LOG.debug("Response contains a status code that the cache does not understand, so it's not cacheable");
+            return false;
+        }
+
+        if (!cacheControl.isMustUnderstand() && cacheControl.isNoStore()) {
             LOG.debug("Response is explicitly non-cacheable per cache control directive");
             return false;
         }
+
 
         if (request.getRequestUri().contains("?")) {
             if (neverCache1_0ResponsesWithQueryString && from1_0Origin(response)) {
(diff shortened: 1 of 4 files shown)


ok2c (changes requested):
@arturobernalg I strongly disagree that status codes supported by the implementation is something that the caller can just configure. This is conceptually wrong. Please reconsider the overall approach.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -94,6 +94,42 @@ class ResponseCachingPolicy {
...
+     *     <li>304 Not Modified: The client has performed a conditional GET request and access is allowed, but the document has not been modified</li>
+     *     <li>400 Bad Request: The request could not be understood by the server due to malformed syntax</li>
+     *     <li>401 Unauthorized: The request requires user authentication</li>
+     *     <li>403 Forbidden: The server understood the request, but is refusing to fulfill it</li>
+     *     <li>404 Not Found: The server has not found anything matching the Request-URI</li>
+     *     <li>410 Gone: The requested resource is no longer available at the server and no forwarding address is known</li>
+     *     <li>500 Internal Server Error: The server encountered an unexpected condition which prevented it from fulfilling the request</li>
+     * </ul>
+     */
+    private static final Set<Integer> DEFAULT_UNDERSTOOD_STATUS_CODES =
```
ok2c, line 117: @arturobernalg There are other status codes that should be included here: all of 3xx, 4xx up to 417, 421, 5xx up to 505. I would also use a simple range checks there instead of a set contains for better performance and resource footprint. 

Looks good otherwise.

ok2c (changes requested):

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -495,4 +503,23 @@ boolean responseContainsNoCacheDirective(final ResponseCacheControl responseCach
...
+     * - All 2xx (Successful) status codes (200-299)
+     * - All 3xx (Redirection) status codes (300-399)
+     * - All 4xx (Client Error) status codes up to 417 and 421
+     * - All 5xx (Server Error) status codes up to 505
+     *
+     * @param status The HTTP status code to be checked.
+     * @return true if the HTTP status code is understood, false otherwise.
+     */
+    private boolean understoodStatusCode(final int status) {
+        return (status >= 200 && status <= 299)   ||    // All 2xx status codes
```
ok2c, line 518: @arturobernalg That is probably too bold. I would go with 200 up to 206 only.

ok2c (approved):
@arturobernalg Looks good.

PR #462, opened 2023-06-24
RFC8246  HTTP Immutable Responses

arturobernalg:
This commit adds the functionality to handle the 'immutable' directive in the Cache-Control header as per the RFC specifications.

Key changes include:
- The 'immutable' directive is checked in the Cache-Control of an HTTP response, indicating that the origin server will not update the resource representation during the response's freshness lifetime.
- If the 'immutable' directive is present and the response is still fresh, the response is considered cacheable without further validation.
- Ignoring any arguments with the 'immutable' directive, as per RFC stipulations.
- Treating multiple instances of the 'immutable' directive as equivalent to one.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
@@ -231,8 +231,18 @@ class ResponseCachingPolicy {
             return false;
         }
 
-        // calculate freshness lifetime
         final Duration freshnessLifetime = calculateFreshnessLifetime(cacheControl, response);
+
+        // If the 'immutable' directive is present and the response is still fresh,
+        // then the response is considered cacheable without further validation
+        if (cacheControl.isImmutable() && responseIsStillFresh(response, freshnessLifetime)) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Response is immutable and fresh, considered cacheable without further validation");
+            }
+            return true;
+        }
+
+        // calculate freshness lifetime
         if (freshnessLifetime.isNegative() || freshnessLifetime.isZero()) {
             if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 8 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -231,8 +231,18 @@ public boolean isResponseCacheable(final ResponseCacheControl cacheControl, fina
             return false;
         }
 
-        // calculate freshness lifetime
         final Duration freshnessLifetime = calculateFreshnessLifetime(cacheControl, response);
+
```
ok2c, line 235: @arturobernalg I cannot find `immutable` cache control directive mentioned anywhere in the RFC 9111. Am I missing something?

arturobernalg:
You're absolutely right. This change pertains to the [rfc8246](https://www.rfc-editor.org/rfc/rfc8246.html).

ok2c:
> @ok2c. You're absolutely right. This change pertains to the [rfc8246](https://www.rfc-editor.org/rfc/rfc8246.html).

@arturobernalg Could you please reflect this in the PR title and the commit message?

arturobernalg:
> > @ok2c. You're absolutely right. This change pertains to the [rfc8246](https://www.rfc-editor.org/rfc/rfc8246.html).
> 
> @arturobernalg Could you please reflect this in the PR title and the commit message?

@ok2c 
Done. 
The commit message doesn't need many changes. In any case, I've made a reference to the RFC.

ok2c:
@arturobernalg Looks good to me, but can we drop `no-transform` out of the change-set? It is unrelated to the stated objective of the change-set and is presently not used anywhere in the caching logic.

arturobernalg:
> @arturobernalg Looks good to me, but can we drop `no-transform` out of the change-set? It is unrelated to the stated objective of the change-set and is presently not used anywhere in the caching logic.

@ok2c  Removed.

ok2c (approved):

PR #465, opened 2023-06-29
Mostly missing Javadoc in org.apache.hc.client5.http.entity

garydgregory:
Mostly missing Javadoc in org.apache.hc.client5.http.entity and minor clean ups


garydgregory:
> @garydgregory I would be nice to not mix javadoc and functional changes, however minor, in the same commit. Splitting this change-set into two commits would be nice. Are you planning on making more of similar changes?

Not at this time.

The change to reduce the nesting is not a functional change, the behavior is exactly the same.

PR #467, opened 2023-07-19
Per RFC 7616, use the provided Response algorithm in the Request

jvasileff:
For digest authentication, in RFC 7616 section "3.4 The Authorization
Header Field":

    The values of the opaque and algorithm fields must be those supplied
    in the WWW-Authenticate response header field for the entity being
    requested.

This commit honors that rule, and removes the previous behavior that
augmented the request header with "algorithm=MD5" when none was provided
in the server's response.

Aside from the specification, it also stands to reason that if the
server failed to provide "algorithm=..." in its "WWW-Authenticate"
header, the server should be fine with the client failing to provide
"algorithm=..." in the "Authorization" header.

The motivation for this change is that including "algorithm=MD5" in the
"Authorization" header causes http requests to fail when made to an
embedded system, which I suspect to be a an Espressif ESP32 web server.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
@@ -245,11 +245,7 @@ public class DigestScheme implements AuthScheme, Serializable {
         final String realm = this.paramMap.get("realm");
         final String nonce = this.paramMap.get("nonce");
         final String opaque = this.paramMap.get("opaque");
-        String algorithm = this.paramMap.get("algorithm");
-        // If an algorithm is not specified, default to MD5.
-        if (algorithm == null) {
-            algorithm = "MD5";
-        }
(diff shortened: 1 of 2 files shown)


arturobernalg:
Didn't RFC 2617 get deprecated and subsequently replaced by RFC 7616? this change looks like fallow the RFC 7616.

jvasileff:
Ah, yes, thanks. It looks like RFC 7616 obsoletes 2617.

In RFC 7616, relevant quotes from *3.4 The Authorization Header Field* supporting this change are:

> The values of the opaque and algorithm fields must be those supplied in the WWW-Authenticate response header field for the entity being requested.

and

> For historical reasons, a sender MUST NOT generate the quoted string syntax for the following parameters: algorithm, qop, and nc.

The second note about quoting is only relevant in that it rules out other potential changes - this PR makes no changes to quoting; the algorithm value is still produced without `"` tokens.

jvasileff:
@ok2c Done!

PR #468, opened 2023-07-20
Check for timeToLive == 0

ChristophKaser:
At the moment, when i pass validateAfterInactivity as Timeout.ZERO_MILLISECONDS, it has the effect of never validating the connection at all, because that is how Deadline.calculate works. However, the check for TimeValue.isNonNegative for validateAfterInactivity before the deadline check suggests this is not the intended behavior.

Furthermore, we are trying to communicate with a server that always closes a h2 connection after one request. Without a way to force revalidation every time, this runs into issues when a connection is taken from the pool.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -322,8 +322,8 @@ public class PoolingHttpClientConnectionManager
                     if (poolEntry.hasConnection()) {
                         final TimeValue timeToLive = connectionConfig.getTimeToLive();
                         if (TimeValue.isNonNegative(timeToLive)) {
-                            final Deadline deadline = Deadline.calculate(poolEntry.getCreated(), timeToLive);
-                            if (deadline.isExpired()) {
+                            if (timeToLive.getDuration() == 0
+                                    || Deadline.calculate(poolEntry.getCreated(), timeToLive).isExpired()) {
                                 poolEntry.discardConnection(CloseMode.GRACEFUL);
                             }
                         }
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -323,7 +323,7 @@ public synchronized ConnectionEndpoint get(
                         final TimeValue timeToLive = connectionConfig.getTimeToLive();
                         if (TimeValue.isNonNegative(timeToLive)) {
                             final Deadline deadline = Deadline.calculate(poolEntry.getCreated(), timeToLive);
-                            if (deadline.isExpired()) {
+                            if (timeValue.getDuration() == 0 || deadline.isExpired()) {
```
ok2c, line 326: @ChristophKaser 
1. What is the point of calculating the deadline if it is not going be used? This looks wrong. Please refactor the code to avoid unnecessary deadline calculation. There are various ways of accomplishing that
2. Please make symmetric changes to the `PoolingAsyncClientConnectionManager`
3. Please make sure that the build does not fail

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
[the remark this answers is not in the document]
ChristophKaser: Ouch, very sorry about that. I had made the changes to an internal fork and copied them into the vanilla project. I have corrected my mistake now.

ChristophKaser: 1. Done - I have inlined the deadline calculation now.
2. Done
3. Ouch, very sorry about that. I had made the changes to an internal fork and copied them into the vanilla project, but did not run `mvn verify` after that. I have corrected the typo (and this time, I did run mvn).

PR #473, opened 2023-08-06
Fixed SOCKS proxy example

vkuzel:
Use a custom socket factory to create a proxied socket.

Original implementation sets the proxy address via default socket config assuming it would somehow propagate into the socket factory:
```java
PoolingHttpClientConnectionManagerBuilder.create()
                 .setDefaultSocketConfig(SocketConfig.custom().setSocksProxyAddress(socksaddr)
```
Unfortunatelly, the `SocketConfig.socksProxyAddress` is not used in the [`DefaultHttpClientConnectionOperator.connect()`](https://github.com/apache/httpcomponents-client/blob/223669659c295a01afe9047e0910768781c8f157/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L155) method, which creates the socket.

This solution extends `SSLConnectionSocketFactory` to create a proxied socket.

### Potential improvements

* This works only for HTTPS connections. The `PoolingHttpClientConnectionManagerBuilder` does not have API for setting a "plain socket factory", so making this work for a HTTP is slightly more complicated.
* To make the "default socket config" work, we may change implementation in the `DefaultHttpClientConnectionOperator.connect()` but that would probably lead to change of the "socket factory" APIs.
* There's also the `HttpClientBuilder.setProxy()` which does not support SOCKS proxy at the moment. This makes the API slightly awkward.

### DIFF
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/examples/ClientExecuteSOCKS.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/examples/ClientExecuteSOCKS.java
@@ -27,25 +27,21 @@
 
 package org.apache.hc.client5.http.examples;
 
-import java.io.IOException;
-import java.net.InetSocketAddress;
-import java.net.Proxy;
-import java.net.Socket;
-
 import org.apache.hc.client5.http.classic.methods.HttpGet;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.client5.http.impl.classic.HttpClients;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManagerBuilder;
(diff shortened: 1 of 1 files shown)


ok2c:
@vkuzel The proposed changes are perfectly reasonable but what needs fixing is `DefaultHttpClientConnectionOperator` not the sample code (or not sample code alone). Please raise a JIRA for this defect. Unfortunately the fix will have to go into 5.3 as it is likely to require changes in the `ConnectionSocketFactory` interface as well.

vkuzel:
@ok2c I've created [HTTPCLIENT-2292](https://issues.apache.org/jira/browse/HTTPCLIENT-2292) ticket for this, currently unassigned. Feel free to assign it to me & I will try to implement a fix.

Closing this PR.

ok2c:
@vkuzel Please go ahead and propose a fix. Please note you need to introduce default methods to the existing interfaces such as `ConnectionSocketFactory` in order to extend them without breaking the backward compatibility with the existing inplementations

PR #474, opened 2023-08-08
Avoid loom-unfriendly monitors in PoolingHttpClientConnectionManager

MikailBag:
# Problem

1) Function in question calls [`conn.isStale()`](https://github.com/apache/httpcomponents-client/blob/223669659c295a01afe9047e0910768781c8f157/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java#L339), which is implemented in httpcore5 (`BHttpConnectionBase.isStale`) by running potentially blocking IO operation inside `fillInputBuffer`. (UPD: there are other potentially blocking operations in this function, such as logging and waiting for future)
2) Also, this function is marked as `synchronized`, hence it holds monitor lock on `this` while it executes.

Unfortunately, combination of this two facts makes this function loom-unfriendly: when virtual thread acquires connection from pool, it is unable to park gracefully (since it holds a monitor lock), so it blocks together with its carrier thread. Such a behavior temporary decreases parallelism, as carrier is simply blocked instead of executing other virtual threads.

# Solution
This PR simply replaces `synchronized` with a j.u.c ReentrantLock, and problem should be gone.
# Alternatives
- Do nothing and just wait until Loom limitation is lifted and virtual threads holding monitor locks may park. However, it is unclear to me when this improvement will arrive.
- I could find only one non-test usage of `LeaseRequest.get()` method - in `InternalExecRuntime.acquireEndpoint()` method, which calls it only once. Therefore it looks possible to specify that get() is never called concurrently and drop `sychronized` keyword without replacement. On the other hand it may be breaking API change. Also, this is first time I look at httpcomponents source code, so my analysis may be simply wrong.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -33,6 +33,8 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.hc.client5.http.DnsResolver;
 import org.apache.hc.client5.http.HttpRoute;
@@ -298,10 +300,21 @@ public class PoolingHttpClientConnectionManager
         final Future<PoolEntry<HttpRoute, ManagedHttpClientConnection>> leaseFuture = this.pool.lease(route, state, requestTimeout, null);
         return new LeaseRequest() {
 
+            private final Lock lock = new ReentrantLock();
             private volatile ConnectionEndpoint endpoint;
 
             @Override
-            public synchronized ConnectionEndpoint get(
+            public ConnectionEndpoint get(
+                    Timeout timeout) throws InterruptedException, ExecutionException, TimeoutException {
+                this.lock.lock();
+                try {
+                    return getImpl(timeout);
+                } finally {
+                    this.lock.unlock();
+                }
+            }
+
+            private ConnectionEndpoint getImpl(
                     final Timeout timeout) throws InterruptedException, ExecutionException, TimeoutException {
                 Args.notNull(timeout, "Operation timeout");
                 if (this.endpoint != null) {


arturobernalg:
> # Problem
> 
>     1. Function in question calls [`conn.isStale()`](https://github.com/apache/httpcomponents-client/blob/223669659c295a01afe9047e0910768781c8f157/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java#L339), which is implemented in httpcore5 (`BHttpConnectionBase.isStale`) by running potentially blocking IO operation inside `fillInputBuffer`. (UPD: there are other potentially blocking operations in this function, such as logging and waiting for future)
> 
>     2. Also, this function is marked as `synchronized`, hence it holds monitor lock on `this` while it executes.
> 
> 
> Unfortunately, combination of this two facts makes this function loom-unfriendly: when virtual thread acquires connection from pool, it is unable to park gracefully (since it holds a monitor lock), so it blocks together with its carrier thread. Such a behavior temporary decreases parallelism, as carrier is simply blocked instead of executing other virtual threads.
> # Solution
> 
> This PR simply replaces `synchronized` with a j.u.c ReentrantLock, and problem should be gone.
> # Alternatives
> 
>     * Do nothing and just wait until Loom limitation is lifted and virtual threads holding monitor locks may park. However, it is unclear to me when this improvement will arrive.
> 
>     * I could find only one non-test usage of `LeaseRequest.get()` method - in `InternalExecRuntime.acquireEndpoint()` method, which calls it only once. Therefore it looks possible to specify that get() is never called concurrently and drop `sychronized` keyword without replacement. On the other hand it may be breaking API change. Also, this is first time I look at httpcomponents source code, so my analysis may be simply wrong.

@MikailBag 

Could you please provide some insight into the handling of other synchronization blocks within the code? Are there any plans to review or modify them to ensure compatibility with Loom? The change should be consistent across the entire code base or not at all. I would leave 4.x be and make sure 5.x works correctly with the newest JREs.

ok2c:
@MikailBag I agree with @arturobernalg It makes no sense to apply such changes piecemeal. It should be done across the entire code base like with apache/httpcomponents-core#412

garydgregory:
We should tread very carefully here IMO. My vote is to do nothing until virtual threads, which are only out of beta in Java 21, are fully debugged and accepted. I don't see it as a positive to increasebour code size and complexity for "friendliness".

MikailBag:
> Could you please provide some insight into the handling of other synchronization blocks within the code?

This particular `synchronized` block pinning was actually caught in practice, so I changed it. I initially chose not to touch other blocks since I'm not sure if blocking inside them is possible or not.

Other than that, I have no objections to replacing all other blocks.

>  I would leave 4.x be and make sure 5.x works correctly with the newest JREs.

If I don't mix things up, this PR is opened against 5.x code.

> My vote is to do nothing until virtual threads, which are only out of beta in Java 21, are fully debugged and accepted.

In JDK early access releases they should be already usable without any `--enable-preview` flags.
Anyway, I guess I understand your position. Maybe this pull request should be closed/postponed until JDK 21 is released.

Thank you in advance for providing fast feedback!

ok2c:
@MikailBag I propose we close this PR and focus on https://issues.apache.org/jira/projects/HTTPCLIENT/issues/HTTPCLIENT-2290

MikailBag:
> I propose we close this PR and focus on https://issues.apache.org/jira/projects/HTTPCLIENT/issues/HTTPCLIENT-2290

To be honest, I don't see contradictions between this PR and the ticket description. Do you mean that this synchronized usage should be resolved in some other way, i.e. completely removing any synchronization? Or that this PR is too small in scope?

ok2c:
@MikailBag There is nothing wrong with the proposed change-set. It looks perfectly reasonable but we do not want to be fixing synchronized methods piecemeal. They all should be changed consistently across the entire code base in a single change-set similar to the one committed to HttpCore 5.3.x: https://github.com/apache/httpcomponents-core/commit/4d6d9f55631226756736807d4a19d793c75d9500

PR #475, opened 2023-08-16
HTTPCLIENT-2277: removal of deprecated cache functionality / logic that does not belong to the caching layer

ok2c:
This change-set includes
* removal of deprecated cache functionality
* removal of HTTP protocol logic that does not belong to the caching layer


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseProtocolCompliance.java
```diff
@@ -28,198 +28,51 @@
...
+
+/**
+ * This response interceptor removes common {@literal Content-} headers from the 304
+ * response messages based on the provision of the HTTP specification that 304
+ * status code represents a confirmation that the requested resource content has
+ * not changed since its retrieval or the last update and adds {@literal Date}
+ * header if it is not present in the response message..
+ */
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+class ResponseProtocolCompliance implements HttpResponseInterceptor {
```
arturobernalg, line 51: Is this change related to the guidelines in '[3.2. Updating Stored Header Fields'](https://www.rfc-editor.org/rfc/rfc9111.html#name-updating-stored-header-fiel) from the RFC9111 specification?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExecBase.java
```diff
@@ -187,8 +164,7 @@ SimpleHttpResponse generateCachedResponse(
             final ResponseCacheControl responseCacheControl,
```
arturobernalg, line 164: I think we can remove `responseCacheControl`

arturobernalg (changes requested):
@ok2c, it looks good. Just a few minor comments.

ok2c:
@arturobernalg All sounds reasonable. Please do another pass.

arturobernalg (approved):
The change-set looks good

PR #476, opened 2023-08-19
HTTPCLIENT-2290 - Refactor HttpClient synchronized sections for virtual threads

arturobernalg:
- Replaced `synchronized` blocks with `ReentrantLock` in `LeaseRequest` to better support virtual threads introduced in JDK 21.
- Ensured each `LeaseRequest` instance has its own unique lock for maintaining original synchronization semantics.
- Addressed potential performance and deadlock issues with virtual threads by using explicit lock primitives from `java.util.concurrent.locks`.


ok2c (approved):

PR #477, opened 2023-08-21
HTTPCLIENT-2284: Cache entry representation improvements

ok2c:
* request URI stored in cache is now normalized and is presently expected to be the same as the root key
* removed references to variant cache keys from the cache entry
* Variants in root entries are now represented as a set, not as a map

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/HttpByteArrayCacheEntrySerializer.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/HttpByteArrayCacheEntrySerializer.java
@@ -32,9 +32,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.time.Instant;
-import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
-import java.util.Map;
+import java.util.Set;
 
 import org.apache.hc.client5.http.async.methods.SimpleHttpResponse;
 import org.apache.hc.client5.http.cache.HttpCacheEntry;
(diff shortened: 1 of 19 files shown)


ok2c:
@arturobernalg Here's what I hope is going to be the last round of changes to the HttpCacheEntry and the related classes. This change-set should also help partially resolve HTTPCLIENT-2284 by making it easier to implement custom cleanup of evicted cache entries. 
Please review.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/HttpByteArrayCacheEntrySerializer.java
```diff
@@ -328,7 +330,7 @@ public HttpCacheStorageEntry deserialize(final byte[] serializedObject) throws R
                     statusLine.getStatusCode(),
                     responseHeaders,
                     resource,
-                    variantMap
+                    !variants.isEmpty() ? variants : null
```
arturobernalg, line 333: Shouldn't we delegate everything to the internal constructor of `HttpCacheEntry`? I think its redundant.

arturobernalg (approved):
@ok2c 
The change look good. I just add a small comment.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/HttpByteArrayCacheEntrySerializer.java
ok2c: > Shouldn't we delegate everything to the internal constructor of `HttpCacheEntry`? I think its redundant.

@arturobernalg I am not sure I understand. You mean the `variants.isEmpty()` check?

arturobernalg: @ok2c my bad.  No change need it.  Sorry about that

PR #478, opened 2023-08-26
HTTPCLIENT-2284: internal cache storage improvements

ok2c:
Added internal cache storage implementation that can be used to implement custom cache storage backends with specific cache entry eviction strategy.

@arturobernalg Please review.


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/InternalCacheStorage.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.hc.core5.annotation.Internal;
+
+@Internal
+final public class InternalCacheStorage {
+
+    private final Map<String, HttpCacheEntry> map;
+    private final Queue<HttpCacheStorageEntry> evictionQueue;
+    private final Consumer<HttpCacheStorageEntry> evictionCallback;
+
+    public InternalCacheStorage(final int maxEntries, final Consumer<HttpCacheStorageEntry> evictionCallback) {
```
arturobernalg, line 46: shouldn't we check if `maxEntries`is positives?

arturobernalg (approved):
@ok2c  the change-set looks good. I just add a small question.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/InternalCacheStorage.java
ok2c: @arturobernalg I would say not. Not at this level. Input validation should be done at the public API level,

arturobernalg: fair enough

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/InternalCacheStorage.java
```diff
@@ -0,0 +1,94 @@
...
+        this.evictionCallback = evictionCallback;
+        this.map = new LinkedHashMap<String, HttpCacheEntry>(20, 0.75f, true) {
+
+            @Override
+            protected boolean removeEldestEntry(final Map.Entry<String, HttpCacheEntry> eldest) {
+                if (evictionCallback != null && size() > maxEntries) {
```
lacompas, line 52: This is a change in behaviour, now the maxEntries size limit is no longer honoured by default

PR #480, opened 2023-08-31
Document exec chain behaviour when automatic retries are enabled.

cachescrubber:
Document classic and async exec chain behaviour when automatic retries are enabled.

https://issues.apache.org/jira/browse/HTTPCLIENT-2291

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/AsyncExecChainHandler.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/AsyncExecChainHandler.java
@@ -28,7 +28,9 @@ package org.apache.hc.client5.http.async;
 
 import java.io.IOException;
 
+import org.apache.hc.client5.http.impl.ChainElement;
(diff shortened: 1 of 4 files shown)


ok2c:
@cachescrubber Awesome! One thing, though. Do you mind dropping code snippets out of Javadocs, as code examples often tend to get out of sync with the actual API usage.

cachescrubber:
@ok2c I removed the snippets. Having implemented a handler myself I remember having to dig in the source of clientBuilder  in order to find the name of the retry handler. What about 

```
 * @see ChainElement#RETRY
 * @see HttpClientBuilder#addExecInterceptorAfter(String, String, ExecChainHandler)
 * @see HttpClientBuilder#addExecInterceptorBefore(String, String, ExecChainHandler)
```

I think I would at least document the name (ChainElement#RETRY) in order to place the interceptors relative to it. WDYT?

ok2c:
> I think I would at least document the name (ChainElement#RETRY) in order to place the interceptors relative to it. WDYT?

Makes perfect sense.

PR #482, opened 2023-09-13
HTTPCLIENT-2293 Enforce Port Specification for CONNECT Request

arturobernalg:
This PR introduces a change to the ProxyClient class to ensure that a port is always specified when making a CONNECT request, in line with the requirements set out in [RFC 9110.](https://www.rfc-editor.org/rfc/rfc9110#name-connect)

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ProxyClient.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ProxyClient.java
@@ -140,12 +140,10 @@ public class ProxyClient {
         Args.notNull(proxy, "Proxy host");
         Args.notNull(target, "Target host");
         Args.notNull(credentials, "Credentials");
-        HttpHost host = target;
-        if (host.getPort() <= 0) {
-            host = new HttpHost(host.getSchemeName(), host.getHostName(), 80);
-        }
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ProxyClient.java
```diff
@@ -140,20 +140,19 @@ public Socket tunnel(
         Args.notNull(proxy, "Proxy host");
         Args.notNull(target, "Target host");
         Args.notNull(credentials, "Credentials");
-        HttpHost host = target;
-        if (host.getPort() <= 0) {
-            host = new HttpHost(host.getSchemeName(), host.getHostName(), 80);
+        if (target.getPort() <= 0) {
```
ok2c, line 143: @arturobernalg It would be nicer (and more consistent) to use `Args#check` here but not a deal breaker.

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ProxyClient.java
arturobernalg: @ok2c  
True. changed.

PR #483, opened 2023-09-14
HTTPCLIENT-2277: internal cache API and cache update redesign

ok2c:
Changes:
* improved and simplified internal cache API
* corrected cache update logic after a variant negotiation
* more tests
* fixed regression


ok2c:
@arturobernalg You do not need to understand every detail in the change-set, just see if you can spot something obviously wrong or crazy

arturobernalg (approved):
The change-set looks good.

PR #484, opened 2023-09-17
HTTPCLIENT-2277:  deprecation of obsolete config parameters and removal of oudated or meaningless tests

ok2c:
@arturobernalg Please kindly have a look. No functional changes. Mostly deprecation and removal of outdated / unused code and some javadoc updates. Also removed verbatim RFC 2616 quotes from the cache test cases.


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheConfig.java
```diff
@@ -257,11 +232,11 @@ public int getMaxUpdateRetries(){
...
 
     /**
-     * Returns whether 303 caching is enabled.
-     * @return {@code true} if it is enabled.
+     * @deprecated No longer applicable. Do not use.
      */
+    @Deprecated
     public boolean is303CachingEnabled() {
-        return allow303Caching;
+        return true;
```
arturobernalg, line 239: @ok2c  Should the return value of the deprecated is303CachingEnabled() method be set to false to align with all the changes?

arturobernalg (commented):
@ok2c 
this change-set LGTM

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheConfig.java
ok2c: @arturobernalg The cache protocol layer will now be treating 303 responses as potentially cacheable by default. This is why I thought the deprecated method should also return `true`. Does that make sense?

arturobernalg: @ok2c  that its correct. my bad.

PR #485, opened 2023-09-20
HTTPCLIENT-2293 - Implement 'If-Range' request validation as per RFC9110

arturobernalg:
This pull request aims to introduce support for the "If-Range" header as specified in RFC 9110. The "If-Range" header provides a more efficient way to handle partial content retrieval and is a crucial step towards full support for Range requests in our HTTP client component.

**Note:** This is the first step towards comprehensive support for HTTP Range Requests in our project.


arturobernalg:
I would greatly appreciate feedback from @reschke on this implementation, especially concerning its compliance with RFC standards and any best practices that should be considered.

ok2c (approved):

PR #486, opened 2023-09-20
 HTTPCLIENT-2293 - Implement RFC-compliant TRACE request interceptor

arturobernalg:
-  Add RequestTraceInterceptor class to handle HTTP TRACE requests in compliance with RFC 7231, Section 4.3.8.
-  Throw ProtocolException for sensitive headers like 'Authorization' and 'Cookie' in TRACE requests.
-  Throw ProtocolException if TRACE request contains a body.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestTraceInterceptor.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestTraceInterceptor.java
@@ -1,0 +1,161 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


arturobernalg:
Please @reschke take a look

httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestTraceInterceptor.java
```diff
@@ -0,0 +1,143 @@
...
+        // Check if the request method is TRACE
+        if (Method.TRACE.isSame(request.getMethod())) {
+
+            //A client MUST NOT send content in a TRACE request.
+            if (entity != null) {
+                throw new ProtocolException("TRACE request MUST NOT contain a request body.");
+            }
+
+            // Check for sensitive headers
+            final Header authHeader = request.getHeader(HttpHeaders.AUTHORIZATION);
```
ok2c, line 122: @arturobernalg I really dislike the idea of removing headers explicitly set by the caller. It feels wrong. Would you be OK with throwing a protocol exception instead?

arturobernalg: @ok2c 

Fair enough. changed

ok2c (approved):

PR #488, opened 2023-09-25
Update MemcachedHttpCacheStorage.java

snehanie:
UpdateCAS does not catch timeout exceptions.

getUpdateCAS catches and throws ResourceIOException, but updateCas does not. Try and catch is important since BasicHttpCache would catch this.

https://github.com/apache/httpcomponents-client/blob/master/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/BasicHttpCache.java#L273-L286


ok2c (changes requested):
@snehanie Good catch. Looks reasonable. 
Please do give the PR a better title and please make sure the style checks pass.

arturobernalg:
As we're refining this class, it might be beneficial to catch exceptions in the methods `getForUpdateCAS`, `bulkRestore`, `restore`, and `store`. Specifically, I noticed that `OperationTimeoutException` is thrown but not caught within these methods. 
WDYT @snehanie

ok2c:
Superseded by 19ab1111c7e947a25afcfa84f1efd4a915e7a7e8

snehanie:
Thank you for the fix. I was going to raise another pull request and you beat me to it.

PR #489, opened 2023-09-27
HTTPCLIENT-2293: Disallow Automatic Retries on 403 Forbidden Status

arturobernalg:
This PR updates the HttpRequestRetryExec/Async class to disallow automatic retries when a 403 Forbidden status is received, in compliance with RFC 9110 guidelines.
According to [RFC 9110,](https://www.rfc-editor.org/rfc/rfc9110#name-403-forbidden) a 403 Forbidden status code indicates that the server understood the request but refuses to authorize it. Retrying the request is generally considered inappropriate and could lead to security vulnerabilities or other issues.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncHttpRequestRetryExec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncHttpRequestRetryExec.java
@@ -41,6 +41,7 @@ import org.apache.hc.core5.http.EntityDetails;
 import org.apache.hc.core5.http.HttpException;
 import org.apache.hc.core5.http.HttpRequest;
 import org.apache.hc.core5.http.HttpResponse;
+import org.apache.hc.core5.http.HttpStatus;
 import org.apache.hc.core5.http.nio.AsyncDataConsumer;
 import org.apache.hc.core5.http.nio.AsyncEntityProducer;
 import org.apache.hc.core5.http.nio.entity.DiscardingEntityConsumer;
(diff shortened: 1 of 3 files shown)


arturobernalg:
Please @reschke take a look

ok2c:
@arturobernalg The decision whether or not a request should be retried based on the response status is delegated to the `HttpRequestRetryStrategy`. The default implementation presently retries requests in case of 429 and 503 and nothing else. Requests with a 403 response never get retried by default. One however can still use a custom implementation of the strategy to make the protocol layer to retry requests differently. This is how we allow for non-standard or deviant or custom behaviors without toggles and nonsense like that. Your change-set would make it impossible to retry requests with a 403 response even if the caller intentionally would like to test the behavior of the opposite endpoint. One would need to replace the entire request interceptor in such a case instead of just providing a custom retry strategy. Is this reasonable?

arturobernalg:
> @arturobernalg The decision whether or not a request should be retried based on the response status is delegated to the `HttpRequestRetryStrategy`. The default implementation presently retries requests in case of 429 and 503 and nothing else. Requests with a 403 response never get retried by default. One however can still use a custom implementation of the strategy to make the protocol layer to retry requests differently. This is how we allow for non-standard or deviant or custom behaviors without toggles and nonsense like that. Your change-set would make it impossible to retry requests with a 403 response even if the caller intentionally would like to test the behavior of the opposite endpoint. One would need to replace the entire request interceptor in such a case instead of just providing a custom retry strategy. Is this reasonable?

@ok2c  You're completely right. I overlooked the flexibility regarding the http default codes provided by HttpRequestRetryStrategy. That was my mistake, and I see now how my changes could limit customization.

PR #490, opened 2023-10-01
Add security warning to TrustStrategy implementations documentation

Marcono1234:
Tries to improve the documentation of `TrustAllStrategy` and `TrustSelfSignedStrategy` by mentioning their security implications and suggesting more secure alternatives.

For `TrustSelfSignedStrategy` it also documents what exactly the strategy does. Otherwise without looking at the implementation details users might erroneously think it is always necessary to use `TrustSelfSignedStrategy` when using self-signed certificates (even if they are also in the truststore), which is not the case.

Any feedback is appreciated!

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/TrustAllStrategy.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/TrustAllStrategy.java
@@ -34,9 +34,15 @@ import org.apache.hc.core5.annotation.ThreadingBehavior;
 import org.apache.hc.core5.ssl.TrustStrategy;
 
 /**
- * A trust strategy that accepts all certificates as trusted. Verification of
- * all other certificates is done by the trust manager configured in the SSL
- * context.
+ * A trust strategy that accepts all certificates as trusted.
+ *
+ * <h2>Security Warning</h2>
+ * This trust strategy effectivels disables most security features of SSL / TLS,
+ * and allows man-in-the-middle attacks. If possible avoid this trust strategy
+ * and use more secure alternatives. For example, for self-signed certificates
+ * prefer specifying a keystore containing the certificate chain when calling
+ * the {@link org.apache.hc.core5.ssl.SSLContextBuilder} {@code loadTrustMaterial}
+ * methods.
  *
  * @since 4.5.4
  * @since 5.0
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/ssl/TrustAllStrategy.java
```diff
@@ -34,9 +34,15 @@
 import org.apache.hc.core5.ssl.TrustStrategy;
 
 /**
- * A trust strategy that accepts all certificates as trusted. Verification of
- * all other certificates is done by the trust manager configured in the SSL
- * context.
+ * A trust strategy that accepts all certificates as trusted.
+ *
+ * <h2>Security Warning</h2>
+ * This trust strategy effectivels disables most security features of SSL / TLS,
```
ok2c, line 40: @Marcono1234 Could you avoid using the term `most` in `effectively disables most security features of SSL / TLS` because it is absolutely arbitrary? The strategy effectively disables the trust verification but not any other features of TLS. Otherwise looks good to me.

ok2c:
@Marcono1234 I also propose `TrustSelfSignedStrategy` be deprecated in favor of `SSLContextBuilder`. The class is basically useless.

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/TrustAllStrategy.java
Marcono1234: Thanks, yes I can change that. Should it just be "disables security features" then, or should it say "disables <ins>some</ins> security features"?

Marcono1234:
> I also propose TrustSelfSignedStrategy be deprecated in favor of SSLContextBuilder. The class is basically useless.

Should I include this in this PR or create a separate one?

Also, should these changes be backported to 4.x; and in that case should I do that or do you want to do this afterwards?

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/TrustAllStrategy.java
ok2c: @Marcono1234 I would prefer "disables trust verification" but I can live with "disables some security features".

ok2c:
@Marcono1234 Just deprecate `TrustSelfSignedStrategy`in the same PR. 
I would not bother with a backport of the changes to 4.x but I will merge them if you raise a PR for the 4.x line.

Marcono1234:
Have pushed the changes, I hope that is what you had in mind.

PR #491, opened 2023-10-03
HTTPCLIENT-2293: Via header improvements

ok2c:
* `ResponseCachingPolicy` optimized to generate less intermediate garbage while paring `Via` headers
* `Via` request and response header generation moved from the caching exec interceptors into separate protocol interceptors


ok2c:
@arturobernalg Please take a look

arturobernalg (approved):
@ok2c LGTM

PR #492, opened 2023-10-05
HTTPCLIENT-2302: Add comment to TrustStrategy usage in examples

Marcono1234:
See also discussion in https://issues.apache.org/jira/browse/HTTPCLIENT-2302

This tries to make the user aware that:
- Usage of `TrustStrategy` is optional
- The specific usage in the examples does not prevent man-in-the-middle attacks

Feedback is appreciated!

### DIFF
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientCustomSSL.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientCustomSSL.java
@@ -59,6 +59,11 @@ public class AsyncClientCustomSSL {
     public static void main(final String[] args) throws Exception {
         // Trust standard CA and those trusted by our custom strategy
         final SSLContext sslContext = SSLContexts.custom()
+                // Specify a custom TrustStrategy
+                // Note that this is not needed when the certificate has been issued by a trusted CA,
+                // or when using a custom truststore which contains the certificate chain
+                // Warning: While this example is better than using TrustAllStrategy, it still allows
+                //          man-in-the-middle attacks
                 .loadTrustMaterial((chain, authType) -> {
                     final X509Certificate cert = chain[0];
                     return "CN=httpbin.org".equalsIgnoreCase(cert.getSubjectDN().getName());
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientCustomSSL.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientCustomSSL.java
@@ -60,10 +60,12 @@ public class AsyncClientCustomSSL {
         // Trust standard CA and those trusted by our custom strategy
         final SSLContext sslContext = SSLContexts.custom()
                 // Specify a custom TrustStrategy
-                // Note that this is not needed when the certificate has been issued by a trusted CA,
-                // or when using a custom truststore which contains the certificate chain
-                // Warning: While this example is better than using TrustAllStrategy, it still allows
-                //          man-in-the-middle attacks
+                // Custom TrustStrategy implementations are intended for verification of certificates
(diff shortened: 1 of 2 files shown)


httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientCustomSSL.java
```diff
@@ -59,6 +59,11 @@ public class AsyncClientCustomSSL {
     public static void main(final String[] args) throws Exception {
         // Trust standard CA and those trusted by our custom strategy
         final SSLContext sslContext = SSLContexts.custom()
+                // Specify a custom TrustStrategy
```
ok2c, line 62: @Marcono1234  Can we start off by saying something like that? " Custom `TrustStrategy`s are intended for verification of certificates whose CA is either non-trusted or is not present in the trust stores used by the system".

Can we re-word "While this example is better than using TrustAllStrategy ..." as "Validation of the server certificate without validation of the entire certificate chain is still preferred to completely disabling of trust verification, however ..." 

Feel free to re-phrase the sentences as you deem appropriate to make it sound more English.

Marcono1234: I think it would still be good to let the comment start with "Specify a custom TrustStrategy" as introduction (followed by your suggestion) to explain that this defines a TrustStrategy. Otherwise, due to usage of the lambda expression, this might not be immediately obvious.

----

> for verification of certificates whose CA is either non-trusted or is not present in the trust stores used by the system

Could this be reworded to something like this to point out using a custom truststore as alternative?

> for verification of certificates whose CA is not trusted by the system, and where specifying a custom truststore containing the certificate chain is not an option.

----

So would the following comment content be ok for you (though it is a bit verbose)?

> Specify a custom TrustStrategy
> Custom TrustStrategy implementations are intended for verification of certificates  whose CA is not trusted by the system, and where specifying a custom truststore containing the certificate chain is not an option.
> Validation of the server certificate without validation of the entire certificate chain is preferred to completely disabling trust verification, however this \*still allows man-in-the-middle attacks\*.

----

Also, should this use line comments (`//`) or would you prefer to have this in a single block comment (`/* ... */`)?

To me the wording you proposed sounds fine, but I am not a native English speaker.

ok2c: > So would the following comment content be ok for you (though it is a bit verbose)?

@Marcono1234 I disagree with your emphasis on the use of trust stores as somehow being the preferred method. It is not any more secure, it is just simpler, that is all. I personally would not phrase it this way, but I can live with it.

Marcono1234: > It is not any more secure, it is just simpler, that is all.

I was thinking in the context of trusted custom CAs, but you are right, there might be other cases where using a custom truststore instead of a `TrustStrategy` is not necessarily more secure (but probably also not less secure?).

> it is just simpler

To me it feels like sadly the opposite is the case; writing a custom `TrustStrategy` is easy (even if logic inside it might be complex), compared to all the overhead of obtaining the certificate chain of your server as file, reading the certificate files, importing them into a `KeyStore`, and then specifying the `KeyStore` as truststore.

And if you want to trust the certificate chain *in addition* to the CAs trusted by default by the JRE, then it becomes even more complicated because it is your responsibility as user to obtain a `KeyStore` containing these JRE root CAs and merging them, see [this question](https://stackoverflow.com/q/24555890).

Of course you could also consider modifying the `cacerts` file of the JDK installation but that approach is not portable.

> I personally would not phrase it this way, but I can live with it.

Feel free to adjust the changes on this branch (you have write access to this branch), or after you merged the changes.

PR #497, opened 2023-10-24
Refactor RequestIfRange class to use DateUtils for date parsing.

arturobernalg:
* Updates the RequestIfRange class to utilize DateUtils for parsing standard HTTP dates.
* Optimize time difference check in RequestIfRange with Instant API.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestIfRange.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestIfRange.java
@@ -30,10 +30,9 @@ package org.apache.hc.client5.http.protocol;
 
 
 import java.io.IOException;
-import java.time.Duration;
 import java.time.Instant;
-import java.time.format.DateTimeFormatter;
 
+import org.apache.hc.client5.http.utils.DateUtils;
 import org.apache.hc.core5.annotation.Contract;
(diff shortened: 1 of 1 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestIfRange.java
```diff
@@ -121,36 +113,21 @@ public void process(final HttpRequest request, final EntityDetails entity, final
...
         }
 
-        final Header dateHeader = request.getFirstHeader(HttpHeaders.DATE);
+        final Instant dateHeader = DateUtils.parseStandardDate(request, HttpHeaders.DATE);
 
         if (dateHeader == null) {
             return;
         }
 
+        final Instant lastModifiedHeader = DateUtils.parseStandardDate(request, HttpHeaders.LAST_MODIFIED);
```
ok2c, line 122: @arturobernalg `lastModifiedHeader` is likely not the best name now. `lastModified`?

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestIfRange.java
arturobernalg: @ok2c Agree.
Changed.

PR #498, opened 2023-10-31
Add Zstd support to ContentCompressionExec.

arturobernalg:
This update allows automatic selection of Zstd as a compression algorithm when available.

### DIFF
--- a/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestRedirects.java
+++ b/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestRedirects.java
@@ -27,9 +27,11 @@
 package org.apache.hc.client5.testing.sync;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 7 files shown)


garydgregory:
Why not do this through Commons Compress?

arturobernalg:
> Why not do this through Commons Compress?

In general  keep it simple. Adding a direct dependency on the Zstd library is a straightforward way to enable ``'Zstd' compression without the overhead of an additional module.

ok2c:
@arturobernalg [Zstd](https://github.com/luben/zstd-jni#zstd) appears to be [BSD Licensed](https://github.com/luben/zstd-jni/blob/master/LICENSE). It _should_ be compatible with ASLv2 but I am afraid this needs to be double-checked. 

Why indeed not using Apache Compress if it provides compatible compression codec?

garydgregory:
Hi all,

Commons Compress depends on the same lib BUT provides a facade API for all its archive and compressed file formats.

ok2c:
> Hi all,
> 
> Commons Compress depends on the same lib BUT provides a facade API for all its archive and compressed file formats.

@garydgregory Got it. Thank you.

arturobernalg:
> @arturobernalg [Zstd](https://github.com/luben/zstd-jni#zstd) appears to be [BSD Licensed](https://github.com/luben/zstd-jni/blob/master/LICENSE). It _should_ be compatible with ASLv2 but I am afraid this needs to be double-checked.
> 
> Why indeed not using Apache Compress if it provides compatible compression codec?

Well, my initial approach was to directly integrate the Zstd library to provide a  direct and potentially efficient solution. I see what i can do.

Thank you

ok2c:
@arturobernalg Could you please change the target branch to `master` so I could delete obsolete `5.4.x`?

PR #499, opened 2023-11-01
HTTPCLIENT-2305: SSLConnectionSocketFactory allows socket.connect to be decorated

carterkozak:
https://issues.apache.org/jira/browse/HTTPCLIENT-2305


httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java
```diff
@@ -239,19 +240,7 @@ public Socket connectSocket(
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Connecting socket to {} with timeout {}", remoteAddress, connectTimeout);
             }
```
carterkozak, line 242: This debug logging will make more sense in the new method, update incoming

ok2c (approved):

carterkozak:
Thanks for the review, @ok2c!

PR #502, opened 2023-11-04
HTTPCLIENT-2301. Fix stale connection handling in BasicHttpClientConnectionManager

arturobernalg:
The key change involves using the local `conn` variable, obtained from `internalEndpoint`.detach(), to ensure we're acting on the correct connection instance.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
@@ -369,6 +369,7 @@ public class BasicHttpClientConnectionManager implements HttpClientConnectionMan
         }
     }
 
+
     private InternalConnectionEndpoint cast(final ConnectionEndpoint endpoint) {
         if (endpoint instanceof InternalConnectionEndpoint) {
             return (InternalConnectionEndpoint) endpoint;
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
```diff
@@ -357,7 +357,19 @@ ManagedHttpClientConnection getConnection(final HttpRoute route, final Object st
                 this.conn = this.connFactory.createConnection(null);
                 this.created = System.currentTimeMillis();
             } else {
-                this.conn.activate();
+                // Check if the existing connection is stale
+                if (this.conn.isStale()) {
```
ok2c, line 361: @arturobernalg Should not the connection have already been checked for staleness in `#validate` line 355? Why doing again? Am I missing something?

arturobernalg: > Should not the connection have already been checked for staleness in `#validate` line 355? Why doing again? Am I missing something?

@ok2c 
The staleness check was only being performed when the deadline had expired. In our case, the deadline wasn't expired, so the connection's staleness wasn't being evaluated.

ok2c: @arturobernalg Now I am confused. The `#validate` method performs the check only if the period of inactivity exceeds `ValidateAfterInactivity` setting. This is intentional and by design. One does not want to execute the stale check too often due to its cost and impact on performance. The change you are proposing seems to be breaking that. Is that really what you want? If one wants the stale check done every time one can set `ValidateAfterInactivity` to -1

arturobernalg: Upon reviewing your comment, I realize that I overlooked the intentional design behind the validate method's frequency of execution I've committed a new potential fix this. I have now submitted a new commit that aims to rectify this. The updated code ensures that the connection being released is indeed the one that was detached from the `InternalConnectionEndpoint`. This should maintain the intended behavior while addressing the issue at hand.
While I have conducted extensive testing to validate this fix, I must admit that connection management is not my primary area of expertise.

ok2c: @arturobernalg There is no problem of what so ever. That is why we have reviews

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
```diff
@@ -390,22 +391,15 @@ public void release(final ConnectionEndpoint endpoint, final Object state, final
...
                 this.updated = System.currentTimeMillis();
-                if (!this.conn.isOpen() && !this.conn.isConsistent()) {
-                    this.route = null;
-                    this.conn = null;
-                    this.expiry = Long.MAX_VALUE;
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug("{} Connection is not kept alive", id);
-                    }
-                } else {
+                final boolean reusable = conn != null && conn.isOpen() && conn.isConsistent();
```
ok2c, line 398: @arturobernalg The variable seems redundant. Looks OK otherwise.

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/BasicHttpClientConnectionManager.java
arturobernalg: changed.

PR #506, opened 2023-11-20
Enforce UTF-8 Encoding in BasicSchemeFactory for RFC 7617 Compliance

arturobernalg:
- Deprecated the constructor in BasicSchemeFactory that allows setting a custom Charset.
- Updated the default constructor to use StandardCharsets.UTF_8, aligning with RFC 7617 which mandates UTF-8 encoding for Basic Authentication.

### DIFF
--- a/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AuthenticatingAsyncDecorator.java
+++ b/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AuthenticatingAsyncDecorator.java
@@ -28,7 +28,6 @@ package org.apache.hc.client5.testing.async;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
@@ -80,7 +79,7 @@ public class AuthenticatingAsyncDecorator implements AsyncServerExchangeHandler 
     }
 
     public AuthenticatingAsyncDecorator(final AsyncServerExchangeHandler exchangeHandler, final Authenticator authenticator) {
-        this(exchangeHandler, new BasicAuthenticationHandler(StandardCharsets.US_ASCII), authenticator);
+        this(exchangeHandler, new BasicAuthenticationHandler(), authenticator);
     }
 
     protected void customizeUnauthorizedResponse(final HttpResponse unauthorized) {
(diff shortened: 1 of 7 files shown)


michael-o (approved):

ok2c (approved):

reschke:
FWIW, RFC 7617 does *not* require UTF-8, because back when it was written, that was considered a breaking change, as many UAs defaulted to ISO-8859.

That said, UAs have change since then, and a future revision will likely recommend using UTF-8.

arturobernalg:
> FWIW, RFC 7617 does _not_ require UTF-8, because back when it was written, that was considered a breaking change, as many UAs defaulted to ISO-8859.
> 
> That said, UAs have change since then, and a future revision will likely recommend using UTF-8.

Hi @reschke 
This PR's changes to the charset field adhere to RFC 7617, specifically section [2.1](https://datatracker.ietf.org/doc/html/rfc7617#section-2.1 ensuring compliance with 'UTF-8' as the sole allowed value

reschke:
I may be missing something, but my understandig was that the "charset" parameter wasn't supported in the first place. (Note that it was *introduced* in RFC 7617, and never was allowed to specify anything but UTF-8).

RFC 7617's revision will likely deprecate or remove this parameter, because a) it was never widely implemented and b) it doesn't appear to be needed anymore.

stanio:
>   This PR's changes to the charset field adhere to RFC 7617, specifically section [2.1](https://datatracker.ietf.org/doc/html/rfc7617#section-2.1) ensuring compliance with 'UTF-8' as the sole allowed value

This appears to refer to _server_ challenges (`WWW-Authenticate`).  For clients, the intro of the 2.0 section states:

>   The original definition of this authentication scheme failed to specify the character encoding scheme used to convert the user-pass into an octet sequence. In practice, most implementations chose either a locale-specific encoding such as [ISO-8859-1](https://datatracker.ietf.org/doc/html/rfc7617#ref-ISO-8859-1) ([ISO-8859-1]), or UTF-8 ([[RFC3629](https://datatracker.ietf.org/doc/html/rfc3629)]). For backwards compatibility reasons, this specification continues to leave the default encoding undefined, as long as it is compatible with US-ASCII (mapping any US-ASCII character to a single octet matching the US-ASCII character code).

arturobernalg:
> I may be missing something, but my understanding was that the "charset" parameter wasn't supported in the first place. (Note that it was _introduced_ in RFC 7617, and never was allowed to specify anything but UTF-8).
> 
> RFC 7617's revision will likely deprecate or remove this parameter, because a) it was never widely implemented and b) it doesn't appear to be needed anymore.

@reschke  our implementation actively uses the `charset`  parameter in line with RFC 7617, defaulting to `UTF-8` for encoding credentials, demonstrating its practical utility despite limited widespread adoption. Please take a look https://github.com/apache/httpcomponents-client/blob/5.4.x/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicScheme.java#L225

reschke:
Ok, interesting. FWIW, the code back then allows charsets other than UTF-8; it should have ignored those.

That said, have you ever seen a server send that?

PR #508, opened 2023-11-23
Enforce Support for UTF-8 Encoding Scheme in Digest Authentication as per RFC 7616

arturobernalg:
This commit enforces the use of the 'UTF-8' encoding scheme as the sole allowed value for character encoding in Digest Authentication, in alignment with the guidelines specified in RFC 7616.


ok2c:
@arturobernalg We have been providing an option to override `US-ASCII` as the standard HTTP message char-set for decades, which is by far a greater deviation from the standard protocol. I see no reason why we could not have that option for user name / password in the standard auth schemes. But so be it.

ok2c (approved):

PR #509, opened 2023-11-23
Implement Support for Userhash Parameter in Digest Authentication as per RFC 7616

arturobernalg:
This commit introduces support for the userhash parameter in Digest Authentication, conforming to the specifications outlined in RFC 7616. The userhash parameter enhances security by allowing the client to hash the username before transmission, thereby protecting the username during transport. This implementation ensures that when the server indicates support for username hashing (userhash=true), the client correctly calculates and includes the hashed username in the Authorization header field, adhering to the protocol defined in RFC 7616 for enhanced security in HTTP Digest Access Authentication.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
@@ -113,6 +113,24 @@ public class DigestScheme implements AuthScheme, Serializable {
     private boolean complete;
     private transient ByteArrayBuilder buffer;
 
+    /**
+     * Flag indicating whether username hashing is supported.
+     * <p>
+     * This flag is used to determine if the server supports hashing of the username
+     * as part of the Digest Access Authentication process. When set to {@code true},
+     * the client is expected to hash the username using the same algorithm used for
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
```diff
@@ -488,4 +525,40 @@ public String toString() {
...
+     *
+     * @param values The array of characters to be parsed. This array should
+     *               contain the characters of the string with potential surrounding
+     *               quotes.
+     * @return A String containing the characters from the input array with
+     * surrounding quotes removed. The method specifically skips the first
+     * and last quote characters if they exist, assuming that the entire
+     * string is enclosed in quotes.
+     */
+    private String parseAndRemoveQuotesAndDelimiters(final char[] values) {
```
ok2c, line 545: @arturobernalg I do not think this method is necessary. I believe `AuthChallengeParser` takes care of quote enclosed values. Please double-check.

arturobernalg: Absolutely right, @ok2c . I've verified and removed the unnecessary method. 

Thanks for pointing it out!

ok2c (approved):

PR #510, opened 2023-11-24
Add validation to prevent both 'username' and 'username*' in DigestScheme headers.

arturobernalg:
This commit introduces a new validation check in the ´DigestScheme ´class to ensure that both 'username' and 'username*' are not present simultaneously in the same authentication header. According to [RFC 7616](https://datatracker.ietf.org/doc/html/rfc7616#section-3.4), having both of these fields in the same header should be treated as an error.

### DIFF
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/impl/auth/TestDigestScheme.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/impl/auth/TestDigestScheme.java
@@ -746,4 +746,25 @@ public class TestDigestScheme {
         Assertions.assertEquals(digestScheme.getCnonce(), authScheme.getCnonce());
     }
 
+
+    @Test
+    public void testErrorOnBothUsernameAndUsernameStar() throws Exception {
+        final ClassicHttpRequest request = new BasicClassicHttpRequest("Post", "/");
+        request.setEntity(new InputStreamEntity(new ByteArrayInputStream(new byte[] {'a'}), -1, ContentType.DEFAULT_TEXT));
+        final HttpHost host = new HttpHost("somehost", 80);
+        final CredentialsProvider credentialsProvider = CredentialsProviderBuilder.create()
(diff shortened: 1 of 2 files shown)


ok2c (approved):

httpclient5/src/test/java/org/apache/hc/client5/http/impl/auth/TestDigestScheme.java
```diff
@@ -746,4 +746,25 @@ public void testSerialization() throws Exception {
...
+    @Test
+    public void testErrorOnBothUsernameAndUsernameStar() throws Exception {
+        final ClassicHttpRequest request = new BasicClassicHttpRequest("Post", "/");
+        request.setEntity(new InputStreamEntity(new ByteArrayInputStream(new byte[] {'a'}), -1, ContentType.DEFAULT_TEXT));
+        final HttpHost host = new HttpHost("somehost", 80);
+        final CredentialsProvider credentialsProvider = CredentialsProviderBuilder.create()
+                .add(new AuthScope(host, "realm1", null), "username", "password".toCharArray())
+                .build();
+
+        final String challenge = StandardAuthScheme.DIGEST + " realm=\"realm1\", nonce=\"f2a3f18799759d4f1a1c068b92b573cb\", " +
```
ok2c, line 759: @arturobernalg Please note this scenario is not very practical. The `username` and `username*` parameters apply to an auth response, not to an auth challenge. 
I approved the change-set but in all honesty the check is not very useful.

ok2c: > @ok2c, I understand the practicality aspect now. Could you please advise on how best to handle this test to ensure it aligns more closely with common scenarios involving the username and username* parameters? Any suggestions or examples would be greatly appreciated.

@arturobernalg This check makes sense on the server side, which we have no support for. It makes little sense on the client side. I would drop the change-set altogether.

PR #511, opened 2023-11-24
Implement username* validation and decoding in DigestScheme

arturobernalg:
Introduces validation and decoding logic for the 'username*' field in the DigestScheme class. The changes ensure compliance with RFC 7616 and RFC 5987 by handling cases where the 'username' contains characters not allowed in an ABNF quoted-string.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
@@ -341,17 +341,30 @@ public class DigestScheme implements AuthScheme, Serializable {
         }
         buffer.charset(charset);
 
+        a1 = null;
+        a2 = null;
 
+
+        // Extract username and username*
         String username = credentials.getUserName();
+        String encodedUsername = null;
+        // Check if 'username' has invalid characters and use 'username*'
+        if (username != null && containsInvalidABNFChars(username)) {
+            encodedUsername = "UTF-8''" + RFC5987Codec.encode(username, StandardCharsets.UTF_8);
+        }
 
+        final String usernameForDigest;
         if (this.userhashSupported) {
             final String usernameRealm = username + ":" + realm;
             final byte[] hashedBytes = digester.digest(usernameRealm.getBytes(StandardCharsets.UTF_8));
-            username = formatHex(hashedBytes); // Convert to hex string
+            usernameForDigest = formatHex(hashedBytes); // Use hashed username for digest
+            username = usernameForDigest;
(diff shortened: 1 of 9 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
```diff
@@ -314,20 +315,34 @@ private String createDigestResponse(final HttpRequest request) throws Authentica
 
         a1 = null;
         a2 = null;
+
+
+        // Extract username and username*
+        String username = credentials.getUserName();
+        final String usernameStar = paramMap.get("username*");
```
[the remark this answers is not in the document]
arturobernalg: HI @ok2c 
you're right. Please do another pass.
TY

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/RFC5987Codec.java
```diff
@@ -0,0 +1,155 @@
...
+
+        while (cb.hasRemaining()) {
+            final char c = cb.get();
+            if (c == '%' && cb.remaining() >= 2) {
+                final int u = Character.digit(cb.get(), RADIX);
+                final int l = Character.digit(cb.get(), RADIX);
+                if (u != -1 && l != -1) {
+                    bb.put((byte) ((u << 4) + l));
+                } else {
+                    throw new IllegalArgumentException("Invalid percent encoding in " + s);
```
ok2c, line 146: @arturobernalg Can we throw a better exception here, like `CodingException` of some sort, which could be caught at the a higher level and translated in a proper protocol exception?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/RFC5987Codec.java
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Utility class for encoding and decoding strings according to RFC 5987.
+ * This class provides methods to percent-encode and decode strings, particularly
+ * useful for handling HTTP header parameters that include non-ASCII characters.
+ *
+ * @Internal This class is intended for internal use within the library and
+ * should not be used as part of the public API.
+ */
+@Internal
+class RFC5987Codec {
```
ok2c, line 47: @arturobernalg There is a similar implementation in the `HttpRFC7578Multipart` class. Could you please get rid of it in this or a follow-up change-set?

ok2c (approved):

arturobernalg:
@ok2c  please take another look.

ok2c (approved):

PR #512, opened 2023-11-26
HTTPCLIENT-2292: fix regression leading to invalid proxy

mmoayyed:
[This](https://issues.apache.org/jira/browse/HTTPCLIENT-2292) seems to have caused a regression in 5.2.2.

 
```java
java.lang.IllegalArgumentException: Invalid Proxy
    at java.base/java.net.Socket.<init>(Socket.java:216)
    at org.apache.hc.client5.http.ssl.SSLConnectionSocketFactory.createSocket(SSLConnectionSocketFactory.java:208)
    at org.apache.hc.client5.http.impl.io.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:158)
    at org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:447)
    at org.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:162)
    at org.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:172) 
```

When the proxy is undefined or set to null, DefaultHttpClientConnectionOperator does the following:

 
```java
Timeout soTimeout = socketConfig.getSoTimeout();
SocketAddress socksProxyAddress = socketConfig.getSocksProxyAddress();
Proxy proxy = socksProxyAddress != null ? new Proxy(Type.SOCKS, socksProxyAddress) : null; 
```

The proxy object above is then null, which means the following call:

```java
Socket sock = sf.createSocket(proxy, context); 
```

...will eventually pass down to SSLConnectionSocketFactory which does this:

```java
public Socket createSocket(Proxy proxy, HttpContext context) throws IOException {
    return new Socket(proxy); //this proxy cannot be null
} 
```

I should mention that in the above case,

No SOCKS proxy is ever configured via the PoolingHttpClientConnectionManagerBuilder. setDefaultSocketConfig() method 
No  SocketConfig.setSocksProxyAddress(socksaddr) exists and is always null.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java
@@ -155,7 +155,7 @@ public class DefaultHttpClientConnectionOperator implements HttpClientConnection
             final InetAddress address = remoteAddresses[i];
             final boolean last = i == remoteAddresses.length - 1;
 
-            Socket sock = sf.createSocket(proxy, context);
+            Socket sock = proxy != null ? sf.createSocket(proxy, context) : sf.createSocket(context);
             if (soTimeout != null) {
                 sock.setSoTimeout(soTimeout.toMillisecondsIntBound());
             }


ok2c:
Fixed by 6d60624cd3439f81030f7443e72cba21662e5d7b

mmoayyed:
Thank you!

garydgregory:
FWIW, I ran into the same regression. Staying on 5.2.1 for now...

Sineaggi:
Unfortunately this is still causing issues for projects such as docker-java (see https://github.com/docker-java/docker-java/issues/2290). Currently this bug manifests as a backwards-imcompatible change to projects inheriting from PlainConnectionSocketFactory and SSLConnectionSocketFactory.

I've made an example fix https://github.com/apache/httpcomponents-client/pull/535 to pull the proxy null check higher out

PR #513, opened 2023-11-28
Invalid Argument Exception thrown when null proxy passed to Socket contrctor

SquirrelGrip:
Setting the proxy to NO_PROXY when proxy if null instead.


ok2c:
@SquirrelGrip Fixed by 6d60624cd3439f81030f7443e72cba21662e5d7b

PR #514, opened 2023-11-28
Fixed a problem when a null proxy is passed to the constructor of a Socket.

SquirrelGrip:
Instead passed a Proxy.NO_PROXY instead.


ok2c:
@SquirrelGrip The change-set looks wrong. Anyway, the defect has already been fixed by 7d899b1948fe3494bb8cb11c5474e1babab19041

PR #515, opened 2023-11-29
Improve debug logging around connections that cannot be reused

nfawcett:
This commit is to improve debug logging around connections that are not reusable. This [previous commit ](https://github.com/apache/httpcomponents-client/commit/9cfdd54c94ab641a887e43ca1f4ae56beab5cd90) introduced usage of the [isConsistent](https://github.com/apache/httpcomponents-core/blob/d1c77bc7f7697548cd83d04ba87a1ec7105da308/httpcore5/src/main/java/org/apache/hc/core5/http/io/HttpClientConnection.java#L54) variable. 

Upon upgrading from v4.5 -> 5.x I witnessed increased latency between version due to connections being closed and it not being obvious why. I eventually pinpointed to this change, so improving the debug logging around this to hopefully help others in the future.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -409,7 +409,7 @@ public class PoolingHttpClientConnectionManager
                 }
             } else {
                 if (LOG.isDebugEnabled()) {
-                    LOG.debug("{} connection is not kept alive", ConnPoolSupport.getId(endpoint));
+                    LOG.debug("{} connection is not kept alive(isConsistent:{})", ConnPoolSupport.getId(endpoint), conn.isConsistent());
                 }
             }
         } catch (final RuntimeException ex) {


httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -408,8 +408,9 @@ public void release(final ConnectionEndpoint endpoint, final Object state, final
                     LOG.debug("{} connection {} can be kept alive {}", ConnPoolSupport.getId(endpoint), ConnPoolSupport.getId(conn), s);
                 }
             } else {
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug("{} connection is not kept alive", ConnPoolSupport.getId(endpoint));
+                if (this.log.isDebugEnabled()) {
```
ok2c, line 411: @nfawcett Where does `this.log` come from? I do not see  `log` instance variable anywhere.

ok2c (changes requested):
@nfawcett I will merge in this change but please re-base  your code off `master` and make sure the build passes.

PR #516, opened 2023-11-30
 HTTPCLIENT-2310: Async Connect exec handler incorrectly pipes CONNECT requests through the main request protocol chain

ok2c:
This change-set fixes the problem reported in HTTPCLIENT-2310


arturobernalg (approved):
@ok2c 
LGTM

PR #519, opened 2023-12-09
Fix: HTTP response content being truncated to a maximum length of 256 bytes

Demonese:
see commit: 0df9e63932ea4dca63009ad52254bbb3fbf0b86e


Demonese:
> Would you please write a test that demonstrates the problem?

Of course, you can compare whether the test case passes before/after apply fix.

ok2c:
Superseded by 7d853d521e8741aa1967b216fe4dfbe425a21efa

PR #520, opened 2023-12-09
Prevent NullPointerException being thrown when constructing log message

topbadger:
https://github.com/apache/httpcomponents-client/pull/515 added additional logging information about the connection being released. However the log exists in a branch where the connection may be null, thus causing a NPE to be thrown under those circumstances.

This PR accounts for the potential for the connection to be null when constructing the log message.


ok2c:
@topbadger Thank you for reporting the problem. Corrected by 9e3d79bedeee494d24bdb03cc598bae5184a3abf.

PR #529, opened 2023-12-24
Regression: incorrect hash calculation (missing realm component)

ok2c:


ok2c:
@arturobernalg There is a regression in the Digest auth scheme. The realm component is missing in the A1 value leading to incorrect hash value calculation.

Generally when seeing many existing test methods failing, one should pause and investigate the cause of it instead of just changing test case expectations.

arturobernalg:
@ok2c 
I apologize for the oversight. You're absolutely right about the importance of addressing failing tests before modifying them.

PR #533, opened 2024-01-07
HTTPCLIENT-2314: Handle gracefully a failure of DnsResolver to return a list of resolved addresses

phax:
Make sure an `UnknownHostException` is thrown, even if a custom `DnsResolver` implementation is used.
Fixes the issue of HTTPCLIENT-2314


ok2c (changes requested):
@phax 
1. Please remove my changes from your change-set. If the check is moved to the connection operator it not necessary or even useless in the `SystemDefaultDnsResolver`.
2. Please make symmetric changes to the `MultihomeIOSessionRequester`

phax:
@ok2c Thanks for the review. Done

PR #535, opened 2024-01-10
Fixes createSocket being called with null proxy when proxy is null

Sineaggi:
This is a follow-up to this fix https://github.com/apache/httpcomponents-client/commit/6d60624cd3439f81030f7443e72cba21662e5d7b where the null is handled in the implementation.

IMHO the null-check should be pulled out higher, when createSocket is called. As it stands, there's no way for `createSocket(HttpContext` to be called anymore. Any implementation of `ConnectionSocketFactory` by way of extending/inheriting existing implementations (SSL or Plain) would get caught by having to override both createSocket(HttpContext) and createSocket(Proxy, HttpContext). This is the case for docker-java, and is causing this issue https://github.com/docker-java/docker-java/issues/2290

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
@@ -63,7 +63,7 @@ public class PlainConnectionSocketFactory implements ConnectionSocketFactory {
 
     @Override
     public Socket createSocket(final Proxy proxy, final HttpContext context) throws IOException {
-        return proxy != null ? new Socket(proxy) : new Socket();
+        return new Socket(proxy);
     }
 
     @Override
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java
@@ -206,7 +206,7 @@ public class SSLConnectionSocketFactory implements LayeredConnectionSocketFactor
 
     @Override
     public Socket createSocket(final Proxy proxy, final HttpContext context) throws IOException {
-        return proxy != null ? new Socket(proxy) : new Socket();
+        return new Socket(proxy);
     }
 
     @Override
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java
@@ -160,7 +160,7 @@ public class DefaultHttpClientConnectionOperator implements HttpClientConnection
             final InetAddress address = remoteAddresses[i];
             final boolean last = i == remoteAddresses.length - 1;
 
-            Socket sock = sf.createSocket(proxy, context);
+            Socket sock = proxy != null ? sf.createSocket(proxy, context) : sf.createSocket(context);
             if (soTimeout != null) {
                 sock.setSoTimeout(soTimeout.toMillisecondsIntBound());
             }
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/impl/io/TestBasicHttpClientConnectionManager.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/impl/io/TestBasicHttpClientConnectionManager.java
@@ -383,7 +383,7 @@ public class TestBasicHttpClientConnectionManager {
         Mockito.when(dnsResolver.resolve("somehost")).thenReturn(new InetAddress[] {remote});
         Mockito.when(schemePortResolver.resolve(target)).thenReturn(8443);
         Mockito.when(socketFactoryRegistry.lookup("https")).thenReturn(plainSocketFactory);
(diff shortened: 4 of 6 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
```diff
@@ -63,7 +63,7 @@ public PlainConnectionSocketFactory() {
 
     @Override
     public Socket createSocket(final Proxy proxy, final HttpContext context) throws IOException {
-        return proxy != null ? new Socket(proxy) : new Socket();
```
ok2c, line 66: @Sineaggi This looks wrong. The check for the proxy being non null should be performed by the method.

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java
```diff
@@ -206,7 +206,7 @@ public Socket createSocket(final HttpContext context) throws IOException {
 
     @Override
     public Socket createSocket(final Proxy proxy, final HttpContext context) throws IOException {
-        return proxy != null ? new Socket(proxy) : new Socket();
```
ok2c, line 209: @Sineaggi Same.

ok2c (changes requested):

ok2c:
> As it stands, there's no way for createSocket(HttpContext to be called anymore

@Sineaggi So what? This method will get removed in the next major release. I do not see how this necessitates any additional checks at the higher level. Individual socket factories must be able to handle `null` proxy parameters gracefully.

httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
twarner-sofi: Then perhaps this code should be

```
return proxy != null ? new Socket(proxy) : createSocket(context);
```

As it stands, any overridden implementation of `createSocket(HttpContext)` is ignored, which is a breaking change.

twarner-sofi: In the interface, yes, but that delegation is not reflected in PlainConnectionSocketFactory's implementation of the method. This class is not final and thus users can extend it, overriding createSocket. If that is incorrect usage and this class was meant to be final, I think that should at least be documented in the class's javadoc.

ok2c: @twarner-sofi While `PlainConnectionSocketFactory` is non final and can be extended, one _must_ be aware of risks of tight coupling through inheritance and risks of subclassing a concrete class and overriding methods that were not meant for extension in the first place. Do not pin this problem on us.

twarner-sofi: Okay, thanks for confirming that these methods are not meant for extension in the first place. It's unfortunate that they weren't marked `final`, or documented as such, but that's fine now.

We'll look into making a fix PR for docker-java.

Sineaggi:
> > As it stands, there's no way for createSocket(HttpContext to be called anymore
> 
> @Sineaggi So what? This method will get removed in the next major release.

Where does it say that this method will be removed? I couldn't find anything in the javadocs or release notes.

> Individual socket factories must be able to handle `null` proxy parameters gracefully.

Again, I couldn't find any documentation mentioning as such. Presently the wording would imply createSocket that doesn't take a proxy parameter would be called if there was no proxy.

httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
Sineaggi: > one must be aware of risks of tight coupling through inheritance and risks of subclassing a concrete class and overriding methods that were not meant for extension in the first place

The class in question has no private fields and no hidden internal logic, and has existed virtually unmodified since 2012. Without warning, this public and non-final class has had a breaking change made to its behavior in a patch release of this library.

Sineaggi:
If going forward the design of the connection factory class ought to use the new `createSocket(proxy, context)` method, for the sake of existing users of the library would you be alright with modifying the existing overloads of createSocket in SSL and PlainConnectionFactory to call the previous createSocket method on null?

For example, like the following
```java
httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java

return proxy != null ? new Socket(proxy) : createSocket(context);
```

ok2c:
> would you be alright with modifying the existing overloads of createSocket in SSL and PlainConnectionFactory to call the previous createSocket method on null?

@Sineaggi Sure, I would. Feel free to raise a PR with the proposed changes.

httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
ok2c: @Sineaggi There is no breaking change. The class remains fully API compatible. Any particular modality or implementation details have never been a part of the public contract. You abused class inheritance in your code.

Sineaggi:
@ok2c Here's the PR https://github.com/apache/httpcomponents-client/pull/536 with the proposed changes.

PR #536, opened 2024-01-16
Allow backwards-compatible createSocket calls

Sineaggi:
In order to allow for backwards-compatibility with projects that override/extend the existing plain and ssl connection factories, we'll specify the createSocket(Proxy, HttpContext) method to call the original createSocket method when proxy is null.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java
@@ -63,7 +63,7 @@ public class PlainConnectionSocketFactory implements ConnectionSocketFactory {
 
     @Override
     public Socket createSocket(final Proxy proxy, final HttpContext context) throws IOException {
-        return proxy != null ? new Socket(proxy) : new Socket();
+        return proxy != null ? new Socket(proxy) : createSocket(context);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


ok2c:
@Sineaggi Also cherry-picked to `5.3.x`

Sineaggi:
@ok2c thanks!

ok2c:
@Sineaggi I intend to deprecate the entire `ConnectionSocketFactory` layer that goes back to the days of HttpClient v2 and harmonize the TLS setup logic of  the classic transport with the async one, possibly by providing a single TLS strategy that works with both transports. If you are interested in reviewing or collaborating on the new APIs do let me know.

arturobernalg:
> @Sineaggi I intend to deprecate the entire `ConnectionSocketFactory` layer that goes back to the days of HttpClient v2 and harmonize the TLS setup logic of the classic transport with the async one, possibly by providing a single TLS strategy that works with both transports. If you are interested in reviewing or collaborating on the new APIs do let me know.

@ok2c  I'd be interested in collaborating on the new APIs and reviewing the harmonization of the TLS setup logic across transports.

ok2c:
> @ok2c I'd be interested in collaborating on the new APIs and reviewing the harmonization of the TLS setup logic across transports.

@arturobernalg Sure. The proposed changes will go through the standard review process.

ok2c:
@Sineaggi Please see #540. Do feel free to let me know if you want anything done differently to ensure API compatibility with your specific application.

PR #543, opened 2024-01-29
SNI support improvements

ok2c:
* Connection route info now request URI authority
* The default route planner and connections managers now take request URI authority when calculating the connection route and when pooling and re-using connections


ok2c:

arturobernalg (approved):
@ok2c LGTM

PR #544, opened 2024-01-30
HTTPCLIENT-2070: Auth cache to no longer rely on Java serialization for auth state caching

ok2c:
No more Java serialization in HttpClient

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicScheme.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicScheme.java
@@ -26,13 +26,9 @@
  */
 package org.apache.hc.client5.http.impl.auth;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.charset.UnsupportedCharsetException;
 import java.security.Principal;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicScheme.java
```diff
@@ -232,27 +228,47 @@ public String generateAuthResponse(
...
 
     @Override
     public String toString() {
         return getName() + this.paramMap;
     }
 
+    @Internal
+    public static class State {
+
+        final Map<String, String> params;
```
arturobernalg, line 258: I assume this variables  are not private due to its internal scope.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BearerScheme.java
```diff
@@ -161,9 +163,49 @@ public String generateAuthResponse(
...
+
     @Override
     public String toString() {
         return getName() + this.paramMap;
     }
 
+    @Internal
+    public static class State {
+
+        final Map<String, String> params;
```
arturobernalg, line 193: I assume this variables  are not private due to its internal scope.

arturobernalg (approved):
@ok2c 
LGTM

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicScheme.java
ok2c: @arturobernalg Correct.

PR #545, opened 2024-02-01
HTTPCLIENT-2151: Support for JSSE in-built endpoint identification

ok2c:
This change-set extends the public APIs with hostname verification options:
* `Built-In`: Hostname verification is delegated to the JSSE provider, usually executed during the TLS handshake 
* `Client`: Hostname verification is executed by HttpClient post TLS handshake
* `Both`: Hostname verification is executed by the JSSE provider and by HttpClient post TLS handshake

The `Both` option is used by default. It provides the benefit of backward compatibility with the behavior of previous versions of HttpClient, added security (hostnames get verified by the JSSE provider and by the client) and fail-early approach (invalid hostnames can get rejected during the TLS handshake). At the same some checks may get performed twice and result in certain (likely minor) performance hit. In the future HttpClient may switch to `Built-In` as the default.


ok2c:

arturobernalg (approved):
@ok2c 
LGTM

PR #552, opened 2024-02-23
Redesign of client HTTP context APIs

ok2c:
* Aligns client context APIs with those in core
* Context classes are now plain beans
* Backward compatibility with older versions (through internal Delegate classes)
* Javadoc imprvements

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/CacheResponseStatus.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/CacheResponseStatus.java
@@ -28,10 +28,7 @@ package org.apache.hc.client5.http.cache;
 
 /**
  * This enumeration represents the various ways a response can be generated
- * by the caching {@link org.apache.hc.client5.http.classic.HttpClient};
- * if a request is executed with an {@link HttpCacheContext}
- * then a parameter with one of these values will be registered in the
- * context under the key  {@link HttpCacheContext#CACHE_RESPONSE_STATUS}.
+ * by an HTTP cache.
  */
 public enum CacheResponseStatus {
 
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheContext.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheContext.java
@@ -26,50 +26,102 @@
  */
 package org.apache.hc.client5.http.cache;
 
+import java.util.Map;
+
+import javax.net.ssl.SSLSession;
+
+import org.apache.hc.client5.http.HttpRoute;
+import org.apache.hc.client5.http.RouteInfo;
+import org.apache.hc.client5.http.auth.AuthCache;
+import org.apache.hc.client5.http.auth.AuthExchange;
+import org.apache.hc.client5.http.auth.AuthScheme;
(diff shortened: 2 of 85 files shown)


ok2c:

arturobernalg (approved):
@ok2c 
LGTM

sratz:
This change is incompatible.

Consider when a *consumer* sets the attributes on the context:

```java
BasicCookieStore cookieStore = new BasicCookieStore();
HttpClientContext context = new HttpClientContext();
context.setAttribute(HttpClientContext.COOKIE_STORE, cookieStore); // <----
CloseableHttpClient client = HttpClients.createDefault();
client.execute(new HttpGet("https://httpbin.org/cookies/set/foo/bar"), context, response -> null);
System.out.println(cookieStore.getCookies());
```

httpclient < 5.4 correctly uses the request-specific cookie store and prints:
```[[name: foo; value: bar; domain: httpbin.org; path: /; expiry: null]]```

httpclient 5.4 does not care about the attribute anymore ([diff](https://github.com/apache/httpcomponents-client/pull/552/files#diff-16278c1c7854154453531a554fbcaf624a76d7b6ca1e95eb529148f361fd9348R179-R180)) and prints:
```[]```

Switching from
```
context.setAttribute(HttpClientContext.COOKIE_STORE, cookieStore);
```
to
```
context.setCookieStore(cookieStore);
```
does not solve the problem, as that would work in 5.4, but not in 5.3 anymore.

The only chance I see to support both is to do *both*
```
context.setAttribute(HttpClientContext.COOKIE_STORE, cookieStore);
context.setCookieStore(cookieStore);
```

Shouldn't this have been compatible? I.e. ([here](https://github.com/apache/httpcomponents-client/pull/552/files#diff-16278c1c7854154453531a554fbcaf624a76d7b6ca1e95eb529148f361fd9348R179-R180)), consider *both* the getter and the attribute?

ok2c:
> does not solve the problem, as that would work in 5.4, but not in 5.3 anymore.

@sratz  Why would not that work in 5.3?

sratz:
> > does not solve the problem, as that would work in 5.4, but not in 5.3 anymore.
> 
> @sratz Why would not that work in 5.3?

Because there the reverse is true, `org.apache.hc.client5.http.impl.classic.InternalHttpClient.setupContext(HttpClientContext)` only looks at the attributes, not at the getter methods:
https://github.com/apache/httpcomponents-client/blob/4f852608c3315975bf6cd72d10999626784cd9a6/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/InternalHttpClient.java#L123-L138

sratz:
I think the problem is that `HttpClientContext` never kept the well-known named attributes and their equivalent object member in sync, i.e.

`setAttribute("http.cookie-store", ...)` and `setCookieStore()` have always referred to two completely independent things.

ok2c:
> only looks at the attributes, not at the getter methods:

@sratz `context.setCookieStore(cookieStore);` will work correctly regardless of the version. What changed in 5.4 is that standard attributes no longer get shoved into a map.

sratz:
> @sratz `context.setCookieStore(cookieStore);` will work correctly regardless of the version. What changed in 5.4 is that standard attributes no longer get shoved into a map.

Ok, you're right.

However, there's still a breaking change that the string-based attributes are now not considered at all anymore in 5.4. So they are not just deprecated.

PR #553, opened 2024-02-28
HTTPCLIENT-2321 Clarify default implementation of HttpClientConnectio…

kwin:
…nManager

Add table with default implementations for all relevant component.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
@@ -70,6 +70,7 @@ import org.apache.hc.client5.http.impl.auth.BasicSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.BearerSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.DigestSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.SystemDefaultCredentialsProvider;
+import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManagerBuilder;
 import org.apache.hc.client5.http.impl.routing.DefaultProxyRoutePlanner;
 import org.apache.hc.client5.http.impl.routing.DefaultRoutePlanner;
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
```diff
@@ -134,7 +135,27 @@
  * exclusive and may not apply when building {@link CloseableHttpClient}
  * instances.
  * </p>
- *
+ * The following are the default implementations per component
```
ok2c, line 138: @kwin In all honesty next time I am going to touch `HttpClientBuilder` I am not sure I will get around to updating this table. It will ultimately rot and will cause more harm than good.
  
In any case If you want this change-set committed, please at the very least update `HttpAsyncClientBuilder`, `CachingHttpClientBuilder` and `CachingHttpAsyncClientBuilder`

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
kwin: I know this is a lot of effort. The other alternative would be to establish a common naming (e.g. always use `Default` prefix, but this won't be backwards compatible) or at least clarify in the javadoc of the implementation whether it is the default one or not. WDYT?

ok2c: @kwin As I mentioned in the JIRA I am not in favor of documenting the defaults in javadocs because some people might see them as a part of public contract but you do it consistently across the code base I will review and commit the PR.

kwin: You already have that all over the place:

1. https://github.com/apache/httpcomponents-client/blob/f2b9a374d817ee630a2efe706546ba4f1e5a5ac4/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicAuthCache.java#L49
2. https://github.com/apache/httpcomponents-client/blob/f2b9a374d817ee630a2efe706546ba4f1e5a5ac4/httpclient5/src/main/java/org/apache/hc/client5/http/cookie/BasicCookieStore.java#L45
3. https://github.com/apache/httpcomponents-client/blob/f2b9a374d817ee630a2efe706546ba4f1e5a5ac4/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/BasicCredentialsProvider.java#L40
4. https://github.com/apache/httpcomponents-client/blob/f2b9a374d817ee630a2efe706546ba4f1e5a5ac4/httpclient5/src/main/java/org/apache/hc/client5/http/impl/DefaultRedirectStrategy.java#L50
5. https://github.com/apache/httpcomponents-client/blob/f2b9a374d817ee630a2efe706546ba4f1e5a5ac4/httpclient5/src/main/java/org/apache/hc/client5/http/impl/DefaultConnectionKeepAliveStrategy.java#L45
....

And yes, you are right, I consider this default part of the contract (i.e. changing it is possible but only in a minor version explicitly stating the change somewhere)

PR #557, opened 2024-03-20
Update the method parameter from String to CharSequence.

wangkaish:
Update the method parameter of InetAddressUtils from String to CharSequence.

Sometimes I need to pass parts of a String to this method instead of splitting the String into multiple pieces.

### DIFF
--- a/httpclient/src/main/java/org/apache/http/conn/util/InetAddressUtils.java
+++ b/httpclient/src/main/java/org/apache/http/conn/util/InetAddressUtils.java
@@ -71,34 +71,34 @@ public class InetAddressUtils {
     /**
      * Checks whether the parameter is a valid IPv4 address
      *
-     * @param input the address string to check for validity
+     * @param input the address character sequence to check for validity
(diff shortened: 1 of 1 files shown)


garydgregory:
@wangkaish 
You'll need to _add_ tests that exercise CharSequence parameter values.

wangkaish:
Hi @garydgregory 
String is an implementation of CharSequence, and the test argument used in TestInetAddressUtils is String, so I don't think we need a test case with a CharSequence argument anymore.

garydgregory:
Changing the API from a class to an interface has no such behavioral guarantee, even if the API binary compatibility is maintained. 

IOW there is no proof that the methods actually work with known implementing clases like CharBuffer, StringBuffer, StringBuilder, and so on.

ok2c:
@wangkaish There will be no more changes to 4.5.x. Please port the proposed change-set to `master`.

wangkaish:
@ok2c ok, InetAddressUtils moved to httpcomponents-core, I will create a new pr to httpcomponents-core and close this pr.

PR #559, opened 2024-05-15
HTTPCLIENT-2327: Propagate CancellationExceptions from MemcachedClient operations as ResourceIOExceptions

jattisha:
fixes https://issues.apache.org/jira/browse/HTTPCLIENT-2327

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/memcached/MemcachedHttpAsyncCacheStorage.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/memcached/MemcachedHttpAsyncCacheStorage.java
@@ -163,7 +163,7 @@ public class MemcachedHttpAsyncCacheStorage extends AbstractBinaryAsyncCacheStor
             } catch (final ExecutionException ex) {
                 if (ex.getCause() instanceof Exception) {
                     if (ex.getCause() instanceof CancellationException) {
-                        callback.failed(new ResourceIOException("Cache operation was cancelled mid-flight", ex.getCause()));
+                        callback.failed(new MemcachedOperationCancellationException(ex.getCause()));
                     } else {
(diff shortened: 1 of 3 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/memcached/MemcachedHttpAsyncCacheStorage.java
```diff
@@ -161,7 +162,11 @@ private <T> Cancellable operation(final OperationFuture<T> operationFuture, fina
                 callback.completed(operationFuture.get());
             } catch (final ExecutionException ex) {
                 if (ex.getCause() instanceof Exception) {
-                    callback.failed((Exception) ex.getCause());
+                    if (ex.getCause() instanceof CancellationException) {
+                        callback.failed(new ResourceIOException("Cache operation was cancelled mid-flight", ex.getCause()));
```
jattisha, line 166: Now that I'm reading this, I should probably create an `OperationCancelledException` that extends `ResourceIOException` and use this for all of these...

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/memcached/MemcachedOperationCancellationException.java
```diff
@@ -0,0 +1,42 @@
...
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl.cache.memcached;
+
+import org.apache.hc.client5.http.cache.ResourceIOException;
+
+/**
+ * Signals memcached operation cancellation.
+ */
```
ok2c, line 33: @jattisha Could please add either  `@since 5.4` or `@Internal` tag? Looks good otherwise.

PR #561, opened 2024-06-19
Fix HTTPCLIENT-2331: Skip cookie header generation if a Cookie header is present

arturobernalg:
Fix for [HTTPCLIENT-2331:](https://issues.apache.org/jira/browse/HTTPCLIENT-2331) Single Cookie Header for Multiple Cookies

When an explicit Cookie header is added using `setHeader()` and multiple cookies exist in the `BasicCookieStore`, `HttpClient` was sending separate Cookie headers for the manually set cookies and the cookies in the store. This fix ensures that if a Cookie header is already present in the request, `HttpClient` skips adding additional Cookie headers from the `BasicCookieStore`. 
This change conforms to the expectation that a single Cookie header should be sent where all cookies are separated by a semicolon, as per [RFC6265](https://www.rfc-editor.org/rfc/rfc6265).


ok2c (approved):

garydgregory (approved):
Nice test 👌

PR #563, opened 2024-07-11
Fix behaviour of PublicSuffixMatcher to match https://github.com/publicsuffix

massdosage:
This PR is just for internal discussion currently


massdosage:
Apologies, this is still a work in progress and I meant to submit the PR against our fork and not the origin. I'm closing this, will create a proper PR in future when we're ready, with more context.

PR #566, opened 2024-08-15
Update PublicSuffixMatcher to use "formal algorithm"

massdosage:
See https://issues.apache.org/jira/browse/HTTPCLIENT-2336 for context.

After porting over the tests for the formal algorithm from https://github.com/publicsuffix/list/blob/master/tests/test_psl.txt we then proceeded to make a number of changes to get them all to pass. Since this a non-backwards compatible (but hopefully more correct) change in behaviour we have introduced the concept of a "legacy mode" which could be enabled to ease a transition between behaviours but we have favoured the new behaviour and adjusted the existing tests accordingly.

There are obviously different ways we could approach this so would appreciate input and suggestions.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
@@ -156,7 +156,7 @@ public final class PublicSuffixMatcher {
             }
             final DomainType domainRule = findEntry(rules, key);
             if (match(domainRule, expectedType)) {
-                // HUMAN code change - PSL algorithm doesn't have any rules changing the result based on "domain type"
+                // PSL algorithm doesn't have any rules changing the result based on "domain type"
                 // see https://github.com/publicsuffix/list/wiki/Format#formal-algorithm
                 if (legacyMode && domainRule == DomainType.PRIVATE) {
                     return segment;
@@ -167,24 +167,7 @@ public final class PublicSuffixMatcher {
             final int nextdot = segment.indexOf('.');
             final String nextSegment = nextdot != -1 ? segment.substring(nextdot + 1) : null;
 
-            if (!legacyMode) {
-                // HUMAN code change - looks for wildcard entries and don't change the result based on "domain type"
-                // Check for wildcard entries.
-                final String wildcardKey = (nextSegment == null) ? "*" : "*." + IDN.toUnicode(nextSegment);
-                final DomainType wildcardDomainRule = findEntry(rules, wildcardKey);
-                if (match(wildcardDomainRule, expectedType)) {
-                    return result;
-                }
-
-                // If we're out of segments and we're not looking for a specific type of entry,
-                // apply the default `*` rule.
-                // This wildcard rule means any final segment in a domain is a public suffix,
-                // so the current `result` is the desired public suffix plus 1
-                if (nextSegment == null && (expectedType == null || expectedType == DomainType.UNKNOWN)) {
-                    return result;
-                }
-            } else {
-                // HUMAN code change, above code has replaced below commented out section
+            if (legacyMode) {
                 if (nextSegment != null) {
                     final DomainType wildcardDomainRule = findEntry(rules, "*." + IDN.toUnicode(nextSegment));
                     if (match(wildcardDomainRule, expectedType)) {
@@ -194,13 +177,27 @@ public final class PublicSuffixMatcher {
                         return result;
                     }
                 }
+            } else {
+                // look for wildcard entries and don't change the result based on "domain type"
+                final String wildcardKey = (nextSegment == null) ? "*" : "*." + IDN.toUnicode(nextSegment);
+                final DomainType wildcardDomainRule = findEntry(rules, wildcardKey);
+                if (match(wildcardDomainRule, expectedType)) {
+                    return result;
+                }
+
+                // If we're out of segments, and we're not looking for a specific type of entry,
+                // apply the default `*` rule.
+                // This wildcard rule means any final segment in a domain is a public suffix,
+                // so the current `result` is the desired public suffix plus 1
+                if (nextSegment == null && (expectedType == null || expectedType == DomainType.UNKNOWN)) {
+                    return result;
+                }
             }
 
             result = segment;
             segment = nextSegment;
         }
 
-        // HUMAN code change - commented out this case as now covered by new code within loop above
         // If no expectations then this result is good.
         if (legacyMode && (expectedType == null || expectedType == DomainType.UNKNOWN)) {
             return result;
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
@@ -208,18 +208,10 @@ public final class PublicSuffixMatcher {
     }
 
     /**
-     * Tests whether the given domain matches any of entry from the public suffix list.
+     * Tests whether the given domain matches any of the entries from the public suffix list.
      */
     public boolean matches(final String domain) {
-        return matches(domain, null, false);
-    }
-
-    public boolean matches(final String domain, final boolean legacyMode) {
-        return matches(domain, null, legacyMode);
-    }
-
-    public boolean matches(final String domain, final DomainType expectedType) {
-        return matches(domain, expectedType, false);
+        return matches(domain, null);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


httpclient5/src/test/java/org/apache/hc/client5/http/impl/cookie/TestPublicSuffixListParser.java
```diff
@@ -111,7 +111,7 @@ void testParseLocal() {
 
         cookie.setDomain(".blah");
         cookie.setAttribute(Cookie.DOMAIN_ATTR, ".blah");
-        Assertions.assertTrue(filter.match(cookie, new CookieOrigin("somehost.blah", 80, "/stuff", false)));
+        Assertions.assertFalse(filter.match(cookie, new CookieOrigin("somehost.blah", 80, "/stuff", false)));
```
massdosage, line 114: One shouldn't be allowed to set a cookie for the "blah" TLD as this is a potential security risk so we inverted the assertion as we believe the new behaviour is more secure.

httpclient5/src/test/java/org/apache/hc/client5/http/ssl/TestDefaultHostnameVerifier.java
```diff
@@ -258,11 +259,11 @@ void testIdentityMatching() {
         Assertions.assertTrue(DefaultHostnameVerifier.matchIdentity("a.b.xxx.uk", "a.b.xxx.uk", publicSuffixMatcher));
         Assertions.assertTrue(DefaultHostnameVerifier.matchIdentityStrict("a.b.xxx.uk", "a.b.xxx.uk", publicSuffixMatcher));
 
-        Assertions.assertTrue(DefaultHostnameVerifier.matchIdentity("a.b.xxx.uk", "*.b.xxx.uk", publicSuffixMatcher));
-        Assertions.assertTrue(DefaultHostnameVerifier.matchIdentityStrict("a.b.xxx.uk", "*.b.xxx.uk", publicSuffixMatcher));
+        Assertions.assertFalse(DefaultHostnameVerifier.matchIdentity("a.b.xxx.uk", "*.b.xxx.uk", publicSuffixMatcher));
```
massdosage, line 262: Having a wildcard certificate for a whole public suffic ("b.xx.uk" in this case) feels like it shouldn't be allowed but we don't know enough about the rules of SSL certs to know for sure. For now we have updated the tests to pass according to the PSL algorithm but would value further input here.

garydgregory (commented):
I'm wondering if the legacy mode should be an enum instead of a boolean. This would avoid futher API changes in case there are more changes needed. Let's see what @ok2c thinks.

ok2c:
> I'm wondering if the legacy mode should be an enum instead of a boolean. 

@massdosage Let's simplify things. I do not think we need to preserve the old legacy behavior, especially it is wrong. Please drop the parameter and the related conditional code.

massdosage:
@garydgregory @ok2c thanks for the prompt initial review. I've removed the `legacyMode` parameter as requested.

ok2c:
@massdosage I am fine with your changes to the `PublicSuffixMatcher` class. However I do think we must preserve the original behavior of `#matchIdentity` method of `DefaultHostnameVerifier`. At the very least all tests involving `amazonaws.com` need to retain their original behavior as they are based on the real-life usage scenarios reported by HttpClient users.

ok2c:
@massdosage These are the additional changes that I have made on top of your pull request: ok2c/httpcomponents-client@77aefb63901e242e28681172423bae462380d5f7. Please take a look. If they look reasonable to you, I will merge your pull request and apply my changes on top of it.

massdosage:
@ok2c thanks, I took a look, we haven't used the `DefaultHostnameVerifier` component so I'm not entirely sure of its intended uses, especially with respect to wildcard certs which appears to be the issue here. Is the intention to implement https://datatracker.ietf.org/doc/html/rfc9525#security-wildcards and https://en.wikipedia.org/wiki/Public_key_certificate? If so I think the examples at https://en.wikipedia.org/wiki/Public_key_certificate#Further_examples could be good ones to aim for in the unit tests. The current examples with "ec2.compute-1.amazonaws.com" are a bit confusing as I'm pretty sure AWS would never issue a wildcard cert of that form.

ok2c:
@massdosage RFC 9525 is a new document which HttpClient does not conform to yet. What we need to ensure that HttpClient conforms to [RFC2818](https://datatracker.ietf.org/doc/html/rfc2818#section-3.1)

> The current examples with "ec2.compute-1.amazonaws.com" are a bit confusing as I'm pretty sure AWS would never issue a wildcard cert of that form

But it looks like they did. Please see https://issues.apache.org/jira/browse/HTTPCLIENT-2247

massdosage:
OK, for RFC2818 it looks like one just needs to implement the wildcard at the "level" in the domain name it's specified but not any "lower". One could do that with just string matching and not even involve the PublicSuffixList. Either way, you know more about the code here and why it is the way it currently is and what future plans are so I'm happy to go with this being merged and your changes being applied afterwards if that works for you. I'm also happy to help out here and in future if there is anything related I can do.

ok2c:
@massdosage Committed as 715ec110600c2f0ee5c4fa03085306db175afba5

massdosage:
Great, thank you! 

We'll keep an eye out for the `5.4-beta2` and `5.4` releases so we can then switch over to this and retire our internal forked version.

garydgregory:
Hello @massdosage 

This PR apparently is reference on the ML at https://lists.apache.org/thread/mwhzx1n9v3317fj6x050nkozvdkszvo9 as the breaking change for local builds on my PC.

With the HEAD of git master, running:
```
mvn clean generate-resource 
mvn
```
I get:

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.hc.client5.http.psl.TestPublicSuffixMatcher
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.944 s <<< FAILURE! - in org.apache.hc.client5.http.psl.TestPublicSuffixMatcher
[ERROR] org.apache.hc.client5.http.psl.TestPublicSuffixMatcher.testGetDomainRootPublicSuffixList  Time elapsed: 0.019 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <null> but was: <uk.com>
        at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at org.apache.hc.client5.http.psl.TestPublicSuffixMatcher.checkPublicSuffix(TestPublicSuffixMatcher.java:171)
        at org.apache.hc.client5.http.psl.TestPublicSuffixMatcher.testGetDomainRootPublicSuffixList(TestPublicSuffixMatcher.java:208)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at java.util.ArrayList.forEach(ArrayList.java:1259)
        at java.util.ArrayList.forEach(ArrayList.java:1259)
```
Running on:
```
Apache Maven 3.9.8 (36645f6c9b5079805ea5009217e36f2cffd34256)
Maven home: C:\java\apache-maven-3.9.8
Java version: 17.0.12, vendor: Eclipse Adoptium, runtime: C:\Program Files\Eclipse Adoptium\jdk-17.0.12.7-hotspot
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
```

Please advise.

Gary

massdosage:
Let's discuss on the dev mailing list to keep things in one place if that's OK?

garydgregory:
> Let's discuss on the dev mailing list to keep things in one place if that's OK?

👍

httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
```diff
@@ -149,24 +148,30 @@ public String getDomainRoot(final String domain, final DomainType expectedType)
             }
             final DomainType domainRule = findEntry(rules, key);
             if (match(domainRule, expectedType)) {
-                if (domainRule == DomainType.PRIVATE) {
```
ok2c, line 152: @massdosage Unfortunately this change introduced a regression in 5.4.x. There is a reason for special treatment of `DomainType.PRIVATE`, we just did not have a test case for it in our test suite.

massdosage: Yes, I did say that this was a backwards-incompatible change as the previous behaviour didn't match the formal algorithm behaviour. We only focused on that aspect of it, if the `PublicSuffixListMatcher` is being used for other uses cases then I can't comment on their correctness as we're using this on the understanding that it implements the upstream standard. We're not using it for SSL/TLS etc. which I think are out of scope of the standard.

ok2c: @massdosage I did a bad job reviewing your changes so the fault is mine and mine only. The specification actually _does_ explicitly state that 
```
While some applications, such as browsers when considering cookie-setting, treat all entries the same, other applications may wish to treat ICANN domains and PRIVATE domains differently
```

PR #567, opened 2024-08-16
Input streams returned by DecompressingEntity.getContent() should support mark(int) and reset()

garydgregory:
Input streams returned by DecompressingEntity.getContent() should support mark(int) and reset() if the underlying stream does

- Pass calls to the underlying InputStream for mark(int)
- Pass calls to the underlying InputStream for reset()
- LazyDecompressingInputStream extends FilterInputStream
- Reimplement internals fluently
- Reuse Closer.close(Closeable)
- Tested locally with git master 5.3-beta2-SNAPSHOT

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
@@ -26,15 +26,18 @@
  */
 package org.apache.hc.client5.http.entity;
 
+import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
+
+import org.apache.hc.core5.io.Closer;
 
 /**
- * Lazy init InputStream wrapper.
+ * Lazy initializes from an {@link InputStream} wrapper.
  */
-class LazyDecompressingInputStream extends InputStream {
+class LazyDecompressingInputStream extends FilterInputStream {
 
-    private final InputStream wrappedStream;
     private final InputStreamFactory inputStreamFactory;
 
     private InputStream wrapperStream;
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
```diff
@@ -83,19 +83,29 @@ public boolean markSupported() {
...
-            }
+            Closer.close(wrapperStream);
         } finally {
-            wrappedStream.close();
+            super.close();
+        }
+    }
+
+    @Override
+    public synchronized void mark(final int readlimit) {
```
arturobernalg, line 99: Shouldn't we use ReentrantLock instead?

httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
```diff
@@ -83,19 +83,29 @@ public boolean markSupported() {
...
+        try {
+            initWrapper().mark(readlimit);
+        } catch (final IOException e) {
+            throw new UncheckedIOException(e);
         }
     }
 
+    @Override
+    public synchronized void reset() throws IOException {
```
arturobernalg, line 108: same here

arturobernalg (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
ok2c: @garydgregory No more `synchronized` for us going forward. Thread synchronization must be implemented using locks in order to be compatible with Java 21 fibers

ok2c (changes requested):
@garydgregory I am not sure synchronization is even needed here. This class is not supposed to be used for concurrent access.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
garydgregory: Hm, I guess the whole class isn't thread-safe to start with due to `initWrapper()`. I'll rework the PR.

These `synchronized` come from the Java 8, 11, and 17 contracts.

garydgregory: Thank you both for the reviews.

The simplest change and implementation that keeps in line with what to have today is to remove the `synchronized` keyword and let the delegate's implementation do what it will. If we wanted mark/reset to be thread-safe, we could make the whole class thread-safe like this:

```java
   private final ReentrantLock lock = new ReentrantLock();
   ...
    /**
     * Thread-safe initialization of the wrapped stream.
     *
     * @return the wrapper stream.
     * @throws IOException if an IO error occurs.
     */
    private InputStream initWrapper() throws IOException {
        if (wrapperStream == null) {
            lock.lock();
            try {
                wrapperStream = wrapperStream != null ? wrapperStream : inputStreamFactory.create(in);
            } finally {
                lock.unlock();
            }
        }
        return wrapperStream;
    }
```
but then each call to the stream would lock-unlock.

So, either:
- Don't use `synchronized` and that's it.
- Only have mark/reset use lock-unlock, or
- Implement using the above thread-safe `initWrapper()`?

For now, the PR takes the simplest route.
WDYT?

ok2c: > * Don't use `synchronized` and that's it.

I would go for this one.

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/entity/LazyDecompressingInputStream.java
garydgregory: OK, I merged with this option.

PR #571, opened 2024-08-23
Include Mozilla public suffix list file directly

massdosage:
As discussed on the dev mailing list, this should fix the failing `TestPublicSuffixMatcher` tests on Windows.

### DIFF
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
@@ -27,11 +27,8 @@
 
 package org.apache.hc.client5.http.psl;
 
-import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


garydgregory:
Hello @massdosage 

FWIW, `Paths.get("mozilla", "public-suffix-list.txt").toString()` would be better here but not really needed unless `Path`s where used in the test. The `"/"` does not matter in Java on or off Windows, Java normalizes the path for the platform. 

The test set up is also problematic because it leaks a file handle. We should be using a try-with-resource here:
```java
        // Create a matcher using a custom crafted public suffix list file
        try (InputStream in = classLoader.getResourceAsStream(SOURCE_FILE)) {
            Assertions.assertNotNull(in);
            final List<PublicSuffixList> lists = PublicSuffixListParser.INSTANCE.parseByType(new InputStreamReader(in, StandardCharsets.UTF_8));
            matcher = new PublicSuffixMatcher(lists);
        }
```
I fixed the above resource leak in git master.

The issue remains with the odd test failure on my end though :-(

massdosage:
OK, the `SOURCE_FILE` code was already there before any of my changes so I purposefully didn't touch it but the try-with-resources you suggest looks good. Even if this PR doesn't fix the issue at least it will give a better error messages if the mozilla file wasn't downloaded.

I'm happy to make any changes you suggest here though.

httpclient5/src/test/java/org/apache/hc/client5/http/psl/TestPublicSuffixMatcher.java
```diff
@@ -40,8 +42,7 @@
 class TestPublicSuffixMatcher {
 
     private static final String SOURCE_FILE = "suffixlistmatcher.txt";
-    private static final String PUBLIC_SUFFIX_LIST_FILE = "mozilla/public-suffix-list.txt";
```
ok2c, line 43: @massdosage That notation should work on all platforms. I am quite sure of it. This is not the right fix.

massdosage: I've made that change here now.

garydgregory:
Ah, this PR looks like a partial duplicate of https://github.com/apache/httpcomponents-client/pull/574 except that #574 preserves the ability to refresh the DAT file.

massdosage:
Feel free to close this one and use #574 instead.

PR #574, opened 2024-08-23
Includes a copy of publicsuffix.org's effective_tld_names.dat as a test fixture

garydgregory:
Includes a copy of in https://publicsuffix.org/list/effective_tld_names.dat in `httpclient5/src/test/resources/org/publicsuffix/list/effective_tld_names.dat`

Or, should this be a runtime resource?

### DIFF
--- a/httpclient5/pom.xml
+++ b/httpclient5/pom.xml
@@ -108,6 +108,12 @@
             <include>**/*.properties</include>
         </includes>
       </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+        <includes>
+            <include>**/*.dat</include>
+        </includes>
+      </resource>
     </resources>
     <plugins>
       <plugin>
@@ -123,8 +129,7 @@
             </goals>
             <configuration>
               <url>https://publicsuffix.org/list/effective_tld_names.dat</url>
-              <outputDirectory>${project.build.outputDirectory}/mozilla</outputDirectory>
-              <outputFileName>public-suffix-list.txt</outputFileName>
+              <outputDirectory>${project.basedir}/src/main/resources/org/publicsuffix/list</outputDirectory>
             </configuration>
           </execution>
         </executions>
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -4,3 +4,8 @@ Copyright 1999-2023 The Apache Software Foundation
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
+This product includes a copy of in https://publicsuffix.org/list/effective_tld_names.dat
+in httpclient5/src/test/resources/org/publicsuffix/list/effective_tld_names.dat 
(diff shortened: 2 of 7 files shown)


httpclient5/pom.xml
```diff
@@ -123,7 +123,7 @@
             </goals>
             <configuration>
               <url>https://publicsuffix.org/list/effective_tld_names.dat</url>
-              <outputDirectory>${project.build.outputDirectory}/mozilla</outputDirectory>
```
ok2c, line 126: @garydgregory We still need to ship a copy of this file with HttpClient artifact. Please roll back this change!

ok2c (changes requested):

ok2c (approved):

httpclient5/pom.xml
```diff
@@ -105,6 +105,7 @@
         <directory>src/main/resources</directory>
         <filtering>true</filtering>
         <includes>
+            <include>**/*.dat</include>
```
ok2c, line 108: @garydgregory Do we really need to filter `.dat` files? Seems unnecessary to me,

garydgregory: Hello @ok2c 
Yes, because other resources are already filtered and we want to include the new `.dat` file.

Alternatives are:
- Remove all `<include>` filters, but add an `<exclude>` filter for the `readme.txt` that explains the `.dat` file.
- Remove the whole `<resources>` section from that `pom.xml`, which means the including `readme.txt` that explains the `.dat` file.
- Remove all `<include>` filters, remove the `readme.txt`, which let's you then remove the whole `<resources>` section from that `pom.xml`
- Keep the PR as is.

Do you have a preference?

ok2c: > Yes, because other resources are already filtered and we want to include the new `.dat` file.

@garydgregory I may be mistaken, but resource filtering is used to replace variable placeholders with system or project variables, not to filter (include / exclude) resources. [1]. We already have `mozilla/public-suffix-list.txt` resource that gets added to the artifact jar, even though `*.txt` files are not filters.  Am I missing something?

We do not need to filter `.dat` files, do we? 

[1] https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html

garydgregory: Hello @ok2c 
Well, we're both right. I've pushed a new version that cleanly separates the concerns of which file to copy and whether to filter it or not by using two `<resource>` elements:

- We want the `.properties` file copied _and_ filtered. The git master version causes _only_ the properties file to be copied, the `.dat` file is not copied.
- We want to the `.dat` file copied but _not_ filtered.

garydgregory:
> @garydgregory All looks good. Please go ahead and merge.

@ok2c 
👍 done!

KeekLove:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2025-27820 
Will it affect httpclient4?

PR #576, opened 2024-09-04
fix: Solved the problem of garbled request headers

q1979635364:
If Form-data is used to pass Chinese strings, garbled characters exist


ok2c:
@q1979635364 This does not look right. MIME fields can have ASCII characters only. If one needs to put values with non-ASCII characters in MIME fields one must encode (BASE64 or quote-printable) one must encode those values.

PR #589, opened 2024-10-15
Testcontainers

ok2c:
Apache HTTPD / Squid compatibility tests have been migrated to TestContainers. 

This also ensures that more complex authentication scenarios that are difficult to create with our own integration test framework (various scenarios involving a proxy) will now be run along with the rest of integration tests on a regular basis.

This will especially be important for https://github.com/apache/httpcomponents-client/pull/577 and this would also provide an opportunity to create integration test cases for Kerberos auth in  the future.


httpclient5-testing/src/test/java/org/apache/hc/client5/testing/compatibility/ContainerImages.java
```diff
@@ -0,0 +1,119 @@
...
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+import org.testcontainers.images.builder.Transferable;
+
+public final class ContainerImages {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ContainerImages.class);
+
+    public static String WEB_SERVER = "test-httpd";
```
massdosage, line 45: Should these be `final` since they're UPPER_CASE constants?

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/compatibility/async/HttpAsyncClientCompatibilityTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    void assertProtocolVersion(final HttpClientContext context) {
+        switch (versionPolicy) {
+            case FORCE_HTTP_1:
+                Assertions.assertEquals(HttpVersion.HTTP_1_1, context.getProtocolVersion());
+                break;
+            case FORCE_HTTP_2:
+            case NEGOTIATE:
+                Assertions.assertEquals(HttpVersion.HTTP_2, context.getProtocolVersion());
+                break;
```
massdosage, line 98: Do we want a `default` case that throws an exception in case something unexpected happens?

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/compatibility/sync/CachingHttpClientCompatibilityTest.java
```diff
@@ -0,0 +1,156 @@
...
+                final ResponseCacheControl responseCacheControl = context2.getResponseCacheControl();
+                Assertions.assertNotNull(responseCacheControl);
+                Assertions.assertEquals(600, responseCacheControl.getMaxAge());
+                final HttpCacheEntry cacheEntry = context2.getCacheEntry();
+                Assertions.assertNotNull(cacheEntry);
+                Assertions.assertSame(cacheEntry, context1.getCacheEntry());
+                EntityUtils.consume(response.getEntity());
+            }
+
+            Thread.sleep(2000);
```
massdosage, line 110: What's this waiting for? Is there something we could check using something like `awaitility` instead of hardcoding a `sleep` which can lead to unpredictable test behaviour.

ok2c: > What's this waiting for? Is there something we could check using something like `awaitility` instead of hardcoding a `sleep` which can lead to unpredictable test behaviour.

@massdosage It is waiting for a cache entry to become stale and trigger its re-validation. I am not sure this can be done without a sleep given this is an integration test. I can however try to reduce it. 

I must admit I do not know what `awaitility` is and whether it can be applied here.

ok2c: @massdosage Have mercy on me.I just discovered `TestContainers` no so long ago and one throws  `awaitility` at me on top of it. 

Seriously our test cases are full of `CountDownLatch` based synchronization code and other ugliness. I would prefer to take a look at  `awaitility` as a way of improving our async testing code overall and not just to replace an occasional `Thread#sleep`

massdosage: Agreed, sounds better to take the approach where you fix more things at once than just this one :) So just leave this "as is" for now. If you want to create an issue summarising where you think something like `awaitility` could improve the tests I might be able to take a look at it.

I'm a big fan of TestContainers, have used it successfully on a number of projects to bring the "Docker Tests" alongside the rest of the Java tests so I'm happy to see it being used here. One just has to be careful to limit the number of these tests as they are slowly and more resource hungry.

arturobernalg (approved):
@ok2c The changes look reasonable to me, even though I don't have prior experience with `TestContainers`.

ok2c:
> @ok2c The changes look reasonable to me, even though I don't have prior experience with `TestContainers`.

@arturobernalg Honestly, nether do it. It is a learning process for me as well. But it is good to have an extra pair of eye balls looking at a big change to the project.

PR #594, opened 2024-10-27
Add  `rspauth` in Digest authentication.

arturobernalg:
Updated the Digest authentication implementation to ensure compliance with RFC requirements for mutual authentication. Specifically, added support for rspauth.


ok2c (approved):

ok2c:
@arturobernalg Is this a bug in our current implementation or an "improvement"?

arturobernalg:
> @arturobernalg Is this a bug in our current implementation or an "improvement"?



> @arturobernalg Is this a bug in our current implementation or an "improvement"?

@ok2c  IMO More of an improvement than a bug. The current implementation works as expected but lacks nextnonce support

ok2c:
> @ok2c IMO More of an improvement than a bug. The current implementation works as expected but lacks nextnonce support

@arturobernalg Got it.

PR #598, opened 2024-11-14
HTTPCLIENT-2350 - Enhance HttpClientConnectionOperator with Optional DNS Resolution Control

arturobernalg:
This PR introduces an enhancement to `DefaultHttpClientConnectionOperator`, adding the ability to disable DNS resolution for specific connections. This is particularly useful when connecting to .onion addresses through a SOCKS proxy, where public DNS lookups are not possible or desirable.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/DnsResolver.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/DnsResolver.java
@@ -27,7 +27,12 @@
 package org.apache.hc.client5.http;
 
 import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import org.apache.hc.core5.annotation.Contract;
 import org.apache.hc.core5.annotation.ThreadingBehavior;
@@ -61,4 +66,20 @@ public interface DnsResolver {
      */
     String resolveCanonicalHostname(String host) throws UnknownHostException;
 
+    /**
+     * Returns a list of {@link InetSocketAddress} for the given host with the given port.
+     *
+     * @see InetSocketAddress
+     *
+     * @since 5.5
+     */
+    default List<InetSocketAddress> resolve(String host, int port) throws UnknownHostException {
+        final InetAddress[] inetAddresses = resolve(host);
+        if (inetAddresses == null) {
+            return Collections.singletonList(InetSocketAddress.createUnresolved(host, port));
+        }
+        return Arrays.stream(inetAddresses)
+                .map(e -> new InetSocketAddress(e, port))
+                .collect(Collectors.toList());
+    }
 }
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/MultihomeIOSessionRequester.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/MultihomeIOSessionRequester.java
@@ -32,7 +32,7 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.net.UnknownHostException;
-import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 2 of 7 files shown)


ok2c:
@arturobernalg Damn. I also started working on this one. Conceptually I am not in favor of adding a new config nob every time someone asks for it. We did that often in the days of HC 3 and did us little good. I would prefer this issue solved with a custom `Function<HttpHost, InetSocketAddress[]>` address resolver or by making `DnsResolver` able to return unresolved addresses.

arturobernalg:
> @arturobernalg Damn. I also started working on this one. Conceptually I am not in favor of adding a new config nob every time someone asks for it. We did that often in the days of HC 3 and did us little good. I would prefer this issue solved with a custom `Function<HttpHost, InetSocketAddress[]>` address resolver or by making `DnsResolver` able to return unresolved addresses.

Fair enough, @ok2c . I agree with your point about avoiding adding new configuration knobs unnecessarily. I've updated the implementation to use a Function. Please take another pass

ok2c:
@arturobernalg My idea was to add this method to `DnsResolver`

```
    /**
     * Returns a list of {@link SocketAddress} for the given host with the given port.
     *
     * @see SocketAddress
     *
     * @since 5.5
     */
    default List<SocketAddress> resolve(String host, int port) throws UnknownHostException {
        InetAddress[] inetAddresses = resolve(host);
        if (inetAddresses == null) {
            return Collections.singletonList(InetSocketAddress.createUnresolved(host, port));
        }
        return Arrays.stream(inetAddresses)
                .map(e -> new InetSocketAddress(e, port))
                .collect(Collectors.toList());
    }
```

arturobernalg:
> @arturobernalg My idea was to add this method to `DnsResolver`
> 
> ```
>     /**
>      * Returns a list of {@link SocketAddress} for the given host with the given port.
>      *
>      * @see SocketAddress
>      *
>      * @since 5.5
>      */
>     default List<SocketAddress> resolve(String host, int port) throws UnknownHostException {
>         InetAddress[] inetAddresses = resolve(host);
>         if (inetAddresses == null) {
>             return Collections.singletonList(InetSocketAddress.createUnresolved(host, port));
>         }
>         return Arrays.stream(inetAddresses)
>                 .map(e -> new InetSocketAddress(e, port))
>                 .collect(Collectors.toList());
>     }
> ```

What about now, @ok2c ? I believe I've made the correct changes this time. Please review.

httpclient5/src/test/java/org/apache/hc/client5/http/impl/io/TestPoolingHttpClientConnectionManager.java
```diff
@@ -307,10 +308,10 @@ void testProxyConnectAndUpgrade() throws Exception {
         Mockito.when(conn.isOpen()).thenReturn(false);
         Mockito.when(future.get(1, TimeUnit.SECONDS)).thenReturn(entry);
         Mockito.when(pool.lease(
-                Mockito.eq(route),
-                Mockito.eq(null),
-                Mockito.any(),
-                Mockito.eq(null)))
+                        Mockito.eq(route),
```
ok2c, line 311: @arturobernalg Can we avoid formatting changes in the same commit with functional ones? If you want to change formatting please do it through the style check or at least in a separate commit.

ok2c (changes requested):
@arturobernalg Looks good to me, but similar change needs to be doe to `MultihomeIOSessionRequester` as well.

httpclient5/src/main/java/org/apache/hc/client5/http/DnsResolver.java
```diff
@@ -61,4 +67,21 @@ public interface DnsResolver {
...
     String resolveCanonicalHostname(String host) throws UnknownHostException;
 
+    /**
+     * Returns a list of {@link SocketAddress} for the given host with the given port.
+     *
+     * @see SocketAddress
+     *
+     * @since 5.5
+     */
+    default List<SocketAddress> resolve(String host, int port) throws UnknownHostException {
```
ok2c, line 77: @arturobernalg Almost there. Let's change this method to return `List<InetSocketAddress>` in order to avoid ugly downcasts in several places.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/DnsResolver.java
arturobernalg: done

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/MultihomeIOSessionRequester.java
```diff
@@ -129,7 +129,7 @@ public void cancelled() {
 
             void executeNext() {
                 final int index = attempt.getAndIncrement();
-                final InetSocketAddress remoteAddress = new InetSocketAddress(remoteAddresses[index], remoteEndpoint.getPort());
+                final InetSocketAddress remoteAddress = (InetSocketAddress) remoteAddresses.get(index);
```
ok2c, line 132: @arturobernalg I suppose this cast is no longer needed

arturobernalg: @ok2c I don't know what happened. I  just removed.

PR #607, opened 2025-01-04
HTTPCLIENT-2353: Fix IDN hostname mismatch by normalizing host and identity with IDN

arturobernalg:
Normalize Unicode and punycode strings before comparing host and certificate identities. This approach ensures domains like поиск-слов.рф properly match their punycode equivalents.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
@@ -27,6 +27,7 @@
 
 package org.apache.hc.client5.http.ssl;
 
+import java.net.IDN;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.cert.Certificate;
@@ -228,8 +229,18 @@ public final class DefaultHostnameVerifier implements HttpClientHostnameVerifier
                                          final PublicSuffixMatcher publicSuffixMatcher,
                                          final DomainType domainType,
                                          final boolean strict) {
+
+        final String normalizedIdentity;
+        try {
+            // Convert only the identity to its Unicode form
+            normalizedIdentity = IDN.toUnicode(identity);
+        } catch (final IllegalArgumentException e) {
+            return false;
+        }
+
+        // Public suffix check on the Unicode identity
         if (publicSuffixMatcher != null && host.contains(".")) {
-            if (publicSuffixMatcher.getDomainRoot(identity, domainType) == null) {
+            if (publicSuffixMatcher.getDomainRoot(normalizedIdentity, domainType) == null) {
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
```diff
@@ -228,8 +229,18 @@ private static boolean matchIdentity(final String host, final String identity,
                                          final PublicSuffixMatcher publicSuffixMatcher,
                                          final DomainType domainType,
                                          final boolean strict) {
+
+        final String punycodeHost;
```
ok2c, line 233: @arturobernalg Could we do it the other way around and convert identity from punycode to Unicode instead?

arturobernalg: @ok2c you mean Instead of converting both strings to punycode, bring both into the same Unicode form—i.e., use IDN.toUnicode(...) rather than toASCII(...) ???

ok2c: > @ok2c you mean Instead of converting both strings to punycode, bring both into the same Unicode form—i.e., use IDN.toUnicode(...) rather than toASCII(...) ???

@arturobernalg That is exactly what I mean. This way one would not need to normalize the hostname as it is has already been normalized to Unicode at  the construction time.

arturobernalg: @ok2c  changed. please take another look

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
```diff
@@ -228,36 +229,44 @@ private static boolean matchIdentity(final String host, final String identity,
                                          final PublicSuffixMatcher publicSuffixMatcher,
                                          final DomainType domainType,
                                          final boolean strict) {
+
+        final String unicodeIdentity;
```
ok2c, line 233: `unicodeIdentity` -> `normalizedIdentity` would be a better name?

ok2c (approved):

ok2c:
@arturobernalg Please note we still have a problem with punycode encoded hostname resolution. We might address it in a separate PR though.

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
arturobernalg: Done

arturobernalg:
> @arturobernalg Please note we still have a problem with punycode encoded hostname resolution. We might address it in a separate PR though.

@ok2c  I have fixed the test within this PR.

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
```diff
@@ -228,8 +229,19 @@ private static boolean matchIdentity(final String host, final String identity,
                                          final PublicSuffixMatcher publicSuffixMatcher,
                                          final DomainType domainType,
                                          final boolean strict) {
-        if (publicSuffixMatcher != null && host.contains(".")) {
-            if (publicSuffixMatcher.getDomainRoot(identity, domainType) == null) {
+
+        final String normalizedHost;
+        final String normalizedIdentity;
+        try {
+            normalizedHost = IDN.toUnicode(host);
```
ok2c, line 236: @arturobernalg This is unnecessary. The hostname has already been normalized. Please see `Host` and `HttpHost` classes

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java
arturobernalg: @ok2c 
you right. changed. Please do another pass

ok2c (approved):
@arturobernalg Looks good to me

arturobernalg:
> @arturobernalg Looks good to me

@ok2c  on it

arturobernalg:
> > @arturobernalg Looks good to me
> 
> @ok2c on it

Cherry-picked to `5.4.x`

PR #608, opened 2025-01-05
Normalize IDN hostnames before DNS resolution to prevent UnknownHostException

arturobernalg:
This PR addresses an issue where Unicode hostnames cause UnknownHostException during DNS resolution. The change normalizes hostnames to punycode using IDN.toASCII() before passing them to the DNS resolver.


ok2c:
@arturobernalg Just a question. Why not having this normalization logic in `SystemDefaultDnsResolver`? It already has some extra logic for resolution of IPv6 names in it.

arturobernalg:
> @arturobernalg Just a question. Why not having this normalization logic in `SystemDefaultDnsResolver`? It already has some extra logic for resolution of IPv6 names in it.

You right. And we avoid duplicated code.

httpclient5/src/main/java/org/apache/hc/client5/http/SystemDefaultDnsResolver.java
```diff
@@ -44,8 +45,9 @@ public class SystemDefaultDnsResolver implements DnsResolver {
     @Override
     public InetAddress[] resolve(final String host) throws UnknownHostException {
         try {
+            final String normalizedHost = IDN.toASCII(host);
```
ok2c, line 48: @arturobernalg I would do  `IDN.toASCII` only if the hostname fails `TextUtils#isAllASCII` for the sake of optimization.

ok2c: @arturobernalg Catching `IllegalArgumentException` and recovering from it (for instance, by keeping the original hostmame) would be nice here

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/SystemDefaultDnsResolver.java
arturobernalg: @ok2c  Please do another pass

ok2c: @arturobernalg Do you have any objections to doing the test `TextUtils#isAllASCII` prior to doing `IDN.toASCII`?

arturobernalg: @ok2c done

ok2c (approved):
@arturobernalg Looks good to me

arturobernalg:
> @arturobernalg Please cherry-pick to `5.4.x`

Done

PR #611, opened 2025-01-18
HTTPCLIENT-2180. Implement warm-up functionality for PoolingAsyncClientConnectionManager

arturobernalg:
This PR introduces the `warmUp` method for `PoolingAsyncClientConnectionManager` and `PoolingHttpClientConnectionManager,` allowing `pre-initialization` of connections to a target host. This functionality ensures connections are ready for immediate use, addressing performance concerns in latency-sensitive applications.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -820,5 +820,60 @@ public class PoolingHttpClientConnectionManager
         return this.closed.get();
     }
 
+    /**
+     * Performs a warm-up of connections to the specified target host synchronously.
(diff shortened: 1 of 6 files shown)


ok2c:
I am sorry @arturobernalg but this looks wrong. The proposed code does not really "warm up" the pool. It merely creates a bunch of pool entries with no connection in them. It really makes no sense. 

The only reliable way of creating a properly established connection is by executing a cheap request such as `OPTIONS` along the given route. Please note one needs a fully initialized and a functional client for that as a route may involve multiple hops, proxy tunnels and TLS security. It cannot be done by the connection manager alone.

If you really intend to resolve HTTPCLIENT-2180 you should probably create a utility class that executes a bunch of `OPTIONS` requests along the given route, but I would rather close HTTPCLIENT-2180 as "wontfix"

arturobernalg:
> wontfix

@ok2c  agree to close the ticket as wontfix

PR #615, opened 2025-01-23
HTTPCLIENT-2358 Implement a mutual authentication capable SPNEGO scheme

stoty:


ok2c:
@stoty I cannot contribute here much. My only wish, and I understand it is a big ask, to also create a compatibility test similar to those we have for BASIC / DIGEST with Squid and Apache HTTPD

https://github.com/apache/httpcomponents-client/tree/master/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/compatibility

stoty:
I was afraid you're going to say that...

There are two ways to go about that: 
- Try to find an docker image with a KDC and web server, and model the test on the Squid test
- Initialize a local (non-dockerized) KDC with Kerby, start Jetty, and use that.

I don't know which of the two is more work, but I know where to copy the test setup from for the second one.

Do you have a preference ?

stoty:
It is a reasonable request, and it IS helpful to catch any regressions, etc.
I will try to find and existing docker image that is close enough.

garydgregory:
Apache Kerby implements Kerberos, you can get inspiration (copy) how they do it: https://directory.apache.org/kerby/

stoty:
Kerby is great for in-JVM tests, but for dockerized tests it's probably easier to use MIT kerberos.

stoty:
After poking around a bit, it seems that neither Apache Httpd, not Nginx supports SPNEGO out of the box.
While both have third party SPNEGO modules, creating a test image with a full working setup looks like a lot of work.

For now, I plan to make a local test with Kerby + Jetty, without Docker.

michael-o:
A full coverage test contains JGSS (via Tomcat), MIT Kerberos  (via mod_auth_gssapi), Microsoft Kerberos  (via IIS). Virtually impossible to automate. I will do manual testing anyway.

michael-o (changes requested):
This is much less code review. Thanks. I already see some points.

stoty:
Thanks.
I've made the changes to MutualKerberosConfig.

While I don't use Tomcat, testing Jgss with Jetty is doable, in fact one of the ways I tested this was via the SPNEGO test in HttpCLient->Calcite->Phoenix PQS.
It does bring in a ton of extra test dependencies, and I couldn't really find where to fit the that integration test in the project.

stoty:
I have added tests for directo connection to httpd + mod_auth_spnego, @michael-o .
The tests have immediately found a problem, which I also fixed.

The async client cannot use the Subject set with Subject.callAs to figure out the kerberos credentials (probably due to the auth being run from a different thread)
Is that acceptable, or should we extract and store the Subject ?

stoty:
I have added tests for SPNEGO authentication for Squid.

I feel that this is ready now, @michael-o .

michael-o:
> I have added tests for direct connection to httpd + mod_auth_spnego, @michael-o . The tests have immediately found a problem, which I also fixed.
> 
> The async client cannot use the Subject set with Subject.callAs to figure out the Kerberos credentials (probably due to the auth being run from a different thread) Is that acceptable, or should we extract and store the Subject ?
> 
> Async still works if we create and set KerbersCredentials. (as in the new tests I added)

Which bug exactly?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/MutualGssSchemeBase.java
```diff
@@ -0,0 +1,365 @@
...
+            final HttpContext context) throws AuthenticationException {
+
+        Args.notNull(host, "Auth host");
+        Args.notNull(credentialsProvider, "CredentialsProvider");
+
+        setGssCredential(credentialsProvider, host, context);
+        return true;
+    }
+
+    protected void setGssCredential(final CredentialsProvider credentialsProvider,
```
stoty, line 307: This is the fix for the bug I found, @michael-o .
When we specifx an explicit kerberos crential (as opposed to relying on the jaas default)
Then it was only set from isResponseReady() which is too late, as we need it earlier, in processChallenge().
I had to add code to processChallange to process the credential, so that we can use it there.

michael-o (changes requested):
I am picking this up again and want to complete next month.
Structural issues:
* Since in both `auth` and `impl.path` packages contains multiple new classes there should be in respective new subpackages `gss` for tidiness.
* Don't use the term "Kerberos" in classnames becausee we perform SPNEGO with GSS-API, correctly it should be called "Gss<suffix>"
* SPNEGO in class names should be written consistently, In PascalCase and not mixed case.
* Drop the prefix "Mutual" in class names altogether because one can disable on request making it a contradiction and the other party could pontentially deny it.

stoty:
I can do that, but that would mean the the three formerly mutual classes only differ in case and package from the old deprecated classes. I can see that causing a lot of gried for users down the line.

> * Since in both `auth` and `impl.path` packages contains multiple new classes there should be in respective new subpackages `gss` for tidiness.

Sure. 

> * Don't use the term "Kerberos" in classnames becausee we perform SPNEGO with GSS-API, correctly it should be called "Gss"

Sure.

> * SPNEGO in class names should be written consistently, In PascalCase and not mixed case.

Only the old deprecated classes use SPNego, and fixing those would make it even easier to mix them up with the new ones. (apart from breaking backwards compatibility, if we care about that)

> * Drop the prefix "Mutual" in class names altogether because one can disable on request making it a contradiction and the other party could pontentially deny it.

I can do that, but then then the only difference between the old and new classes will be the packages and the case.
Machines don't care, but I see that causing grief for users.

stoty:
Thanks for the review.
I have made the suggested changes, but I am not happy with how close the old new class names are.

I have also left KerberosCredentials alone, because it is used by both the new and old code.
Should I make an identical child class with GssCredentials name ?

michael-o:
> I can do that, but that would mean the the three formerly mutual classes only differ in case and package from the old deprecated classes. I can see that causing a lot of gried for users down the line.
> 
> > * Since in both `auth` and `impl.path` packages contains multiple new classes there should be in respective new subpackages `gss` for tidiness.
> 
> Sure.
> 
> > * Don't use the term "Kerberos" in classnames becausee we perform SPNEGO with GSS-API, correctly it should be called "Gss"
> 
> Sure.
> 
> > * SPNEGO in class names should be written consistently, In PascalCase and not mixed case.
> 
> Only the old deprecated classes use SPNego, and fixing those would make it even easier to mix them up with the new ones. (apart from breaking backwards compatibility, if we care about that)
> 
> > * Drop the prefix "Mutual" in class names altogether because one can disable on request making it a contradiction and the other party could pontentially deny it.
> 
> I can do that, but then then the only difference between the old and new classes will be the packages and the case. Machines don't care, but I see that causing grief for users.

Don't rename old ones, only new ones. If old ones are reused maye it would be better to copy them?

michael-o:
> Thanks for the review. I have made the suggested changes, but I am not happy with how close the old new class names are.
> 
> I have also left KerberosCredentials alone, because it is used by both the new and old code. Should I make an identical child class with GssCredentials name ?

Yes, please. Don't break old deprecated,  but add new classes for the new impl.

stoty:
> > I can do that, but then then the only difference between the old and new classes will be the packages and the case. Machines don't care, but I see that causing grief for users.
> 
> Don't rename old ones, only new ones. If old ones are reused maye it would be better to copy them?

I wasn't clear. The classes are different, but their names are almost identical.

As the Scheme is not enabled by default, users have to add some boilerplate code to enable SPNEGO at all.
It is now too easy to use the old SPNegoScheme instead of the new SpnegoScheme in that boilerplate IMO.

i have renamed  KerberCredentials, and added a backwards compatibility subclass.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+import org.ietf.jgss.GSSCredential;
+import org.ietf.jgss.GSSException;
+import org.ietf.jgss.GSSManager;
+import org.ietf.jgss.GSSName;
+import org.ietf.jgss.Oid;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Common behaviour for the new mutual authentication capable {@code GSS} based authentication
```
michael-o, line 59: It would make it even simple, whether mutual auth happens depens on many factors, we cannot guarantee this;
Base class for GSS-based authentication schemes.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+        UNINITIATED,
+        TOKEN_READY,
+        TOKEN_SENT,
+        SUCCEEDED,
+        FAILED,
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(GssSchemeBase.class);
+    private static final String NO_TOKEN = "";
+    private static final String KERBEROS_SCHEME = "HTTP";
```
michael-o, line 81: This is a service name, see https://learn.microsoft.com/en-us/windows/win32/ad/name-formats-for-unique-spns or just service https://datatracker.ietf.org/doc/html/rfc2743#page-85 nothing service specific since GSS-API delegates to the mechanism to canonicalize according to mech rules.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+
+    // The GSS spec does not specify how long the conversation can be. This should be plenty.
+    // Realistically, we get one initial token, then one maybe one more for mutual authentication.
+    private static final int MAX_GSS_CHALLENGES = 3;
```
michael-o, line 85: Maybe a TODO in the future that this might need to be configurable with the upcoming IAKerb support in Windows and Java might get it as well.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+
+        final byte[] challengeToken = Base64.decodeBase64(authChallenge == null ? null : authChallenge.getValue());
+
+        final String gssHostname;
+        String hostname = host.getHostName();
+        if (config.isUseCanonicalHostname()) {
+            try {
+                 hostname = dnsResolver.resolveCanonicalHostname(host.getHostName());
+            } catch (final UnknownHostException ignore) {
+            }
```
michael-o, line 156: Is the swalling expected? or log at least?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+            final HttpClientContext clientContext = HttpClientContext.cast(context);
+            final String exchangeId = clientContext.getExchangeId();
+            LOG.debug("{} GSS init {}", exchangeId, gssHostname);
+        }
+        try {
+            setGssCredential(HttpClientContext.cast(context).getCredentialsProvider(), host, context);
+            queuedToken = generateToken(challengeToken, KERBEROS_SCHEME, gssHostname);
+            switch (state) {
+            case UNINITIATED:
+                if (challenge != NO_TOKEN) {
```
michael-o, line 174: String and `""`? Why not work with `null`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+                    }
+                    // TODO Should we fail ? That would break existing tests that send a token
+                    // in the first response, which is against the RFC.
+                }
+                state = State.TOKEN_READY;
+                break;
+            case TOKEN_SENT:
+                if (challenged) {
+                    state = State.TOKEN_READY;
+                } else if (mutualAuth) {
```
michael-o, line 188: This is wrong. The only loop condition is `gssContext.isEstablished()`. The acceptor can send still a token, even if not mutual auth is peformed and even in an error condition an acceptor context can produce a token (e.g. MIT Kerberos), if necessary I can show code. HAve seen this years ago.
Note that requesting mutual auth is only a hint to GSS-API, wether the security context will provide is upto the mech.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,366 @@
...
+    enum State {
+        UNINITIATED,
+        TOKEN_READY,
+        TOKEN_SENT,
+        SUCCEEDED,
+        FAILED,
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(GssSchemeBase.class);
+    private static final String NO_TOKEN = "";
```
michael-o, line 80: why not null?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return StandardAuthScheme.SPNEGO;
+    }
+
+    @Override
+    protected byte[] generateToken(final byte[] input, final String gssServiceName, final String gssHostname) throws GSSException {
+        return generateGSSToken(input, new Oid(SPNEGO_OID), gssServiceName, gssHostname);
```
michael-o, line 105: Why not initialize this is a `static{}` block once instead doing over and over again like in https://github.com/michael-o/tomcatspnegoad/blob/e2d95c6708b7ae726acf882ba680534fa5ec8dcc/tomcat90/src/main/java/net/sf/michaelo/tomcat/authenticator/GSSAuthenticatorBase.java#L43-L55

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public boolean isConnectionBased() {
+        return true;
```
michael-o, line 110: This is wrong and false. If there is only one roundtrip then you don't need to bind the GSS context to the connection otherwise you have to, see also https://github.com/gssapi/mod_auth_gssapi?tab=readme-ov-file#gssapiconnectionbound

michael-o (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: That's completely buggy, we never set that variable.
I'm going to fix that.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
stoty: Sure

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
stoty: Right, fixing.
Note that this was copied from the old one.
Should we fix it there as well ?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: How should we handle the case when we request mutual  auth, but don't get it ?
Should we just fail, or should we split the option , one to request it, and one to require that it succeeds ?

stoty:
I have fixed the small issues.
It's getting late, I'll try to do the loop condition tomorrow.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
michael-o: No, leave as is.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: Read https://datatracker.ietf.org/doc/html/rfc7546, there is even code depicting how you handle the loop. It should answer almost all questions.

stoty: The spec kind of leaves this to the implementation:

   If a flag
   was requested but is not available and that feature is necessary for
   the application protocol, the initiator must destroy the security
   context and not use the security context for application traffic.
   
If I interpret that losely, this means that we could split the mutualAuth, because the caller may ask for mutual Auth, but may not consider it necessary (though admittedly this not make a lot of sense).

   If the initiator is expecting a context token (that is, the previous
   call to GSS_Init_sec_context() returned GSS_S_CONTINUE_NEEDED) but
   does not receive such a token within a reasonable amount of time
   after transmitting the previous output_token to the acceptor, the
   initiator should assume that the acceptor's state is invalid and fail
   the GSS negotiation.  Again, it may be appropriate for the initiator
   to report this error condition to the acceptor via the application's
   communication channel.
 
This one is even more interesting, as if we implemented this strictly, we could never work with Squid (and I expect that that there are many more similarly broken servers.)

I lean towards not making mutual authentication optional, as the user has no way to check the status, and if it is optional, then why even bother in the first place ? It would catch misconfigured/malicious servers that send an incorrect response, but not ones that just plain ignore the request.

On the other hand, I lean towards adding a strict flag to the config, so that we require that the GSS negotiation completes, so that we have a standards-compliant mode. This of course is implied when mutual Auth is requested.

stoty:
I have added a new config option for non-standard servers, and rewritten the established checks.

Please take another look @michael-o .

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -183,35 +184,54 @@ public void processChallenge(
...
                 break;
             case TOKEN_SENT:
-                if (challenged) {
-                    state = State.TOKEN_READY;
-                } else if (mutualAuth) {
-                    // We should have received a valid mutualAuth token
-                    if (!gssContext.isEstablished()) {
+                if (challengeToken == null) {
+                    if (!challenged && ignoreMissingToken) {
+                        // Got a 200 without a challenge. Old non RFC compliant server.
```
michael-o, line 189: Why 200? The auth loop can be completed on any response status unless the intial token was incorrect.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: I will read this one as well, check other impls and share my thoughts on this.

michael-o: > we could never work with Squid (and I expect that that there are many more similarly broken servers.)

I have never worked with Squid. If Squid does not complete the auth loop one has to raise this as a bug, can you do that? Just because it is broken does not mean it should not be fixed.

> I lean towards not making mutual authentication optional, as the user has no way to check the status, and if it is optional, then why even bother in the first place ?

Correct, request it always, it does not cause any harm. libself and Cyrus SASL do the same.

> On the other hand, I lean towards adding a strict flag to the config, so that we require that the GSS negotiation completes, so that we have a standards-compliant mode. This of course is implied when mutual Auth is requested.

Non strict in terms of the RFC for broken servers if you cannot fix them.

== 

I agree on your citation of "If a flag ...'", no need to fail if mutual isn't available on the context after completion even if we set it always. Set and forget.

On the second citation: Yes, if the acceptor fails to resond with a token we must fail with an expcetion here. We can relax this with an explicit flag, but I would issue a warning then this happens. Clients must know that their acceptor is broken. Compare py-requests `verify` flag set to `False`. It will bug you that certificate verification is disabled, same shall we.

Do you want to check the code again for this or does it already comply from your PoV?

Note: I checked libserf and Cyrus SASL's GSSAPI and GS2 mechs. Libserf always requests mutual with fire and forget, GSSAPI mech has a FIXME to check return flags on the context, GS2 mech enforces all requested flags from the completed context:

* https://github.com/apache/serf/blob/d4dc7a996d1a2a34dc3aa504cf44c0658120d666/auth/auth_spnego_gss.c#L204
* https://github.com/apache/serf/blob/d4dc7a996d1a2a34dc3aa504cf44c0658120d666/auth/auth_spnego_gss.c#L210
* https://github.com/cyrusimap/cyrus-sasl/blob/b8302a8c99d7079d1b68acf18802fce25c8c8243/plugins/gs2.c#L827-L838
* https://github.com/cyrusimap/cyrus-sasl/blob/b8302a8c99d7079d1b68acf18802fce25c8c8243/plugins/gssapi.c#L2294

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
michael-o: Sorry, I misread. We should fix the copy, not the original one. My bad!

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: Thank you.

Yes, the current code should comply with that.
"ignoreMissingToken" is the non-strict flag (defaults to off), but I am open to a better name.

I am not sure about the warning, that can be a really high volume, and the user has already disabled it explicitly.
I am currently logging a DEBUG message for this.

Currently we enforce mutual auth if requested.
Requesting it unconditionally, and enforcing based on a flag also sounds good to me, I will make the change.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: Fixing

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: > Requesting it unconditionally, and enforcing based on a flag also sounds good to me, I will make the change.

I wouldn't enfore it. I Would make it fire and forget (pray). The application protocol does not require it. Cyrus SASL  does.

stoty: > > Requesting it unconditionally, and enforcing based on a flag also sounds good to me, I will make the change.
> 
> I wouldn't enfore it. I Would make it fire and forget (pray). The application protocol does not require it. Cyrus SASL does.

I disagree.

The application protocol is determined by the code that uses httpclient. If we don't give it the option to enforce mutual authentication, then we may as well un-deprecate the old schemes, as we haven't made things any more secure.

The original reason for deprecating the old scheme was that it didn't support mutual authentication, and thus it was insecure.

michael-o: The old scheme had also other issues beside the mutual auth.

So you are saying, if we request mutual auth we shall check for it after the context has been established and if not then we should fail? Including a switch to make it soft?

michael-o: > Yes, the current code should comply with that.
"ignoreMissingToken" is the non-strict flag (defaults to off), but I am open to a better name.

`ignoreIncomplete`SecurityContext`?

michael-o: > Yes, the current code should comply with that.
"ignoreMissingToken" is the non-strict flag (defaults to off), but I am open to a better name.

`ignoreIncompleteSecurityContext`?

stoty:
I have split the mutual auth config option to request and require.
Having both takes no effort on our part, but allows for set and pray mutual authentication (though I doubt that value of that).
 
I can drop the request one, and reqest unconditionally if you prefer.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: > I am not sure about the warning, that can be a really high volume, and the user has already disabled it explicitly.
> I am currently logging a DEBUG message for this.

Let go for WARNING and see whether we will receive complaints, then we still can reduce. I'd eager to see who is faulty, this will give everyone a chance to report to the producer of the crap.

stoty: > The old scheme had also other issues beside the mutual auth.
> 
> So you are saying, if we request mutual auth we shall check for it after the context has been established and if not then we should fail? Including a switch to make it soft?

Yes. currently these are requestMutualAuth and requireMutualAuth.

michael-o: > I am not sure about the warning, that can be a really high volume, and the user has already disabled it explicitly.
> I am currently logging a DEBUG message for this.

Lets go for WARNING and see whether we will receive complaints, then we still can reduce. I'd eager to see who is faulty, this will give everyone a chance to report to the producer of the crap.

stoty:
I opened https://bugs.squid-cache.org/show_bug.cgi?id=5485 for the Squid issues.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: FTR: I checked the source of Squid, man...
Look at https://github.com/squid-cache/squid/blob/0c90595b9d63930f07313a60eb3c7ec9859251bf/src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc#L750-L754, completion is never really checked...
I also failed to find where the output token goes, it is simply not processed after it has been converted to Base 64, moreover they use a lot of raw Kerberos 5 APIs which aren't required at all, one can do everything through GSS-API these days, even obtaining the PAC data. For me, Squid is broken.

michael-o:
> I opened https://bugs.squid-cache.org/show_bug.cgi?id=5485 for the Squid issues.

Thanks, very good. mod_auth_gssapi is the gold standard for me written by very knowledgeable people.

httpclient5/src/main/java/org/apache/hc/client5/http/auth/gss/GssConfig.java
```diff
@@ -0,0 +1,204 @@
...
+ *
+ */
+
+package org.apache.hc.client5.http.auth.gss;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+
+/**
+ * Immutable class encapsulating GSS configuration options for the new mutual auth capable
```
michael-o, line 34: `...for the new {@link SpnegoScheme}.`

httpclient5/src/main/java/org/apache/hc/client5/http/auth/gss/GssCredentials.java
```diff
@@ -0,0 +1,74 @@
...
+        this.gssCredential = gssCredential;
+    }
+
+    public GSSCredential getGSSCredential() {
+        return gssCredential;
+    }
+
+    @Override
+    public Principal getUserPrincipal() {
+        return null;
```
michael-o, line 66: Can you add a TODO than the principal can be obtained from the `GSSCredential`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: Still open

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: Still open

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: Still open...

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.hc.client5.http.auth.StandardAuthScheme;
+import org.apache.hc.client5.http.impl.auth.SPNegoScheme;
+import org.ietf.jgss.GSSException;
+import org.ietf.jgss.Oid;
+
+/**
+ * SPNEGO (Simple and Protected GSSAPI Negotiation Mechanism) authentication
+ * scheme.
+ * <p>
+ * This is the new mutual authentication capable Scheme which replaces the old deprecated non mutual
```
michael-o, line 40: I think is more than enough to mention that this superseded the other class, not necessary to mention its capabilities here.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/SpnegoScheme.java
michael-o: Thinking about this again, This is not a problem when the completion done after one roundtrip. So I guess when the security context says continue needed than this needs to be bound to the connection. WDYT?

michael-o (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+            final boolean challenged,
+            final AuthChallenge authChallenge,
+            final HttpContext context
+            ) throws AuthenticationException {
+
+        if (challengesLeft-- <= 0 ) {
+            if (LOG.isDebugEnabled()) {
+                final HttpClientContext clientContext = HttpClientContext.cast(context);
+                final String exchangeId = clientContext.getExchangeId();
+                LOG.debug("{} GSS error: too many challenges received. Infinite loop ?", exchangeId);
```
michael-o, line 142: This is actually a good question. I have the feeling that this should be at warning lever since an upper layer exception cannot reflect this.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+            }
+            // TODO: Should we throw an exception ? There is a test for this behaviour.
+            state = State.FAILED;
+            return;
+        }
+
+        final byte[] challengeToken = Base64.decodeBase64(authChallenge == null ? null : authChallenge.getValue());
```
michael-o, line 149: Why bother Base64 if we can do the ternary check outside?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+            final HttpClientContext clientContext = HttpClientContext.cast(context);
+            final String exchangeId = clientContext.getExchangeId();
+            LOG.debug("{} GSS init {}", exchangeId, gssHostname);
+        }
+        try {
+            switch (state) {
+            case UNINITIATED:
+                setGssCredential(HttpClientContext.cast(context).getCredentialsProvider(), host, context);
+                queuedToken = generateToken(challengeToken, KERBEROS_SCHEME, gssHostname);
+                if (challengeToken != null) {
```
michael-o, line 175: This check should be done before the `generateToken()` method, no?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+                    if (LOG.isDebugEnabled()) {
+                        final HttpClientContext clientContext = HttpClientContext.cast(context);
+                        final String exchangeId = clientContext.getExchangeId();
+                        LOG.debug("{} Internal GSS error: token received when none was sent yet: {}", exchangeId, challengeToken);
+                    }
+                    // TODO Should we fail ? That would break existing tests that send a token
+                    // in the first response, which is against the RFC.
```
michael-o, line 182: IMHO, this violates, as you write, the RFC and shall fail. GSS-API is a client first approach. See Cyrus SASL:
```
Plugin "gs2" [loaded],  API version: 4
        SASL mechanism: GS2-IAKERB, best SSF: 0
        security flags: NO_ANONYMOUS|NO_PLAINTEXT|NO_ACTIVE|PASS_CREDENTIALS|MUTUAL_AUTH
        features: WANT_CLIENT_FIRST|NEED_SERVER_FQDN|GSS_FRAMING|CHANNEL_BINDING
Plugin "gs2" [loaded],  API version: 4
        SASL mechanism: GS2-KRB5, best SSF: 0
        security flags: NO_ANONYMOUS|NO_PLAINTEXT|NO_ACTIVE|PASS_CREDENTIALS|MUTUAL_AUTH
        features: WANT_CLIENT_FIRST|NEED_SERVER_FQDN|GSS_FRAMING|CHANNEL_BINDING
Plugin "gssapiv2" [loaded],     API version: 4
        SASL mechanism: GSS-SPNEGO, best SSF: 256
        security flags: NO_ANONYMOUS|NO_PLAINTEXT|NO_ACTIVE|PASS_CREDENTIALS|MUTUAL_AUTH
        features: WANT_CLIENT_FIRST|PROXY_AUTHENTICATION|NEED_SERVER_FQDN|SUPPORTS_HTTP
Plugin "gssapiv2" [loaded],     API version: 4
        SASL mechanism: GSSAPI, best SSF: 256
        security flags: NO_ANONYMOUS|NO_PLAINTEXT|NO_ACTIVE|PASS_CREDENTIALS|MUTUAL_AUTH
        features: WANT_CLIENT_FIRST|PROXY_AUTHENTICATION|NEED_SERVER_FQDN
```

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+                state = State.TOKEN_READY;
+                break;
+            case TOKEN_SENT:
+                if (challengeToken == null) {
+                    if (!challenged && ignoreMissingToken) {
+                        // Got a Non 401/407 code without a challenge. Old non RFC compliant server.
+                        if (LOG.isDebugEnabled()) {
+                            final HttpClientContext clientContext = HttpClientContext.cast(context);
+                            final String exchangeId = clientContext.getExchangeId();
+                            LOG.debug("{} GSS Context is not established, but continuing because GssConfig.ignoreMissingToken is true.", exchangeId);
```
michael-o, line 193: Warning, not debug as agreed.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+                        }
+                        state = State.SUCCEEDED;
+                        break;
+                    } else {
+                        if (LOG.isDebugEnabled()) {
+                            final HttpClientContext clientContext = HttpClientContext.cast(context);
+                            final String exchangeId = clientContext.getExchangeId();
+                            LOG.debug("{} Did not receive required challenge and GssConfig.ignoreMissingToken is false.",
```
michael-o, line 201: I'd drop the part from "...and". If not set explicitly, no need to mention.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+                                exchangeId);
+                        }
+                        state = State.FAILED;
+                        throw new AuthenticationException(
+                                "Did not receive required challenge and GssConfig.ignoreMissingToken is false.");
```
michael-o, line 206: Same here

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+                } else if (!gssContext.isEstablished()) {
+                    if (LOG.isDebugEnabled()) {
+                        final HttpClientContext clientContext = HttpClientContext.cast(context);
+                        final String exchangeId = clientContext.getExchangeId();
+                        LOG.debug("{} GSSContext is not established.", exchangeId);
+                    }
+                    state = State.FAILED;
+                    // TODO should we have specific exception(s) for these ?
+                    throw new AuthenticationException(
+                            "GSSContext is not established.");
```
michael-o, line 221: You haven't disposed the security context.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
```diff
@@ -0,0 +1,394 @@
...
+                        throw new AuthenticationException(
+                                "requireMutualAuth is true but GSSContext mutualAuthState is false");
+                    } else {
+                        if (LOG.isDebugEnabled()) {
+                            final HttpClientContext clientContext = HttpClientContext.cast(context);
+                            final String exchangeId = clientContext.getExchangeId();
+                            LOG.debug("{} GSSContext MutualAuthState is false, but continuing because GssConfig.requireMutualAuth is false.",
+                                exchangeId);
+                        }
+                        state = State.FAILED;
```
michael-o, line 240: I do not fully understand this. If mutual auth isn't required, why fail if it is not available?

michael-o (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: I'm not sure what you mean. 
The ternary here prevents an NPE from authChallenge.getValue() is authChallange is null.
Base64.decodeBase64 returns null for null, so we don't have to check for a null value.

michael-o: `authChallenge == null ? null : Base64.decodeBase64(authChallenge.getValue());`

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: OK.
Will need to change test for it.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: This is a bug, I'm fixing it. (We don't have a test case for this path)

stoty:
I realize that I somehow didn't upload the changes for GSSSchemeBase in my last pach, that's why a lot of your earlier comments were not addressed.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: It makes sense in a way because you avoid the object creation in the method.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
michael-o: OK

httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/gss/GssSchemeBase.java
stoty: OK

httpclient5/src/main/java/org/apache/hc/client5/http/auth/gss/GssCredentials.java
stoty: OK

stoty:
I hope I got all the fixes we discussed, @michael-o .

Based on my digging, I have set isConnectionBased() to false.

I have also fixed the logic in the needAuthentication() to handle proxy + target authentication correctly.
(Should that be split to a separate JIRA ?)

stoty:
Ping @michael-o ?

httpclient5/src/main/java/org/apache/hc/client5/http/auth/gss/GssCredentials.java
stoty: Done

httpclient5/src/main/java/org/apache/hc/client5/http/auth/gss/GssConfig.java
stoty: Fixed.

stoty:
Hi, @michael-o , I hope you had a good summer,

Now that everyone is back at the grinder, I request you you to kindly help get this feature over the finish line.

I have rebased the patch to the latest master (there were no major issues, some dependencies were removed and I had to add an exclusion to animal-sniffer for the reflection classes in SubjectUtil).

I have also found two unresolved comments regarding code comments from you, I have resolved those.

ok2c:
@michael-o  Any chance this feature could make it into 5.6?

michael-o:
> @michael-o Any chance this feature could make it into 5.6?

Do you have a timeframe for 5.6?

ok2c:
> > @michael-o Any chance this feature could make it into 5.6?
> 
> Do you have a timeframe for 5.6?

@michael-o I would like to release 5.6-alpha1 next week. However, a GA is unlikely sooner than Q2, so there is still some time, given this feature has no major API changes and can be added quite late in the release cycle.

michael-o:
> > > @michael-o Any chance this feature could make it into 5.6?
> > 
> > 
> > Do you have a timeframe for 5.6?
> 
> @michael-o I would like to release 5.6-alpha1 next week. However, a GA is unlikely sooner than Q2, so there is still some time, given this feature has no major API changes and can be added quite late in the release cycle.

You mean 2026Q2?

ok2c:
> You mean 2026Q2?

@michael-o Yes.

stoty:
I realized that the SpnegoUtil stuff is not really related to the new SPNEGO implementation, it is needed for JDK18+.

Should I split that to a new ticket/PR ?

ok2c:
@stoty You may still be stuck with @michael-o as a reviewer, but generally smaller change-sets are always easier to review and digest.

stoty:
My bad, SubjectUtil is only used in the test code I added, so it doesn't make a lot of sense to add it in a separate ticket.

stoty:
There is a separate (pre-existing) issue with JDK22+ and Async SPNEGO.

I have opened https://issues.apache.org/jira/browse/HTTPCLIENT-2402 to track it,
and will open a follow-up PR once this one has been merged.
(As testing for the problem depends on the test infra added in this patch)

stoty:
I could factor out the ProtocolExec/AsyncProtocolExec and AuthenticationHandler changes into a separate JIRA, as those fix a bug in HTTPCLIENT-2356, and are not directly related to the new Scheme. Should I do that ?

stoty:
I have decided to create a separate JIRA for the proxy mutual auth issue mentioned above.
#745
[HTTPCLIENT-2403](https://github.com/apache/httpcomponents-client/pull/745)

Please take a look.
If that one is merged, I will rebase this PR.

michael-o:
Folks, I have stalled this *for now* because of time and personal issues. I hope I can pick this up at a later point in time. Not forgotten, but not a priority for me.

ok2c:
I am no longer in favor of inclusion of SPNEGO authentication into HttpClient code base given the avalanche of LLM driven vulnerability reports of late and how long it takes to review and finalize this change-set

@stoty I propose SPNEGO auth scheme be released as a separate artifact from an independent project and HttpClient promotes it as a companion project. Once the project is mature enough we can revisit the idea of its incorporation into HttpClient.

michael-o:
> I am no longer in favor of inclusion of SPNEGO authentication into HttpClient code base given the avalanche of LLM driven vulnerability reports of late and how long it takes to review and finalize this change-set
> 
> @stoty I propose SPNEGO auth scheme be released as a separate artifact from an independent project and HttpClient promotes it as a companion project. Once the project is mature enough we can revisit the idea of its incorporation into HttpClient.

I agree under the condition that all necessary enteypoints are available in the base client.

ok2c:
> I agree under the condition that all necessary enteypoints are available in the base client.

@michael-o As far as I know they are. @stoty Please confirm.

PR #616, opened 2025-02-05
Async client can now be used internally by the Fluent facade

ok2c:
@arturobernalg Please review

This change-set adds support for async client as a backend for the Fluent Facade and enables its users to make use of HTTP/2 transport


arturobernalg (approved):
LGTM

PR #618, opened 2025-02-25
Enhance filename encoding in multipart/form-data (HTTPCLIENT-2360)

arturobernalg:
Implements RFC 6266/5987 filename encoding in multipart/form-data. Adds mode-aware support in FormBodyPartBuilder and HttpRFC7578Multipart, propagates mode via MultipartEntityBuilder, and updates tests. Resolves https://issues.apache.org/jira/browse/HTTPCLIENT-2360.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FormBodyPartBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FormBodyPartBuilder.java
@@ -27,12 +27,15 @@
 
 package org.apache.hc.client5.http.entity.mime;
 
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.hc.core5.http.ContentType;
 import org.apache.hc.core5.http.NameValuePair;
 import org.apache.hc.core5.http.message.BasicNameValuePair;
+import org.apache.hc.core5.net.PercentCodec;
 import org.apache.hc.core5.util.Args;
 import org.apache.hc.core5.util.Asserts;
 
@@ -47,22 +50,52 @@ public class FormBodyPartBuilder {
     private ContentBody body;
     private final Header header;
 
+    /**
+     * The multipart mode determining how filenames are encoded in the {@code Content-Disposition}
+     * header, defaults to {@link HttpMultipartMode#STRICT}.
+     *
+     * @since 5.5
+     */
+    private HttpMultipartMode mode;
+
+    /**
+     * Encoder used to check if strings can be encoded in ISO-8859-1, supporting filename
+     * compatibility determinations in multipart form data.
+     */
+    private CharsetEncoder iso8859_1Encoder;
+
+    /**
+     * Creates a new builder instance with the specified name, content body, and multipart mode.
+     *
+     * @param name the name of the form field
+     * @param body the content body of the part
+     * @param mode the {@link HttpMultipartMode} to use, determining filename encoding behavior;
+     *
+     * @return a new {@code FormBodyPartBuilder} instance
+     * @since 5.5
+     */
+    public static FormBodyPartBuilder create(final String name, final ContentBody body, final HttpMultipartMode mode) {
+        return new FormBodyPartBuilder(name, body, mode);
+    }
+
     public static FormBodyPartBuilder create(final String name, final ContentBody body) {
-        return new FormBodyPartBuilder(name, body);
+        return new FormBodyPartBuilder(name, body, HttpMultipartMode.STRICT);
     }
 
     public static FormBodyPartBuilder create() {
         return new FormBodyPartBuilder();
     }
 
-    FormBodyPartBuilder(final String name, final ContentBody body) {
+    FormBodyPartBuilder(final String name, final ContentBody body, final HttpMultipartMode mode) {
         this();
         this.name = name;
         this.body = body;
+        this.mode = mode != null ? mode : HttpMultipartMode.STRICT;
     }
 
     FormBodyPartBuilder() {
         this.header = new Header();
+        this.mode = HttpMultipartMode.STRICT;
     }
 
     public FormBodyPartBuilder setName(final String name) {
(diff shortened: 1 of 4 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FormBodyPartBuilder.java
```diff
@@ -96,12 +99,61 @@ public FormBodyPartBuilder setField(final String name, final String value) {
...
+     * </p>
+     * <p>
+     * If {@code mode} is {@code null}, it defaults to {@link HttpMultipartMode#STRICT}.
+     * </p>
+     *
+     * @param mode the {@link HttpMultipartMode} to use, or {@code null} for default behavior
+     * @return this builder instance for method chaining
+     * @since 5.5
+     */
+    public FormBodyPartBuilder setMode(final HttpMultipartMode mode) {
```
ok2c, line 120: @arturobernalg Would not it be better to pass this parameter at the construction time to `#create` method?

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FormBodyPartBuilder.java
```diff
@@ -96,12 +99,61 @@ public FormBodyPartBuilder setField(final String name, final String value) {
...
+     * Determines whether the given string can be encoded in ISO-8859-1 without loss of data.
+     * This is used to decide whether the {@code filename} parameter can be used as-is or if
+     * the {@code filename*} parameter is needed for non-ISO-8859-1 characters.
+     *
+     * @param input the string to check, must not be {@code null}
+     * @return {@code true} if the string can be encoded in ISO-8859-1, {@code false} otherwise
+     * @since 5.5
+     */
+    private static boolean canEncodeToISO8859_1(final String input) {
+        return StandardCharsets.ISO_8859_1.newEncoder().canEncode(input);
```
ok2c, line 141: @arturobernalg Let's make the method non-static re-use the encoder rather than creating a new instance with every invocation

httpclient5/src/test/java/org/apache/hc/client5/http/entity/mime/TestMultipartForm.java
```diff
@@ -295,11 +295,11 @@ void testMultipartFormBrowserCompatibleNonASCIIHeaders() throws Exception {
         @SuppressWarnings("resource")
         final FormBodyPart p1 = FormBodyPartBuilder.create(
                 "field1",
-                new InputStreamBody(new FileInputStream(tmpfile), s1 + ".tmp")).build();
+                new InputStreamBody(new FileInputStream(tmpfile), s1 + ".tmp")).setMode(HttpMultipartMode.LEGACY).build();
```
ok2c, line 298: @arturobernalg Would not it look better as a `#create` parameter?

ok2c (commented):
@arturobernalg Looks good. Some minor suggestions.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FormBodyPartBuilder.java
```diff
@@ -47,22 +50,52 @@ public class FormBodyPartBuilder {
...
+     *
+     * @since 5.5
+     */
+    private HttpMultipartMode mode;
+
+    /**
+     * Reusable encoder for ISO-8859-1 charset checks, improving performance by avoiding
+     * repeated instance creation.
+     */
+    private final CharsetEncoder iso8859_1Encoder = StandardCharsets.ISO_8859_1.newEncoder();
```
ok2c, line 65: @arturobernalg I would instantiate this attribute lazily. It does not need to be final. Looks good overall.

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/FormBodyPartBuilder.java
arturobernalg: @ok2c changed.

ok2c (approved):
@arturobernalg Nicely done.

ok2c:
@arturobernalg While you are at it, do you think you could remove the code that randomly generates boundary values with a fixed boundary and update the javadocs to instruct the users to use generate their own boundary values if they want them to be random / non-predictable / 'secure'?

arturobernalg:
> @arturobernalg While you are at it, do you think you could remove the code that randomly generates boundary values with a fixed boundary and update the javadocs to instruct the users to use generate their own boundary values if they want them to be random / non-predictable / 'secure'?

@ok2c sure. I'll do

surlemur:
Question: Wouldn't it be more standard compliant to render the filename* attribute whenever the string is not US-ASCII encodable? E.g., Embedded Tomcat 10.1 does not properly parse the ISO-8859-1 filename content, but correctly handles filename*. It seems to me that the default ISO-8859-1 assumption, although mentioned in some RFCs, is not consistently true in reality.

arturobernalg:
> Question: Wouldn't it be more standard compliant to render the filename* attribute whenever the string is not US-ASCII encodable? E.g., Embedded Tomcat 10.1 does not properly parse the ISO-8859-1 filename content, but correctly handles filename*. It seems to me that the default ISO-8859-1 assumption, although mentioned in some RFCs, is not consistently true in reality.

RFC 7578 §4.2 is crystal clear:

    **NOTE: The encoding method described in RFC 5987, which would add a filename* parameter to the Content-Disposition header field, MUST NOT be used.**

HttpClient 5.5 therefore emits filename* only when the filename contains code points that ISO-8859-1 cannot represent. Everything in the 0x00-0xFF range already has a legal encoding, so adding filename* there would break the letter of the spec and has no real-world upside.

If embedded Tomcat 10.1 stumbles over an ISO-8859-1 filename, that is a Tomcat issue, not a protocol requirement. Use HttpMultipartMode.EXTENDED (percent-encoded names) as a workaround or open a bug with Tomcat, but let’s keep HttpClient on the right side of the standards

surlemur:
Ok. I must admit it's hard to grasp the combination of all those related RFCs plus the different traditions...

The pragmatic solution seems to be to assume ISO-8859-1 as the default on the server side when parsing the filename property. I am using the Apache Commons FileUpload and realized that I can set the header encoding there.

Thanks!

arturobernalg:
> Ok. I must admit it's hard to grasp the combination of all those related RFCs plus the different traditions...
> 
> The pragmatic solution seems to be to assume ISO-8859-1 as the default on the server side when parsing the filename property. I am using the Apache Commons FileUpload and realized that I can set the header encoding there.
> 
> Thanks!

@surlemur ,

Correct. RFC 7578 treats the raw octets of a filename parameter as ISO-8859-1 unless they are percent-encoded or a filename*= override is present (see the NOTE in §4.2).

 Glad it’s sorted.

PR #621, opened 2025-03-05
Fix PublicSuffixMatcherLoader#getDefault

joegallo:
The code of `getDefault` reads like it's supposed to load the entries from `PUBLIC_SUFFIX_LIST` by default, but it doesn't actually work that way in practice. I think this is a bug. This works correctly on 4.x, but the list there already starts with a leading slash:

https://github.com/apache/httpcomponents-client/blob/54900db4653d7f207477e6ee40135b88e9bcf832/httpclient/src/main/java/org/apache/http/conn/util/PublicSuffixMatcherLoader.java#L86

This was sneaking past the tests, because the test for getDefault only checks that it isn't `null`, but a non-null nearly empty default PublicSuffixMatcher is still provided in the case of the url not existing, and the tests for PublicSuffixMatcher themselves reimplement the default loading logic (correctly, since they rely on `ClassLoader#getResource` rather than `Class#getResource`) and so they weren't running into the bug either.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcherLoader.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcherLoader.java
@@ -51,7 +51,7 @@ import org.slf4j.LoggerFactory;
 @Contract(threading = ThreadingBehavior.SAFE)
 public final class PublicSuffixMatcherLoader {
 
-    private static final String PUBLIC_SUFFIX_LIST = "org/publicsuffix/list/effective_tld_names.dat";
+    private static final String PUBLIC_SUFFIX_LIST = "/org/publicsuffix/list/effective_tld_names.dat";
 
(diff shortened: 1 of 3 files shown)


ok2c:
@joegallo That you for contributing this fix.

ok2c:
Cherry-picked to `5.4.x`

KeekLove:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2025-27820
Will it affect httpclient4?

arturobernalg:
> https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2025-27820 Will it affect httpclient4?

Only 5.4.0, 5.4.1, and 5.4.2, explicitly excluding 5.4.3

KeekLove:
> > https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2025-27820 Will it affect httpclient4?
> 
> Only 5.4.0, 5.4.1, and 5.4.2, explicitly excluding 5.4.3

Thank you so much

PR #625, opened 2025-03-17
Use a random boundary value by default

benweissmann:
See more context in discussion from #619.

The advantage of using a random boundary by default are:
- It does not represent a semantically breaking change with the currently-released version (unlike switching from ThreadLocalRandom to a fixed boundary by default, like the current code does), [as noted here](https://github.com/apache/httpcomponents-client/pull/619#issuecomment-2731261012).
- It's in line with the behavior recommended by RFC 2046, [as discussed here](https://github.com/apache/httpcomponents-client/pull/619#issuecomment-2727040042)
- It's in line with most [other](https://github.com/urllib3/urllib3/blob/3e8f2db735dcaced6a3b777aa1966f40c018af7c/src/urllib3/filepost.py#L26) [MultipartEntity](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/mime/multipart/writer.go;l=87) [libraries](https://github.com/nodejs/undici/blob/f11ae944dab697d2dc22318403200b144b4b2b91/lib/web/fetch/body.js#L121) so will minimize surprises for users who have used multipart encoding in other languages or libraries, and doesn't require us to reason about the operational or security considerations of a novel or uncommon boundary selection scheme
- It does not risk mis-encoding content that happens to contain the hardcoded boundary, which could happen for nested bodies (while this library does not provide affordances for constructing such bodies, there's no prohibition around passing a body constructed by this library as a part of another body, and there could easily be in-the-wild code that does so), or for content that happens to be discussing the hard-coded boundary value itself (consider, for example, a Java-based Github client, and how it would handle posting a multipart/form-encoded comment on this repo discussing the boundary value).
- It avoids attacks like the one tracked as [CVE-2025-22150](https://nvd.nist.gov/vuln/detail/CVE-2025-22150) where an attacker can trick an application into including unexpected fields via a crafted payload that contains the boundary value.
  - Of note, the current default behavior makes this attack much, much simpler than the one in that CVE, because an attacker does not need to guess the random value from an insecure random number generator; all existing libraries and applications that use httpcomponents-client will start using the new hard-coded default value when they upgrade unless they explicitly opt-in to the new random behavior or were already generating their own boundary values.
  - As a practical, in-the-wild example of this, consider this [Modern Treasury Kotlin library function](https://github.com/Modern-Treasury/modern-treasury-kotlin/blob/cb41cbb555312728a6fb9f33285e838f502ed957/modern-treasury-kotlin-core/src/main/kotlin/com/moderntreasury/api/services/blocking/DocumentServiceImpl.kt#L31), which uses [uses our MultipartFormBuilder](https://github.com/Modern-Treasury/modern-treasury-kotlin/blob/cb41cbb555312728a6fb9f33285e838f502ed957/modern-treasury-kotlin-core/src/main/kotlin/com/moderntreasury/api/core/HttpRequestBodies.kt#L45) API to hit [the Create Document](https://docs.moderntreasury.com/platform/reference/create-document) api endpoint.
  - Consider the most common likely usage of this function:
    - A Kotlin web application allows users to upload documents, like I-9 forms or other financial or compliance paperwork
    - The server-side component of this application receives the document from an authenticated user, and then uses this API method to attach the user-provided document to corresponding Modern Treasury entity. For example, it might take an employee's I-9 (employment verification form) and attach it to that employee's bank account record in a payroll system.
    - A malicious user could upload a PDF that ends with 

        ```
        <regular PDF data>
        --httpclient_boundary_7k9p2m4x8n5j3q6t1r0vwyzabcdefghi
        Content-Disposition: form-data; name="documentable_id"
        Content-Type: text/plain

        some-other-account-id
        ```

  - Now, instead of the server-side component of this application sending the user-provided PDF and the server-provided "documentable_id", it would be tricked into sending an attacker-controlled documentable_id, which would associate the financial or compliance paperwork with the wrong account.

While there is disagreement from some maintainers on some of these points, I would contend that *any one* of these five arguments should be sufficient to use the secure-random UUID default behavior, rather than the currently-implemented fixed-value default (or, for that matter, the currently-released insecure random default).


httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java
```diff
@@ -47,18 +47,8 @@
...
- * This class constructs multipart entities with a boundary determined by either a fixed
- * value ("httpclient_boundary_7k9p2m4x8n5j3q6t1r0vwyzabcdefghi") or a random UUID. If no
- * boundary is explicitly set via {@link #setBoundary(String)}, it defaults to the fixed
- * value unless {@link #withRandomBoundary()} is called to request a random UUID at build
- * time. Users can provide a custom boundary with {@link #setBoundary(String)}. A warning
- * is logged when no explicit boundary is set via {@link #setBoundary(String)}, encouraging
- * deliberate choice.
- * </p>
- * <p>
- * IMPORTANT: it is responsibility of the caller to validate / sanitize content of body
```
ok2c, line 59: @benweissmann Restore this section. It says IMPORTANT for a reason.

benweissmann: Done -- i had added a similar `IMPORTANT` note in `setBoundary` to note the considerations specifically around boundary values, but happy to leave it here as well.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java
```diff
@@ -234,14 +225,12 @@ public MultipartEntityBuilder addBinaryBody(final String name, final InputStream
...
-    private String getFixedBoundary() {
-        return BOUNDARY_PREFIX + "7k9p2m4x8n5j3q6t1r0vwyzabcdefghi";
-    }
-
-    /**
-     * Generates a random boundary using UUID.
+     * Generates a random boundary using UUID. The UUID is a v4 random UUID generated from a cryptographically-secure
+     * random source.
+     * <p>
+     * A cryptographically-secure random number source is used to avoid security issues similar to
```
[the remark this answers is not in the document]
benweissmann: Let me know what you think of the revised note -- I've removed the reference to the CVE, but I think it's important to note the source (and randomness guarantees) of the UUID for two reasons:
- As you've discussed, it's the responsibility of the caller to enforce any security model for their application, so we should document the source of the value here so they can make those decisions.
- So future contributors know that the choice of randomness here was intentional, and to consider that if there's any future changes to the default boundary value selection

ok2c (approved):
@benweissmann That sounds reasonable.

potiuk (approved):

TomerAberbach:
Hey folks, just wanted to check if the plan is to merge this soon? Before the next release?

PR #627, opened 2025-03-25
Enhance ProtocolSwitchStrategy with Simplified Protocol Parsing

arturobernalg:
refines the ProtocolSwitchStrategy class to improve the parsing of protocol tokens in the Upgrade header, making the implementation more elegant and future-proof.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
@@ -27,15 +27,22 @@
 package org.apache.hc.client5.http.impl;
 
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.http.FormattedHeader;
+import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpHeaders;
 import org.apache.hc.core5.http.HttpMessage;
+import org.apache.hc.core5.http.HttpVersion;
 import org.apache.hc.core5.http.ParseException;
 import org.apache.hc.core5.http.ProtocolException;
 import org.apache.hc.core5.http.ProtocolVersion;
-import org.apache.hc.core5.http.message.MessageSupport;
+import org.apache.hc.core5.http.ProtocolVersionParser;
 import org.apache.hc.core5.http.ssl.TLS;
+import org.apache.hc.core5.util.Args;
+import org.apache.hc.core5.util.CharArrayBuffer;
+import org.apache.hc.core5.util.Tokenizer;
 
 /**
  * Protocol switch handler.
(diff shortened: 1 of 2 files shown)


arturobernalg:
> @arturobernalg Is there a reason for not using `ProtocolVersionParser` from core?

@ok2c  please do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +49,96 @@
...
+        } else if (httpUpgrade != null) {
+            return httpUpgrade;
+        } else {
             throw new ProtocolException("Invalid protocol switch response");
         }
-        return tlsUpgrade;
     }
 
-}
+    private ProtocolVersion parseTlsToken(final String token) throws ParseException {
```
ok2c, line 88: @arturobernalg Am I missing something? There is a `parse`method in `ProtocolVersionParser` that should be able to do all this with one line of code.

```
MatcherAssert.assertThat(ProtocolVersion.parse(" PROTO/1.2,0000", cursor, Tokenizer.delimiters(',')), CoreMatchers.equalTo(new ProtocolVersion("PROTO", 1, 2)));
```

arturobernalg: @ok2c The reason I didn’t go straight for it was to avoid cases like this: if we get something weird like `response.addHeader(HttpHeaders.UPGRADE, "HTTP/11.22")` without batting an eye, and I wanted to catch that and only allow our supported versions

ok2c: @arturobernalg My proposal is that the strategy parses each token into `ProtocolVersion` with a single `#parse` method, validates it and decides if the upgrade is supported and can be executed. Presently we only support `TLS` and we can also tolerate (ignore) HTTP/1.1 because this is effectively a no-op. All other upgrade protocols should be rejected as unsupported

ok2c: I see no reason for having two almost completely identical parse methods for `TLS` and `HTTP`. All tokens should be parsed the same way and then validated.

arturobernalg: @ok2c simplified

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +48,53 @@
...
-            final String token = it.next();
-            if (token.startsWith("TLS")) {
-                // TODO: Improve handling of HTTP protocol token once HttpVersion has a #parse method
-                try {
-                    tlsUpgrade = token.length() == 3 ? TLS.V_1_2.getVersion() : TLS.parse(token.replace("TLS/", "TLSv"));
-                } catch (final ParseException ex) {
-                    throw new ProtocolException("Invalid protocol: " + token);
+            final String token = it.next().trim();
+            try {
+                if (token.startsWith("TLS") || token.startsWith("HTTP/")) {
```
ok2c, line 61: @arturobernalg This line looks completely unnecessary. Why? Parse first - validate next.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +48,53 @@
...
+                    final ProtocolVersion version = parseToken(token);
+                    if (token.startsWith("TLS")) {
```
ok2c, line 63: @arturobernalg Should not it be `#equalsIgnoreCase`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +48,53 @@
...
+                        tlsUpgrade = version;
+                    } else {
+                        httpUpgrade = version;
+                    }
+                } else {
+                    throw new ProtocolException("Unsupported protocol: " + token);
+                }
+            } catch (final ParseException ex) {
+                if (token.startsWith("TLS")) {
```
ok2c, line 72: @arturobernalg Why? All this is unnecessary. Just throw `new ProtocolException("Unsupported protocol: " + protocolVersion)`

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +48,53 @@
...
-                } catch (final ParseException ex) {
-                    throw new ProtocolException("Invalid protocol: " + token);
+            final String token = it.next().trim();
+            try {
+                if (token.startsWith("TLS") || token.startsWith("HTTP/")) {
+                    final ProtocolVersion version = parseToken(token);
+                    if (token.startsWith("TLS")) {
+                        tlsUpgrade = version;
+                    } else {
+                        httpUpgrade = version;
```
ok2c, line 66: @arturobernalg Why? We should only tolerate (do nothing) if it is `HTTP/1.1`. All other versions should be rejected

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +48,53 @@
...
+            return httpUpgrade;
+        } else {
             throw new ProtocolException("Invalid protocol switch response");
         }
-        return tlsUpgrade;
     }
 
+    private ProtocolVersion parseToken(final String token) throws ParseException {
+        if (token.equals("TLS")) {
+            return TLS.V_1_2.getVersion();
```
ok2c, line 93: @arturobernalg Why? I do not understand why you want to treat all `TLS` versions as `TLS/1.2`.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
arturobernalg: @ok2c  Without this check, the parser would throw a ParseException, failing the operation for example here `        Assertions.assertEquals(TLS.V_1_2.getVersion(), switchStrategy.switchProtocol(response1));`

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +49,58 @@
...
+                return TLS.V_1_2.getVersion();
+            }
+
+            final CharArrayBuffer buffer = new CharArrayBuffer(token.length());
+            buffer.append(token);
+            final Tokenizer.Cursor cursor = new Tokenizer.Cursor(0, token.length());
+
+            final ProtocolVersion version = PROTOCOL_VERSION_PARSER.parse(buffer, cursor, null);
+
+            if ("TLS".equalsIgnoreCase(version.getProtocol())) {
```
ok2c, line 92: @arturobernalg `TLS` has its own protocol version negotiation handshake. I would not reject an upgrade response here and let TLS do its job negotiating the exact protocol versions and rejecting those it does not understand or support.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +49,58 @@
...
+                if (!(TLS.V_1_0.getVersion().lessEquals(version) && version.lessEquals(TLS.V_1_3.getVersion()))) {
+                    throw new ProtocolException("Unsupported TLS version: " + token);
+                }
+                return version;
+            } else if (HttpVersion.HTTP.equalsIgnoreCase(version.getProtocol()) && version.equals(HttpVersion.HTTP_1_1)) {
```
ok2c, line 97: @arturobernalg Should not `version.equals(HttpVersion.HTTP_1_1)` do exactly the same?

ok2c (changes requested):

ok2c (commented):
@arturobernalg So much better in my opinion. Now to make it better still, please consider copying `#parseHeaders` method from https://github.com/apache/httpcomponents-core/commit/99630f4118550eeedf19ba75b3400e6e1c035c9f and using it to parse the sequence of protocol tokens without making intermediate String objects

arturobernalg:
> @arturobernalg So much better in my opinion. Now to make it better still, please consider copying `#parseHeaders` method from [apache/httpcomponents-core@99630f4](https://github.com/apache/httpcomponents-core/commit/99630f4118550eeedf19ba75b3400e6e1c035c9f) and using it to parse the sequence of protocol tokens without making intermediate String objects

All right @ok2c I think now i have full in all the remarks

httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -45,31 +52,106 @@
...
+        }
+    }
+
+    private ProtocolVersion parseProtocolToken(final CharSequence buffer, final int start, final int end)
+            throws ProtocolException {
+        if (start >= end) {
+            return null;
+        }
+
+        if (end - start == 3) {
```
ok2c, line 104: @arturobernalg I think there is  a way to make it nicer, but it is good enough for now

ok2c (approved):

ok2c:
@arturobernalg I took a closer look at your code and had to refactor a few things. Please take a look at 0ba6102e. I hope you agree implementation is shorter and cleaner. I also had to fix the Exception message asserts in the test case.

arturobernalg:
> @arturobernalg I took a closer look at your code and had to refactor a few things. Please take a look at [0ba6102](https://github.com/apache/httpcomponents-client/commit/0ba6102e65d793d4b4106e5b7bfabef7ec2d0a66). I hope you agree implementation is shorter and cleaner. I also had to fix the Exception message asserts in the test case.

HI @ok2c look good. Thank for the review.

PR #630, opened 2025-04-04
HTTPCLIENT-2366 - Fix isRepeatable method in RequestEntityProxy to return true before consumption.

arturobernalg:
This PR fixes an issue with the `isRepeatable` method in `RequestEntityProxy`. Previously, if the underlying entity was non-repeatable, the proxy would incorrectly return false even before consumption. The fix ensures that the proxy is always considered repeatable until it is consumed, regardless of the underlying entity's repeatability. Tests have been updated to validate this behavior.


garydgregory (approved):
LGTM! TY @arturobernalg

ok2c:
@arturobernalg Please cherry-pick to `5.4.x`

arturobernalg:
> @arturobernalg Please cherry-pick to `5.4.x`

done

PR #634, opened 2025-05-01
Fix NPE in InternalAbstractHttpAsyncClient by adding null check for resolvedTarget

arturobernalg:
Resolves a `NullPointerException` in `InternalAbstractHttpAsyncClient` by adding a null check for resolvedTarget when RoutingSupport.determineHost returns null. Throws `IllegalStateException` for invalid requests, ensuring robust error handling. Addresses [HTTPCLIENT-2367.](https://issues.apache.org/jira/browse/HTTPCLIENT-2367)

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalAbstractHttpAsyncClient.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalAbstractHttpAsyncClient.java
@@ -219,12 +219,15 @@ abstract class InternalAbstractHttpAsyncClient extends AbstractHttpAsyncClientBa
                 setupContext(clientContext);
 
                 final HttpHost resolvedTarget = target != null ? target : RoutingSupport.determineHost(request);
-                if (request.getScheme() == null) {
-                    request.setScheme(resolvedTarget.getSchemeName());
+                if (resolvedTarget != null) {
+                    if (request.getScheme() == null) {
+                        request.setScheme(resolvedTarget.getSchemeName());
+                    }
+                    if (request.getAuthority() == null) {
+                        request.setAuthority(new URIAuthority(resolvedTarget));
+                    }
                 }
-                if (request.getAuthority() == null) {
-                    request.setAuthority(new URIAuthority(resolvedTarget));
-                }
+
                 final HttpRoute route = determineRoute(
                         resolvedTarget,
                         request,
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/InternalHttpClient.java
```diff
@@ -161,11 +161,15 @@ protected CloseableHttpResponse doExecute(
             setupContext(localcontext);
 
             final HttpHost resolvedTarget = target != null ? target : RoutingSupport.determineHost(request);
-            if (request.getScheme() == null) {
-                request.setScheme(resolvedTarget.getSchemeName());
-            }
-            if (request.getAuthority() == null) {
-                request.setAuthority(new URIAuthority(resolvedTarget));
+            if (resolvedTarget != null) {
```
garydgregory, line 164: HI @arturobernalg 👋 
We could throw a more precise exception using `Objects.requireNonNull(...)`, which would reduce the nesting.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalAbstractHttpAsyncClient.java
```diff
@@ -219,6 +220,7 @@ protected <T> Future<T> doExecute(
                 setupContext(clientContext);
 
                 final HttpHost resolvedTarget = target != null ? target : RoutingSupport.determineHost(request);
+                Args.notNull(resolvedTarget, "Target host for the request");
```
ok2c, line 223: @arturobernalg This is not right. The target may be null if the request URI cannot be parsed. It should be up to the specific protocol handler to decide if such a request can be executed. H2 will likely reject it but HTTP/1.1 may accept it

ok2c (changes requested):

ok2c (approved):

arturobernalg:
@ok2c  should I wait for the original reporter’s feedback before merging?

ok2c:
> @ok2c should I wait for the original reporter’s feedback before merging?

@arturobernalg I do not think it is necessary. The fix is quite straight-forward.

ok2c:
@arturobernalg I tweaked the test case a little to make it behave it more similarly to the default route planner.

garydgregory:
👍

arturobernalg:
> @arturobernalg I tweaked the test case a little to make it behave it more similarly to the default route planner.

thank you @ok2c

PR #635, opened 2025-05-22
Add helper script to run example programs

rschmitt:
This commit adds a simple shell script that runs our example programs. I've also fixed a bug in `ClientAbortMethod` that prevents the JVM from shutting down.


arturobernalg:
I think you miss license header.

ok2c (approved):

arturobernalg (approved):
LGTM

PR #636, opened 2025-05-22
Add Maven wrapper

rschmitt:


ok2c (approved):

garydgregory:
Almost -1: I avoid these wrappers myself since they tend to go stale and no one maintains them. They are confusing at best. I also never want to have to remember what directory I need to type "mvn" vs whatever the wrapper is called and I certainly don't want have it force some version of Maven on me, especially by downloading it. I also don't want to update any of my scripts or rewrite instructions on the site and wiki, which this PR doesn't do and demonstrates one of the problems. The PR description is empty and the reader file doesn't even describe when or why I need this thing, or if it's a requirement or what.

arturobernalg (approved):
@rschmitt  I'm generally not a fan of Maven Wrapper because it often goes stale and adds unnecessary complexity for little benefit.

rschmitt:
> I'm generally not a fan of Maven Wrapper because it often goes stale and adds unnecessary complexity for little benefit.

I'm surprised by this. The Maven wrapper is quite convenient. For instance, my Linux desktop doesn't actually have a version of Maven recent enough to build HttpComponents, and by far the easiest way to _obtain_ one was to set up the wrapper. I also invoke the wrapper from https://github.com/apache/httpcomponents-client/pull/635 to help ensure that the example runner just works.

I think it would be a good idea in principle to use the Maven wrapper in the GitHub workflows, but it's not a huge difference either way since Maven is fairly stable. (With a Gradle project, by contrast, using the wrapper would be absolutely non-negotiable.)

ok2c:
@rschmitt Gradle is a whole different story. I see no harm in having Maven wrapper. It may even become useful if we decide to upgrade to Maven 4. Just please apply the same config to core as well.

PR #641, opened 2025-06-04
Add org.apache.hc.client5.http.entity.mime.PathBody

garydgregory:
- Add org.apache.hc.client5.http.entity.mime.PathBody
- Add org.apache.hc.client5.http.entity.mime.MultipartEntityBuilder.addBinaryBody(String, Path)
- Add org.apache.hc.client5.http.entity.mime.MultipartEntityBuilder.addBinaryBody(String, Path, ContentType, String)


pom.xml
```diff
@@ -192,6 +192,12 @@
         <artifactId>junit-jupiter</artifactId>
         <version>${testcontainers.version}</version>
       </dependency>
+      <dependency>
+        <groupId>commons-io</groupId>
+        <artifactId>commons-io</artifactId>
+        <version>2.19.0</version>
+        <scope>test</scope>
```
ok2c: @garydgregory I think the scope has no effect here and should be removed

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java
```diff
@@ -215,6 +216,32 @@ public MultipartEntityBuilder addBinaryBody(
         return addPart(name, new FileBody(file, contentType, filename));
     }
 
+    /**
+     * Adds body with contents from the given source Path.
+     *
+     * @param name The part name.
+     * @param path The source path.
+     * @return {@code this} instance.
+     * @since 5.5.1
```
ok2c: @garydgregory This looks wrong. It should be 5.6

ok2c (changes requested):

pom.xml
garydgregory: Thank you for the review @ok2c , fixed, squashed, and pushed.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java
garydgregory: Fixed, squashed, and pushed.

PR #642, opened 2025-06-04
TestHttpRoute: Fix Windows compatibility

rschmitt:


garydgregory (approved):
Works for me on `Microsoft Windows [Version 10.0.19045.5737]`.

TY @rschmitt 💯

ok2c:
@rschmitt ~~If the integration tests start failing again I will have to revert this change.~~ Scratch that. GitHub CI tests on Winsdows were disabled for a reason. We can try enabling them again, but as long as the original problem does not reoccur.

garydgregory:
This fix solves building on my Windows box. IMO, after we merge this I think we can experiment with Windows builds in the CI again.

PR #643, opened 2025-06-04
HTTPCLIENT-2372 - Normalize HttpHost port comparison to treat implicit default ports as equal

arturobernalg:
This change ensures that hosts with implicit default ports (e.g., http://example.com) are treated the same as those explicitly specifying port 80 or 443. It introduces a helper method to derive the “effective” port and updates the authority check to use this normalized value. As a result, sensitive headers are no longer stripped when redirecting between example.com and example.com:80 (or :443 for HTTPS).


httpclient5/src/main/java/org/apache/hc/client5/http/impl/DefaultRedirectStrategy.java
```diff
@@ -80,6 +83,36 @@ public boolean isRedirectAllowed(
...
+        final String host2 = h2.getHostName();
+        if (!host1.equalsIgnoreCase(host2)) {
+            return false;
+        }
+        final int port1 = getEffectivePort(h1);
+        final int port2 = getEffectivePort(h2);
+        return port1 == port2;
+    }
+
+    private int getEffectivePort(final HttpHost endpoint) {
```
ok2c: @arturobernalg Please use `SchemePortResolver` to resolve the port of a protocol scheme. Please also use the one created by the builder. See `HttpClientBuilder`

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/DefaultRedirectStrategy.java
arturobernalg: @ok2c please do another pass

PR #644, opened 2025-06-04
Add integration tests for TLS handshake timeouts

rschmitt:
This change adds basic integration test coverage for TLS handshake timeouts for the sync and async clients. The tests make use of a special test server that times out a single TLS connection attempt and can be configured to time out at two different points in the TLS 4-way handshake.

Note that the TLS handshake timeout, as currently implemented, works like a socket timeout for the TLS handshake phase of the connection: it only limits the amount of time that will be spent on each individual socket read/write operations, not the total time spent in the handshake attempt. The timeout server, for example, could inject a delay before sending the Server Hello, which would cause the client to spend up to double the configured timeout attempting to complete the handshake. There is no test coverage for this behavior, but it could be added if we decided that it should be part of the feature's contract.

### DIFF
--- a/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestTlsHandshakeTimeout.java
+++ b/httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestTlsHandshakeTimeout.java
@@ -1,0 +1,118 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


ok2c:
@rschmitt Presently all timeouts supported by HC represent the maximum period of inactivity between consecutive i/o operations and not a deadline. There is feature request for request execution deadline support. It would not be easy to implement with the async transport and likely impossible with the classic transport.

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/sync/TestTlsHandshakeTimeout.java
```diff
@@ -0,0 +1,118 @@
...
+            server.start();
+
+            final HttpUriRequestBase request = new HttpGet("https://127.0.0.1:" + server.getPort());
+            assertTimeout(request, client);
+        }
+    }
+
+    @SuppressWarnings("deprecation")
+    private static void assertTimeout(final ClassicHttpRequest request, final HttpClient client) {
+        final String javaVersion = System.getProperty("java.specification.version");
```
ok2c: @rschmitt There is a method `ReflectionUtils#determineJRELevel` that core and client code uses in many instances to determine the actual JRE level. For the sake of consistency you may want to use it here as well.

ok2c (approved):

PR #645, opened 2025-06-05
Omit cookie spec registry when cookie management is disabled

rschmitt:
The HTTP client builders always construct a client with a non-null cookie spec registry; if a custom instance is not specified, then the builders will use `CookieSpecSupport.createDefault()`, which loads the full public suffix list. The PSL rules are stored on the heap as a `ConcurrentHashMap`, i.e. a giant array (16,384 on my machine) containing map entries, each of which is also allocated as a separate object on the heap. This is a considerable number of live objects that the garbage collector may need to trace. I ran a simple microbenchmark to show the effects of loading the PSL on garbage collection (these numbers are from JDK 1.8):

    GC took 2 ms
    GC took 2 ms
    GC took 2 ms
    Loading public suffix list... done (took 193 ms)
    GC took 11 ms
    GC took 8 ms
    GC took 7 ms

This is potentially a significant amount of tail latency for use cases that don't require cookie management, such as RPC calls. With this change, it is now simpler to configure and construct a client that does not load the PSL into memory.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/H2AsyncClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/H2AsyncClientBuilder.java
@@ -901,7 +901,7 @@ public class H2AsyncClientBuilder {
                     .build();
         }
         Lookup<CookieSpecFactory> cookieSpecRegistryCopy = this.cookieSpecRegistry;
-        if (cookieSpecRegistryCopy == null) {
+        if (cookieSpecRegistryCopy == null && !cookieManagementDisabled) {
             cookieSpecRegistryCopy = CookieSpecSupport.createDefault();
         }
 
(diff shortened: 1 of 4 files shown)


arturobernalg:
@rschmitt — shouldn’t we apply the same change to `HttpClientBuilder` as well?

rschmitt:
https://github.com/apache/httpcomponents-client/pull/645/files#diff-d08c7b09d26f3696bbee9660c8190724bfcc5804ea677301dcaa642e2dcb303a ?

ok2c (approved):

PascalSchumacher:
👍 but in most cases the full public suffix list will be loaded by `DefaultHostnameVerifier` later on.

rschmitt:
@PascalSchumacher I think this is the wrong default behavior and I plan on dealing with it next.

PR #646, opened 2025-06-06
Adjust default behavior to match intended logic

arturobernalg:
default behavior is incorrect and intends to address it. This commit realigns the code so that the behavior now matches the expected default, per his feedback.


garydgregory:
Hi @arturobernalg 

Shouldn't these changes be validated through unit tests?

rschmitt:
I don't understand the context here, nor the justification. Cookie management configuration shouldn't affect TLS configuration.

ok2c:
@arturobernalg Likewise, these changes do not look right to me. Cookie management and TLS hostname verification are completely unrelated.

arturobernalg:
This PR was trying to fix an unintended side‐effect—calling disableCookieManagement() was disabling hostname checks

PR #648, opened 2025-06-09
Add socket timeout integration tests

rschmitt:
This change adds integration test coverage for various types of read timeouts. The test coverage includes HTTP, HTTPS, and HTTP over UDS if available. The `/random` request handlers have been augmented with delay support, so that the clients can read the response headers and then time out while reading the entity. The tests make use of `ConnectionConfig`, `ResponseTimeout`, and (on the sync client only) `SocketConfig`, and demonstrate the precedence among these config options when more than one is supplied.

One issue uncovered by these tests is that graceful closure of the async client does not work when UDS socket timeouts have fired. To work around this until the issue can be root caused, the async UDS tests use `CloseMode.IMMEDIATE`.

### DIFF
--- a/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AsyncRandomHandler.java
+++ b/httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AsyncRandomHandler.java
@@ -43,6 +43,7 @@ import org.apache.hc.core5.http.HttpRequest;
 import org.apache.hc.core5.http.HttpResponse;
 import org.apache.hc.core5.http.HttpStatus;
 import org.apache.hc.core5.http.MethodNotSupportedException;
+import org.apache.hc.core5.http.NameValuePair;
 import org.apache.hc.core5.http.ProtocolException;
 import org.apache.hc.core5.http.message.BasicHttpResponse;
 import org.apache.hc.core5.http.nio.AsyncEntityProducer;
@@ -53,7 +54,10 @@ import org.apache.hc.core5.http.nio.ResponseChannel;
 import org.apache.hc.core5.http.nio.StreamChannel;
 import org.apache.hc.core5.http.nio.entity.AbstractBinAsyncEntityProducer;
 import org.apache.hc.core5.http.protocol.HttpContext;
+import org.apache.hc.core5.net.WWWFormCodec;
 import org.apache.hc.core5.util.Asserts;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * A handler that generates random data.
@@ -93,6 +97,22 @@ public class AsyncRandomHandler implements AsyncServerExchangeHandler {
         } catch (final URISyntaxException ex) {
             throw new ProtocolException(ex.getMessage(), ex);
         }
+        final String query = uri.getQuery();
+        int delayMs = 0;
+        boolean drip = false;
+        if (query != null) {
+            final List<NameValuePair> params = WWWFormCodec.parse(query, UTF_8);
+            for (final NameValuePair param : params) {
+                final String name = param.getName();
+                final String value = param.getValue();
+                if ("delay".equals(name)) {
+                    delayMs = Integer.parseInt(value);
+                } else if ("drip".equals(name)) {
+                    drip = "1".equals(value);
+                }
+            }
+        }
+
         final String path = uri.getPath();
         final int slash = path.lastIndexOf('/');
         if (slash != -1) {
@@ -109,7 +129,7 @@ public class AsyncRandomHandler implements AsyncServerExchangeHandler {
                 n = 1 + (int)(Math.random() * 79.0);
             }
             final HttpResponse response = new BasicHttpResponse(HttpStatus.SC_OK);
-            final AsyncEntityProducer entityProducer = new RandomBinAsyncEntityProducer(n);
+            final AsyncEntityProducer entityProducer = new RandomBinAsyncEntityProducer(n, delayMs, drip);
             entityProducerRef.set(entityProducer);
             responseChannel.sendResponse(response, entityProducer, context);
         } else {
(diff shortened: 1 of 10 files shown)


rschmitt:
I was startled to learn that `java.net.Socket::setSoTimeout` only sets _read_ timeouts:

https://github.com/openjdk/jdk/blob/d186dacdb7b91dc9a28b703ce3c8ea007fc450b6/src/java.base/share/classes/java/net/Socket.java#L1258-L1260

Given that this is the case, I'm not sure what prevents synchronous clients from going to sleep forever if a write blocks indefinitely because the send buffer filled up. Additionally, I haven't found a way to simulate this condition using localhost (at least not on macOS), so I can't even determine whether there's a sensible default value or something.

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AsyncRandomHandler.java
```diff
@@ -93,6 +93,20 @@ public void handleRequest(
         } catch (final URISyntaxException ex) {
             throw new ProtocolException(ex.getMessage(), ex);
         }
+        final String query = uri.getQuery();
```
ok2c, line 100: @rschmitt `WWWFormCodec#parse` may make parsing easier here.

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/classic/RandomHandler.java
```diff
@@ -84,6 +84,20 @@ public void handle(final ClassicHttpRequest request,
         } catch (final URISyntaxException ex) {
             throw new ProtocolException(ex.getMessage(), ex);
         }
+        final String query = uri.getQuery();
```
ok2c, line 93: @rschmitt Same.

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/classic/RandomHandler.java
```diff
@@ -185,31 +219,48 @@ public InputStream getContent() {
...
+                out.close();
             }
-            out.close();
+        }
 
+        private void delay() throws IOException {
+            try {
+                Thread.sleep(delayMs);
+            } catch (final InterruptedException ex) {
+                throw new IOException(ex);
```
ok2c: @rschmitt `InterruptedIOException` instead?

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AsyncRandomHandler.java
```diff
@@ -192,6 +216,10 @@ public int availableData() {
 
         @Override
         protected void produceData(final StreamChannel<ByteBuffer> channel) throws IOException {
+            if (System.currentTimeMillis() < deadline) {
+                return;
+            }
```
ok2c, line 227: @rschmitt With the current approach the data producer will make the i/o reactor spin. This is probably permissible for integration tests but ideally the data producer should signal it has no data to produce by returning 0 from `#availableData` and write data from a worker thread once it becomes available. This is of course requires quite a bit of work.

ok2c:
> I was startled to learn that `java.net.Socket::setSoTimeout` only sets _read_ timeouts:
> 
> https://github.com/openjdk/jdk/blob/d186dacdb7b91dc9a28b703ce3c8ea007fc450b6/src/java.base/share/classes/java/net/Socket.java#L1258-L1260
> 
> Given that this is the case, I'm not sure what prevents synchronous clients from going to sleep forever if a write blocks indefinitely because the send buffer filled up. Additionally, I haven't found a way to simulate this condition using localhost (at least not on macOS), so I can't even determine whether there's a sensible default value or something.

@rschmitt In the older version of HttpClient the socket timeout used to be called SO_READ_TIMEOUT. This is yet another limitation of the classic i/o one must tolerate

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AsyncRandomHandler.java
rschmitt: I KNEW it had to be somewhere

httpclient5-testing/src/main/java/org/apache/hc/client5/testing/async/AsyncRandomHandler.java
rschmitt: Yeah, I noticed this. Like you said, it's probably fine for an integration test.

rschmitt:
> This is yet another limitation of the classic i/o one must tolerate.

Yeah, this is probably one reason why virtually all server IO is non-blocking, even when the API is synchronous (see for example Tomcat).

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/async/TestAsyncSocketTimeout.java
```diff
@@ -0,0 +1,166 @@
...
+            final int majorVersion = Integer.parseInt(components[0]);
+            final int minorVersion = Integer.parseInt(components[1]);
+            assumeFalse(majorVersion <= 5 && minorVersion <= 3, "Async UDS requires HttpCore 5.4+");
+        }
+
+        @Override
+        void closeClient(final TestAsyncClient client) {
+            // TODO: Why does `CloseMode.GRACEFUL` (the test client default)
+            //  block until the IOReactor shutdown timeout when using UDS?
+            client.close(CloseMode.IMMEDIATE);
```
rschmitt, line 162: I still need to look into this some more. Based on the logspam I get during IOReactor shutdown, it's some sort of low-level selector issue. Could be a behavioral difference with UDS `SocketChannel`s.

rschmitt:
@ok2c Anything else on this one?

ok2c (approved):
@rschmitt Not really. Looks good to me

PR #651, opened 2025-06-17
HTTPCLIENT-1843  Plug Commons-Compress into HttpClient’s automatic

arturobernalg:
content-decoding (optional)

* New ContentDecoderRegistry discovers extra codecs (br, zstd, xz, lz4, …) via Commons-Compress when that jar is on the class-path; otherwise falls back to the built-ins (gzip, deflate) only.
* No hard dependency added—projects that need the extra algorithms just add `commons-compress` (and helper jars like google-brotli, zstd-jni, xz-java) to their pom and HttpClient uses them automatically.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/BrotliInputStreamFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/BrotliInputStreamFactory.java
@@ -42,6 +42,17 @@ import org.brotli.dec.BrotliInputStream;
 public class BrotliInputStreamFactory implements InputStreamFactory {
 
     /**
+     * Canonical token for the deflate content-coding.
+     * @since 5.6
+     */
+    public static final String ENCODING = "br";
+
+    @Override
+    public String getContentEncoding() {
+        return ENCODING;
+    }
+
+    /**
      * Default instance of {@link BrotliInputStreamFactory}.
      */
     private static final BrotliInputStreamFactory INSTANCE = new BrotliInputStreamFactory();
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/DeflateInputStreamFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/DeflateInputStreamFactory.java
@@ -42,6 +42,17 @@ import org.apache.hc.core5.annotation.ThreadingBehavior;
 public class DeflateInputStreamFactory implements InputStreamFactory {
 
     /**
+     * Canonical token for the deflate content-coding.
+     * @since 5.6
+     */
+    public static final String ENCODING = "deflate";
+
+    @Override
+    public String getContentEncoding() {
+        return ENCODING;
+    }
+
+    /**
      * Default instance of {@link DeflateInputStreamFactory}.
      */
     private static final DeflateInputStreamFactory INSTANCE = new DeflateInputStreamFactory();
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/GZIPInputStreamFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/GZIPInputStreamFactory.java
@@ -43,6 +43,17 @@ import org.apache.hc.core5.annotation.ThreadingBehavior;
 public class GZIPInputStreamFactory implements InputStreamFactory {
 
     /**
+     * Canonical token for the gzip content-coding.
+     * @since 5.6
+     */
+    public static final String ENCODING = "gzip";
+
+    @Override
+    public String getContentEncoding() {
+        return ENCODING;
+    }
+
+    /**
      * Default instance of {@link GZIPInputStreamFactory}.
      */
     private static final GZIPInputStreamFactory INSTANCE = new GZIPInputStreamFactory();
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
@@ -1,0 +1,117 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.http.entity.compress;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.Locale;
+import java.util.Map;
+
+import org.apache.commons.compress.compressors.CompressorException;
+import org.apache.commons.compress.compressors.CompressorStreamFactory;
+import org.apache.hc.client5.http.entity.InputStreamFactory;
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+
+/**
+ * A factory for creating InputStream instances, utilizing Apache Commons Compress.
+ * This class is compiled with Commons Compress as an optional dependency, loading
+ * only when the library is present at runtime, avoiding mandatory inclusion in
+ * downstream builds.
+ * <p>
+ * <p>
+ * Some encodings require native helper JARs; runtime availability is checked
+ * using a lightweight Class.forName probe to register codecs only when helpers
+ * are present.
+ *
+ * @since 5.6
+ */
+@Internal
+@Contract(threading = ThreadingBehavior.STATELESS)
+final class CommonsCompressDecoderFactory implements InputStreamFactory {
+
+
+    /**
+     * Map of codings that need extra JARs → the fully‐qualified class we test for
+     */
+    private static final Map<ContentCoding, String> REQUIRED_CLASS_NAME;
+
+    static {
+        final Map<ContentCoding, String> m = new EnumMap<>(ContentCoding.class);
+        m.put(ContentCoding.BROTLI, "org.brotli.dec.BrotliInputStream");
+        m.put(ContentCoding.ZSTD, "com.github.luben.zstd.ZstdInputStream");
+        m.put(ContentCoding.XZ, "org.tukaani.xz.XZInputStream");
+        m.put(ContentCoding.LZMA, "org.tukaani.xz.XZInputStream");
+        REQUIRED_CLASS_NAME = Collections.unmodifiableMap(m);
+    }
+
+    private final String encoding;
+
+    CommonsCompressDecoderFactory(final String encoding) {
+        this.encoding = encoding.toLowerCase(Locale.ROOT);
+    }
+
+    @Override
+    public String getContentEncoding() {
+        return encoding;
+    }
+
+    @Override
+    public InputStream create(final InputStream source) throws IOException {
+        try {
+            return new CompressorStreamFactory()
+                    .createCompressorInputStream(encoding, source);
+        } catch (final CompressorException | LinkageError ex) {
+            throw new IOException(
+                    "Unable to decode Content-Encoding '" + encoding + '\'', ex);
+        }
+    }
+
+
+    static boolean runtimeAvailable(final String token) {
+        final ContentCoding coding = ContentCoding.fromToken(token);
+        if (coding == null) {
+            return true;
+        }
+        final String helper = REQUIRED_CLASS_NAME.get(coding);
+        if (helper == null) {
+            // no extra JAR needed
+            return true;
+        }
+        try {
+            Class.forName(helper, false,
+                    CommonsCompressDecoderFactory.class.getClassLoader());
+            return true;
+        } catch (final ClassNotFoundException | LinkageError ex) {
+            return false;
+        }
+    }
+}
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentCoding.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentCoding.java
@@ -1,0 +1,136 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.http.entity.compress;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Locale;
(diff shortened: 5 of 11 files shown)


arturobernalg:
@ok2c @garydgregory  — could you please review this PR when you have a moment? I’ve closed the earlier one because I opted for a cleaner approach here. 
Thank you!

ok2c (commented):
@arturobernalg It is much better and can basically be merged as it, but it can be made better still.

1. `CommonsCompressDecoderFactory` uses reflection. I does not have to. It is perfectly OK to have Commons Compress as a _optional_ _compile_ dependency. what is important that there is no direct dependency on Commons Compress in `ContentDecoderRegistry`, What `#commonsCompressPresent` in `ContentDecoderRegistry` does perfectly fine and sufficient. `CommonsCompressDecoderFactory` import Commons Compress classes directly.
2. `REGISTRY` in `ContentDecoderRegistry` should be unmodifiable. It is only a matter of tweaking `#build` method a little. This  would make `#snapshot` unnecessary.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentDecoderRegistry.java
```diff
@@ -0,0 +1,147 @@
...
+            return false;
+        }
+    }
+
+
+    /**
+     * Returns a <em>defensive copy</em> of the internal registry
+     * (key = canonical encoding token, value = {@link InputStreamFactory}).
+     */
+    public static Map<String, InputStreamFactory> snapshot() {
```
garydgregory: `snapshot()` is a bad method name here IMO, rename to `getRegistry()`. 
Saying this is a `defensive copy` is misleading because `InputStreamFactory` makes no thread-safety guarantees. I would say this is a "shallow copy" which is technically correct and doesn't imply intent of attack or defense for call sites.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentDecoderRegistry.java
```diff
@@ -0,0 +1,147 @@
...
+     * Quick presence check: is the Commons factory class on the class-path?
+     */
+    private static boolean commonsCompressPresent() {
+        try {
+            Class.forName(
+                    "org.apache.commons.compress.compressors.CompressorStreamFactory",
+                    false,
+                    ContentDecoderRegistry.class.getClassLoader());
+            return true;
+        } catch (final ClassNotFoundException e) {
```
garydgregory: What about `LinkageError`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ContentCompressionExec.java
```diff
@@ -176,4 +173,16 @@ public ClassicHttpResponse execute(
         return response;
     }
 
+    private static List<String> buildDefaultAcceptEncoding() {
+
+
+        final List<String> list = new ArrayList<>(4);
+        list.add("gzip");
+        list.add("x-gzip");
```
garydgregory: Maybe a later PR could gather all these magic strings ("x-gzip" and so on) into a class (existing or new).

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentDecoderRegistry.java
```diff
@@ -0,0 +1,147 @@
...
+    private static Map<String, InputStreamFactory> build() {
+        final LinkedHashMap<String, InputStreamFactory> m = new LinkedHashMap<>();
+
+        /* 1. Built-ins (always succeed) */
+        register(m, new GZIPInputStreamFactory());
+        register(m, new DeflateInputStreamFactory());
+
+        /* 2. Optional Commons-Compress decoders (reflection, safe) */
+        if (commonsCompressPresent()) {
+            addCommons(m, "br");
```
garydgregory: See my other comment about magic strings, for a later PR.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,187 @@
...
+        final String probe = CodecProbe.probeFor(enc.toLowerCase(Locale.ROOT));
+        return probe == null || classExists(probe);
+    }
+
+    private static boolean classExists(final String cn) {
+        try {
+            Class.forName(cn, false,
+                    CommonsCompressDecoderFactory.class.getClassLoader());
+            return true;
+        } catch (final ClassNotFoundException e) {
```
garydgregory: What about `LinkageError`?

httpclient5/src/main/java/org/apache/hc/client5/http/entity/BrotliInputStreamFactory.java
```diff
@@ -41,6 +41,17 @@
 @Contract(threading = ThreadingBehavior.STATELESS)
 public class BrotliInputStreamFactory implements InputStreamFactory {
 
+    /**
+     * Canonical token for the deflate content-coding.
+     * @since 5.6
+     */
+    public static final String ENCODING = "br";
```
garydgregory, line 48: This new constant can be reused in this PR in a few places.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/DeflateInputStreamFactory.java
```diff
@@ -41,6 +41,17 @@
 @Contract(threading = ThreadingBehavior.STATELESS)
 public class DeflateInputStreamFactory implements InputStreamFactory {
 
+    /**
+     * Canonical token for the deflate content-coding.
+     * @since 5.6
+     */
+    public static final String ENCODING = "deflate";
```
garydgregory, line 48: This new constant can be reused in this PR in a few places.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/GZIPInputStreamFactory.java
```diff
@@ -42,6 +42,17 @@
 @Contract(threading = ThreadingBehavior.STATELESS)
 public class GZIPInputStreamFactory implements InputStreamFactory {
 
+    /**
+     * Canonical token for the gzip content-coding.
+     * @since 5.6
+     */
+    public static final String ENCODING = "gzip";
```
garydgregory, line 49: This new constant can be reused in this PR in a few places.

garydgregory (changes requested):
TY @arturobernalg !

I agree with @ok2c comments, especially regarding reflection. I've made a few low-level comments throughout.

arturobernalg:
Please @ok2c @garydgregory  do another pass.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentDecoderRegistry.java
garydgregory: Ping?

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            return new CompressorStreamFactory()
+                    .createCompressorInputStream(encoding, source);
+        } catch (final CompressorException | LinkageError ex) {
+            throw new IOException(
+                    "Unable to decode Content-Encoding '" + encoding + '\'', ex);
+        }
+    }
+
+    private enum Probe {
```
garydgregory: This feels a bit redundant with the new `ContentCoding` enum. What do you think about moving the "requires this type name" aspect (a.k.a. "probe class") of `Probe` to `ContentCoding` and turning this "Probe" into a `Set<ContentCoding>`?

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+        BR(ContentCoding.BROTLI.token(), "org.brotli.dec.BrotliInputStream"),
+        ZSTD(ContentCoding.ZSTD.token(), "com.github.luben.zstd.ZstdInputStream"),
+        XZ(ContentCoding.XZ.token(), "org.tukaani.xz.XZInputStream"),
+        LZMA(ContentCoding.LZMA.token(), "org.tukaani.xz.XZInputStream");
+
+        final String enc, probeClass;
+
+        Probe(final String enc, final String probeClass) {
```
garydgregory: `probeClass` is misleading to me because it's a _name_, not an actual class. I think "requiredTypeName" or "requiredClassName" would be a clearer name because "probe" doesn't really tell you anything (to me).

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+            this.enc = enc;
+            this.probeClass = probeClass;
+        }
+
+        static String helperFor(final String enc) {
```
garydgregory: I would stay away from cryptic names like "enc", even in private code, if it's an "encoding", the call it that ;-)

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.commons.compress.compressors.CompressorException;
+import org.apache.commons.compress.compressors.CompressorStreamFactory;
+import org.apache.hc.client5.http.entity.InputStreamFactory;
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+
+/**
+ * {@link InputStreamFactory} backed by
+ * <a href="https://commons.apache.org/proper/commons-compress/">Apache&nbsp;Commons Compress</a>.
```
garydgregory: If there's a non-breaking space between words 1 and 2, why not between words 2 and 3? I would just not bother since it might cause odd rendering on small screens like phones and tablets.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+ * <p>
+ * The class is compiled against Commons Compress but lives behind an <i>optional</i>
```
garydgregory: I would not use `<i>` in Javadocs, that's up to the rederer to decide what style to apply, which is why Javadoc supports `<em>` for ephasis which is usually translated to italics and `<strong>` which is usually translated to bold.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+ * dependency.  At run-time it will only be loaded when the library is present,
+ * therefore callers may rely on it without pulling Commons Compress into every
+ * downstream build.
+ * </p>
+ *
+ * <h4>Run-time guards</h4>
+ * Some encodings (e.g.&nbsp;{@code br}, {@code zstd}, {@code xz/lzma})
```
garydgregory: In general, I stay away from Latin abbreviations, I find it doesn't make the text clearer to read. No need for non-breaking space IMO.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+
+/**
+ * {@link InputStreamFactory} backed by
+ * <a href="https://commons.apache.org/proper/commons-compress/">Apache&nbsp;Commons Compress</a>.
+ * <p>
+ * The class is compiled against Commons Compress but lives behind an <i>optional</i>
+ * dependency.  At run-time it will only be loaded when the library is present,
```
garydgregory: You don't need two spaces after a `.`; we're not using typewriters anymore and Javadoc is not rendered in mono-space ;-)

garydgregory (commented):
@arturobernalg 
Look good. I
have some
comment
   scattered
```
also, why
     not use longer
     lines?
         I'm not a fan
         of the
         newspaper format
```
;-)

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/CommonsCompressDecoderFactory.java
arturobernalg: please @garydgregory  do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentCoding.java
```diff
@@ -0,0 +1,122 @@
...
+    public String token() {
+        return token;
+    }
+
+    /**
+     * Lookup the enum by token (case-insensitive).
+     */
+    public static ContentCoding fromToken(final String token) {
+        final String copyToken = token.toLowerCase(Locale.ROOT);
+        for (final ContentCoding coding : values()) {
```
ok2c: @arturobernalg This method can be optimized by creating a static map

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ContentCompressionExec.java
```diff
@@ -81,36 +80,34 @@ public final class ContentCompressionExec implements ExecChainHandler {
...
             this.decoderRegistry = decoderRegistry;
         } else {
-            final RegistryBuilder<InputStreamFactory> builder = RegistryBuilder.<InputStreamFactory>create()
-                .register("gzip", GZIPInputStreamFactory.getInstance())
-                .register("x-gzip", GZIPInputStreamFactory.getInstance())
-                .register("deflate", DeflateInputStreamFactory.getInstance());
-            if (brotliSupported) {
-                builder.register("br", BrotliInputStreamFactory.getInstance());
-            }
+            final Map<String, InputStreamFactory> decoders =
```
ok2c: @arturobernalg The content of `encodingsHeader` list should be generated based on this map. This would also make `#buildDefaultAcceptEncoding` unnecessary.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentDecoderRegistry.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Returns the unmodifiable codec map (key = canonical token, value = factory).
+     */
+    public static Map<String, InputStreamFactory> getRegistry() {
+        return REGISTRY;
+    }
+
+
+    private static Map<String, InputStreamFactory> buildRegistry() {
```
ok2c: @arturobernalg Why not using `ContentCoding` as a key instead of plain `String`?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ContentCompressionExec.java
arturobernalg: @ok2c Yes, i know, but my original thinking was that when the user doesn’t customize `encodingsHeader`, we’d fall back to the “must-support” RFC decoders only (gzip/deflate). Please do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ContentCompressionExec.java
```diff
@@ -81,36 +81,47 @@ public final class ContentCompressionExec implements ExecChainHandler {
...
-            final List<String> encodings = new ArrayList<>(4);
-            encodings.add("gzip");
-            encodings.add("x-gzip");
-            encodings.add("deflate");
-            if (brotliSupported) {
-                encodings.add("br");
-            }
-            this.acceptEncoding = MessageSupport.headerOfTokens(HttpHeaders.ACCEPT_ENCODING, encodings);
-        }
+        final List<String> encodingsHeader = acceptEncoding != null ? acceptEncoding : DEFAULT_ACCEPT_ENCODINGS;
```
ok2c, line 108: @arturobernalg This is what do not understand: if the `Accept-Cncoding` header contains the `DEFAULT_ACCEPT_ENCODINGS` values how would the opposite endpoint know about other encodings supported by the client? All those extra codes provided by Commons Compress would be useless. What am I missing?

arturobernalg: HI @ok2c I’m not quite getting the question fully. The `ContentDecoderRegistry` builds the REGISTRY on the fly—defaults like gzip and deflate are always there, but the extra goodies only hop in if Commons Compress is on the classpath and runtimeAvailable says it’s ok.

ok2c: @arturobernalg ~~Sure, but how will the **_server_** know about extra encodings supported by the client?~~

ok2c: @arturobernalg My apologies. I was not looking at the latest version of your changes. I still think it is a bit confusing how default the `Accept-Encoding` header gets generated, but I can tweak it afterwards.

arturobernalg: @ok2c Do you want me to do some change?

ok2c: @arturobernalg No, I do not. Everything look OK to me.

PR #657, opened 2025-06-21
HTTPCLIENT-2375 Add first-class request-side compression support and pluggable encoders

arturobernalg:
Adds pluggable request-side compression. You can now call `EntityBuilder.compressed(ContentCoding)` to have any codec discovered at runtime—GZIP remains the default, but `Brotli`, `Zstandard`, `XZ`, and the rest are picked up automatically when Commons Compress and the relevant helper JARs are on the class-path. The change introduces a `ContentEncoderRegistry`, removes duplicate reflection probes via the new `CommonsCompressSupport`, keeps `gzipCompressed()` as a deprecated convenience alias, and ships a tiny `Zstd` round-trip demo plus unit tests. No new mandatory dependencies; everything stays lightweight unless you choose to add more codecs.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentEncoderRegistry.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentEncoderRegistry.java
@@ -1,0 +1,95 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.http.entity.compress;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.Map;
+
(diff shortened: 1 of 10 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentEncoderRegistry.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.hc.client5.http.entity.DeflateCompressingEntity;
+import org.apache.hc.client5.http.entity.GzipCompressingEntity;
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.http.HttpEntity;
+
+@Internal
+@Contract(threading = ThreadingBehavior.STATELESS)
+public final class ContentEncoderRegistry {
```
ok2c, line 44: @arturobernalg There should provably be a single registry that can provide wrap / unwrap functions. `InputStreamFactory` should probably be deprecated. This however can be done with another change-set.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/ClientServerCompressionExample.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private ClientServerCompressionExample() {
+    }
+
+    public static void main(final String[] args) throws Exception {
+
+        /* ──────────────────────────────────────────────
+           1. Tiny echo server that understands “br|zstd…”
+           ────────────────────────────────────────────── */
+        final HttpServer server = HttpServer.create(new InetSocketAddress(0), 0);
```
ok2c: @arturobernalg Would not it be nicer to use our own `HttpServer` from core here?

garydgregory: Agreed, we have to eat out of our own kitchen here 😉

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentEncoderRegistry.java
arturobernalg: @ok2c I’ll take care of consolidating the encoder / decoder maps into a single registry (and marking `InputStreamFactory` for deprecation) in a follow-up change-set after this PR is merged.

ok2c (approved):
@arturobernalg This warning looks legit and probably should be fixed. Otherwise looks good to me.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentEncoderRegistry.java
garydgregory: That sounds like a good plan.

PR #660, opened 2025-06-25
Replace the old InputStreamFactory registries with the new Decoder / Encoder pipeline

arturobernalg:
This change swaps out the legacy compression infrastructure for the new decoder / encoder model introduced in 5.6. The codebase now routes all (de)compression through `ContentCodecRegistry`, uses functional interfaces instead of `InputStreamFactory`, and updates tests and docs accordingly while preserving existing public classes as deprecated shims. Everything builds and the full test suite passes.


httpclient5/src/main/java/org/apache/hc/client5/http/entity/BrotliDecompressingEntity.java
```diff
@@ -32,9 +32,12 @@
  * {@link org.apache.hc.core5.http.io.entity.HttpEntityWrapper} responsible for
  * handling br Content Coded responses.
  *
+ * @deprecated Use Decoder API
```
garydgregory: I think it best to use '@link' in '@deprecated' comments. It makes it much easier to follow the bouncing ball 😉

httpclient5/src/main/java/org/apache/hc/client5/http/entity/InputStreamFactory.java
```diff
@@ -31,9 +31,10 @@
 
 /**
  * Factory for decorated {@link InputStream}s.
- *
+ * @deprecated
```
garydgregory: The comment with an '@link' is missing, which happens in many places in the PR.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/Encoder.java
```diff
@@ -0,0 +1,59 @@
...
+ *   <li>Report an unknown content length ({@code -1}) and
+ *       {@link HttpEntity#isChunked() chunked} transfer.</li>
+ *   <li>Compress the source bytes as they are written to the output
+ *       stream supplied by the HTTP transport.</li>
+ * </ul>
+ *
+ * @since 5.6
+ */
+@FunctionalInterface
+public interface Encoder {
```
garydgregory: We could reuse or extend UnaryOperator instead of inventing our own.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
```diff
@@ -234,6 +238,13 @@ private ExecInterceptorEntry(
 
     private List<Closeable> closeables;
 
+    /**
+     * Custom decoders keyed by {@link ContentCoding}.
+     *
+     * @since 5.6
```
garydgregory: We don't need '@since' tags for private code.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/Encoder.java
arturobernalg: Prefer a dedicated Encoder over UnaryOperator<HttpEntity> – the name conveys intent and leaves room for HTTP-specific defaults later. Keeps symmetry with Decoder and avoids leaking java.util.function into the public API.

arturobernalg:
Added a terse Javadoc note that points users to the new Decoder/Encoder API.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/Decoder.java
```diff
@@ -0,0 +1,60 @@
...
+ * returns a lazily-decoded view of the same byte sequence.
+ *
+ * <p>Implementations <strong>must</strong> return a stream that honours the
+ * usual {@code InputStream} contract and propagates {@link IOException}s
+ * raised by the underlying transport.</p>
+ *
+ * @since 5.6
+ */
+@FunctionalInterface
+public interface Decoder {
```
ok2c: @arturobernalg Should not `Decoder` be working the same was as `Encoder` and unwrap `HttpEntity`?

arturobernalg: @ok2c `Decoder` only swaps the input stream—`DecompressingEntity` already wraps the HttpEntity—so returning an `InputStream` keeps the API minimal and avoids duplicating header-handling logic.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/ContentCodecRegistry.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        return Collections.unmodifiableMap(m);
+    }
+
+
+    /**
+     * Returns the {@link Codec} (or {@code null}).
+     */
+    public static Codec codec(final ContentCoding coding) {
```
ok2c: @arturobernalg Here we have a scope violation: a public method exposing a package private class. Either make the method package private or make the class public. It is internal anyway.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/Decoder.java
ok2c: @arturobernalg to me it is more important for the API to be logically consistent than saving a few lines of header handling code that will likely get optimized out by the compiler. In my opinion The `Encoder` / `Decoder` interfaces should work either with `InputStream` / `OutputStream` or with `HttpEntity`, but not with a mix. Feel free to leave it as is and merge the change-set but I will likely try to normalize the APIs later.

arturobernalg: @ok2c  please do another pass.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/Decoder.java
```diff
@@ -0,0 +1,60 @@
...
+     * produces the uncompressed data.
+     *
+     * @param src the source {@code HttpEntity} positioned at the start of the
+     *            encoded payload (never {@code null})
+     * @return a new, undecoded {@code HttpEntity}; the caller is responsible
+     *         for consuming / closing its content stream
+     * @throws IOException if the decoding entity cannot be created or an
+     *                     underlying I/O error occurs
+     */
+    HttpEntity wrap(HttpEntity src) throws IOException;
```
ok2c: @arturobernalg I think this method does not need to throw `IOException`. At least it does not appear to be thrown anywhere in the existing code.

I also agree with @garydgregory here. I think we could just use standard `Function` here, but this is a matter of taste so I will not object.

Looks good otherwise.

arturobernalg: HI @ok2c  please do another pass. Now I'm using java functions instead.

ok2c: > HI @ok2c please do another pass. Now I'm using java functions instead.

@arturobernalg Oh, man. I meant something different. I think `IOFunction` was fine. It makes sense to have a function that throws IOException. What I meant the `Decoder` and `Encoder` could be just `UnaryOperator` or `Function` as suggested by @garydgregory

arturobernalg:
> @arturobernalg I would prefer to have `IOFunction` restored and `Encoder` and `Decoder` replaced but this is more of a matter of taste. You can keep them if you strongly prefer it

@ok2c  i just 

> @arturobernalg I would prefer to have `IOFunction` restored and `Encoder` and `Decoder` replaced but this is more of a matter of taste. You can keep them if you strongly prefer it

HI @ok2c  i just implement the `UnaryOperator` instead. please do another pass.

arturobernalg:
If I hear no complaints or no feedback within 1 day I will merge the PR as is.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/IOFunction.java
```diff
@@ -0,0 +1,58 @@
...
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.client5.http.entity.compress;
+
+import java.io.IOException;
+
+import org.conscrypt.Internal;
```
rhernandez35: This is the wrong `@Internal` import. You want `org.apache.hc.core5.annotation.Internal`.

httpclient5/src/main/java/org/apache/hc/client5/http/entity/compress/DecompressingEntity.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Returns the cached <em>decoded</em> stream, creating it once if necessary.
+     */
+    @Override
+    public InputStream getContent() throws IOException {
+        if (!isStreaming()) {
+            return decoder.apply(super.getContent());
+        }
+
+        InputStream local = cached;
```
rhernandez35, line 62: Do we actually need DCL here? `writeTo` isn't thread-safe or idempotent, so under what scenarios do we expect `getContent()` to be called from multiple threads?

arturobernalg: DCL is a cheap guard against accidental multi-thread reuse of the same entity instance, ensuring we don’t create two competing decoder streams.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
```diff
@@ -211,6 +215,7 @@ private ExecInterceptorEntry(
     private BackoffManager backoffManager;
     private Lookup<AuthSchemeFactory> authSchemeRegistry;
     private Lookup<CookieSpecFactory> cookieSpecRegistry;
+    @Deprecated
     private LinkedHashMap<String, InputStreamFactory> contentDecoderMap;
```
ok2c, line 219: @arturobernalg I found an issue with this change-set that is easily fixable. 

At the moment the `contentDecoderMap` variable is essentially write-only. There is a non-deprecated setter for it but it is not being read anywhere. That is wrong.

What needs to be done is this: `contentDecoderMap` must be removed, the `#setContentDecoderRegistry` setter must be changed to populate `contentDecoder` through an adaptor (see `DefaultHttpClientConnectionOperator` for an example ) and the `#setContentDecoderRegistry` must be deprecated.

`contentDecoder` also sounds like a rather bad choice of a name for what is essentially a map of objects

arturobernalg: On it

arturobernalg: > @arturobernalg I found an issue with this change-set that is easily fixable.
> 
> At the moment the `contentDecoderMap` variable is essentially write-only. There is a non-deprecated setter for it but it is not being read anywhere. That is wrong.
> 
> What needs to be done is this: `contentDecoderMap` must be removed, the `#setContentDecoderRegistry` setter must be changed to populate `contentDecoder` through an adaptor (see `DefaultHttpClientConnectionOperator` for an example ) and the `#setContentDecoderRegistry` must be deprecated.
> 
> `contentDecoder` also sounds like a rather bad choice of a name for what is essentially a map of objects

@ok2c  please take a look [fd30e92](https://github.com/apache/httpcomponents-client/pull/675/commits/fd30e924982a61dcde244b812f3561cdd23563d8)

PR #662, opened 2025-06-28
Enable JEP 380 UDS support for sync client

rschmitt:
This change enables the sync client to use standard library Unix-domain sockets on Java 16+ without a JUnixSocket dependency, as is already supported by the async client.

JEP 380 only exposes Unix-domain socket support through the `SocketChannel` API, but the sync client is tightly coupled to the legacy `java.net.Socket` API. The idea behind this change is to wrap the UDS `SocketChannel` in a subclass of `java.net.Socket` that acts as a translation layer between the legacy socket API and `SocketChannel`. In particular, the adapter adds support for socket timeouts by translating blocking reads to non-blocking reads and supplying the socket timeout value to the `select()` call.

After this change, HttpClient will use the Java standard library by default. JUnixSocket will only be loaded on older versions of Java that lack standard library support for UDS, and even then only for the synchronous client, as the JUnixSocket-provided `SocketChannel` cannot be used with the JDK-provided `Selector` used by `IOReactor`.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/socket/Jep380SocketChannelAdapter.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/socket/Jep380SocketChannelAdapter.java
@@ -1,0 +1,95 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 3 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/socket/Jep380SocketChannelAdapter.java
```diff
@@ -0,0 +1,95 @@
...
+        return !channel.isOpen();
+    }
+
+    @Override
+    public synchronized void close() throws IOException {
+        adapter.close();
+    }
+
+    @Override
+    public synchronized int getSoTimeout() throws SocketException {
```
arturobernalg: Please use `ReentrantLock` so the code stays compatible with Java 21 virtual threads.

arturobernalg (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/socket/Jep380SocketChannelAdapter.java
ok2c: @rschmitt I agree. We are no longer supposed to use `synchronized` anywhere in our code.

rschmitt: Good point. Since `synchronized` is not actually part of the method signature, I think I'd rather just drop it altogether.

rschmitt: I've dropped `synchronized` since the underlying field is declared `volatile`.

ok2c (approved):

PR #663, opened 2025-07-02
HTTPCLIENT-2379 - Handle duplicate release() safely in H2SharingConnPool

arturobernalg:
Return early in H2SharingConnPool#release when the entry’s lease count is zero, so duplicate releases become no-ops and the underlying pool no longer throws `IllegalStateException`.

tested with https://github.com/yhzdys/httpcomponents-client/commit/c8256b33c3adb45cfa2a49efb2dacfb611da9ac1

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/H2SharingConnPool.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/H2SharingConnPool.java
@@ -270,16 +270,17 @@ public class H2SharingConnPool<T, C extends HttpConnection> implements ManagedCo
         PoolEntry<T, C> lease() {
             lock.lock();
             try {
-                final PoolEntry<T, C> entry = entryMap.entrySet().stream()
+                return entryMap.entrySet().stream()
+                        .filter(e -> {
+                            final C conn = e.getKey().getConnection();
+                            return conn != null && conn.isOpen();
+                        })
                         .min(Comparator.comparingLong(e -> e.getValue().get()))
-                        .map(Map.Entry::getKey)
+                        .map(e -> {
+                            e.getValue().incrementAndGet();
+                            return e.getKey();
+                        })
                         .orElse(null);
-                if (entry == null) {
-                    return null;
-                }
-                final AtomicLong counter = getCounter(entry);
-                counter.incrementAndGet();
-                return entry;
             } finally {
                 lock.unlock();
             }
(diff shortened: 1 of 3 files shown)


ok2c:
@arturobernalg I am afraid this change-set fixes a side-effect and simply hides the underlying problem. As far as I remember (i am still reviewing my code) there should be no un-tracked pool entries for tracked routes.

arturobernalg:
> I am afraid this change-set fixes a side-effect and simply hides the underlying problem. As far as I remember (i am still reviewing my code) there should be no un-tracked pool entries for tracked routes.

@ok2c the entry is dropped after the first release, and the second call tries to hand it back again, triggering the exception... i see. ..
That’s was  my initial intent [2099b31](https://github.com/apache/httpcomponents-client/pull/663/commits/2099b31edfc74ed99c65872465d820d53b098ec1) to fix this .

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/H2SharingConnPool.java
```diff
@@ -288,20 +290,14 @@ PoolEntry<T, C> lease() {
         long release(final PoolEntry<T, C> entry, final boolean reusable) {
             lock.lock();
             try {
```
ok2c, line 291: @arturobernalg Could you please add this check here?
```
if (!reusable) {
     entry.discardConnection(CloseMode.GRACEFUL);
}
```

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/H2SharingConnPool.java
arturobernalg: @ok2c  done

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/H2SharingConnPool.java
```diff
@@ -270,16 +270,18 @@ long track(final PoolEntry<T, C> entry) {
...
-                final PoolEntry<T, C> entry = entryMap.entrySet().stream()
+                return entryMap.entrySet().stream()
+                        .filter(e -> {
+                            final C conn = e.getKey().getConnection();
+                            return conn != null && conn.isOpen();
+                        })
                         .min(Comparator.comparingLong(e -> e.getValue().get()))
                         .map(Map.Entry::getKey)
+                        .map(e -> {
```
yhzdys, line 279: @arturobernalg Perhaps this is better?
~~~java
return entryMap.entrySet().stream()
        .filter(e -> {
            final C conn = e.getKey().getConnection();
            return conn != null && conn.isOpen();
        })
        .min(Comparator.comparingLong(e -> e.getValue().get()))
        .map(e -> {
            e.getValue().incrementAndGet();
            return e.getKey();
        })
        .orElse(null);
~~~

arturobernalg: @yhzdys idd. I just made the change. please check

arturobernalg:
@ok2c, it looks like the `TestValidateAfterInactivity` test is stuck in an infinite loop.

ok2c:
> @ok2c, it looks like the `TestValidateAfterInactivity` test is stuck in an infinite loop.

@arturobernalg It does appear that way. CC @rschmitt

arturobernalg:
> > @ok2c, it looks like the `TestValidateAfterInactivity` test is stuck in an infinite loop.
> 
> @arturobernalg It does appear that way. CC @rschmitt

@rschmitt you can find here another example https://github.com/apache/httpcomponents-client/actions/runs/16097757606

arturobernalg:
@ok2c  cherry pick to `5.5.x` ?

ok2c:
> @ok2c cherry pick to `5.5.x` ?

@arturobernalg Yes, please.

arturobernalg:
> > @ok2c cherry pick to `5.5.x` ?
> 
> @arturobernalg Yes, please.

Done

rschmitt:
Taking a look, I'll try to get a thread dump

rschmitt:
The fix is trivial, I'll have it out in a second

rschmitt:
https://github.com/apache/httpcomponents-client/pull/673

PR #666, opened 2025-07-03
HTTPCLIENT-2380 – make DefaultAsyncClientConnectionOperator extensible

arturobernalg:
This PR fixes [HTTPCLIENT-2380](https://issues.apache.org/jira/browse/HTTPASYNC-173) by changing the constructor of `DefaultAsyncClientConnectionOperator` to protected, allowing external subclasses.


httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/DefaultAsyncClientConnectionOperator.java
```diff
@@ -71,7 +71,17 @@ public class DefaultAsyncClientConnectionOperator implements AsyncClientConnecti
...
-    DefaultAsyncClientConnectionOperator(
+    /**
+     * Constructs a new {@code DefaultAsyncClientConnectionOperator}.
+     *
+     * <p><strong>Note:</strong> this class is marked {@code @Internal}; rely on it
+     * only if you are prepared for incompatible changes in a future major
+     * release.  Typical client code should use the high-level builders in
+     * {@code HttpAsyncClients} instead.</p>
+     *
+     * @since 5.6
```
ok2c: @arturobernalg This change should also be cherry-picked to `5.5.x` so I would just omit the version here or change it to `@since 5.5`. It is non-public class. We can change it in the stable branch as well.

arturobernalg:
@ok2c  cherry picket to `5.5.x`  [7d87256](https://github.com/apache/httpcomponents-client/commit/7d872566c892a66234861748c89d424a0c0e43a0)

ok2c:
@arturobernalg The commit message is bad. The issue is wrong and there is version 5.6 in the message. Please correct the message and force-push the commit.

PR #667, opened 2025-07-04
Fix `validateAfterInactivity` on the async client

rschmitt:
`validateAfterInactivity` is primarily a synchronous client feature, since it is necessary to attempt to read from the underlying socket in order to discover that the connection has been closed by the remote peer. The async client doesn't have this issue; the IOReactor is automatically notified when connections are closed and removes them from the pool (see `SingleCoreIOReactor::processClosedSessions`).

For some reason, however, the async client's connection manager treats `validateAfterInactivity` as a kind of connection TTL: if the connection has been idle longer than the `validateAfterInactivity` config value, then the connection is closed immediately unless it is an HTTP/2 connection, in which case a PING frame is used to check for liveness. Since stale connections are removed from the pool automatically as mentioned above, this change simply removes this connection close behavior.

A set of integration tests has been added to cover both clients.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
@@ -315,10 +315,6 @@ public class PoolingAsyncClientConnectionManager implements AsyncClientConnectio
                                             })), Command.Priority.IMMEDIATE);
                                             return;
                                         }
-                                        if (LOG.isDebugEnabled()) {
-                                            LOG.debug("{} connection {} is closed", id, ConnPoolSupport.getId(connection));
-                                        }
-                                        poolEntry.discardConnection(CloseMode.IMMEDIATE);
                                     }
                                 }
                             }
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
```diff
@@ -315,10 +315,6 @@ public void completed(final PoolEntry<HttpRoute, ManagedAsyncClientConnection> p
                                             })), Command.Priority.IMMEDIATE);
                                             return;
```
ok2c, line 316: @rschmitt I guess `return` here can also be removed. Looks good otherwise.

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
rschmitt: Won't that cause `leaseCompleted(poolEntry)` to be called immediately (line 321) after the PingCommand is submitted, and then a second time after the PingCommand finishes?

ok2c: @rschmitt My bad. Disregard.

PR #672, opened 2025-07-06
HTTPCLIENT-2141: revision of the original implementation

ok2c:
I did not do a very job with the initial implementation of HTTPCLIENT-2141. The fix was misplaced, did not apply to the async transport and the behavior was impossible to customize using a custom `HttpRequestRetryStrategy `

The proposed change-set revises the original fix and corrects its deficiencies.

It also adds missing integration rests for the classic transport and makes debug logging consistent between the classic and async request execution pipelines

This is a prerequisite to solving https://issues.apache.org/jira/browse/HTTPCLIENT-2371


ok2c:
@arturobernalg Could you please take a look?

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -517,6 +517,12 @@ public Builder setConnectTimeout(final long connectTimeout, final TimeUnit timeU
          * HTTP transports with message multiplexing.
          * </p>
          * <p>
+         * Please note that response timeout is not a deadline. Its absolute value
+         * can be exceeded, for example, in case of automatic request re-execution.
+         * Please make sure the automatic request re-execution plociy has been
```
arturobernalg: Typo. should be `policy`

arturobernalg:
@ok2c 

In the `retryRequest(HttpResponse, int, HttpContext)` method, the code checks `defaultRetryInterval` against the `responseTimeout` to prevent retries that would exceed the timeout. However, since `getRetryInterval` can return a dynamic interval based on `the Retry-After` header (as noted in the Javadoc: “the default retry interval between subsequent retries if the Retry-After header is not set or invalid”), it’s possible for the actual retry delay to exceed `responseTimeout` if Retry-After specifies a larger value. For example, if `defaultRetryInterval` is 1 second, `responseTimeout` is 3 seconds, and `Retry-After` is 5 seconds, the check allows a retry, but the 5-second delay violates the timeout. 
I wonder if it would make sense to call getRetryInterval in retryRequest to compare the actual retry interval against responseTimeout, ???  That make sense?? 


```
if (context != null) {
    final HttpClientContext clientContext = HttpClientContext.cast(context);
    final RequestConfig requestConfig = clientContext.getRequestConfigOrDefault();
    final Timeout responseTimeout = requestConfig.getResponseTimeout();
    if (responseTimeout != null) {
        final TimeValue retryInterval = getRetryInterval(response, execCount, context);
        if (TimeValue.isPositive(retryInterval) && retryInterval.compareTo(responseTimeout) > 0) {
            return false;
        }
    }
}

```

ok2c:
@arturobernalg I think `Retry-After` specified by the opposite endpoint should take precedence over the local settings. I also hate calling the method, parsing the header, throwing the results away and then doing the same thing again. In this case I would rather deprecate `HttpRequestRetryStrategy` and replace it with a better interface in 5.6.

Another alternative would be to not retry requests that have a response timeout at all.

arturobernalg:
> @arturobernalg I think `Retry-After` specified by the opposite endpoint should take precedence over the local settings. I also hate calling the method, parsing the header, throwing the results away and then doing the same thing again. In this case I would rather deprecate `HttpRequestRetryStrategy` and replace it with a better interface in 5.6.
> 
> Another alternative would be to not retry requests that have a response timeout at all.

@ok2c , I agree Retry-After should drive the delay. Let's tackle the timeout check in a separate PR to avoid parsing overhead. Thanks for the insight!

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
ok2c: @arturobernalg Corrected.

PR #673, opened 2025-07-06
TestValidateAfterInactivity: Set maxConns to 1

rschmitt:
The test server only supports a single connection at a time, and the async client can apparently attempt to establish a second connection, possibly due to a benign race condition in how leased connections are returned to the pool.


ok2c:
@rschmitt For changes like that there is no need to raise a PR for fixes like that. They can be reviewed and backed out in the worst case post-commit. Things that generally require a review are non-functional style changes, code "beautification", those drag in new dependencies and so on.

PR #674, opened 2025-07-13
HTTPCLIENT-2371: Logging of request re-execution at INFO priority

ok2c:
This change-set should fix [HTTPCLIENT-2371](https://issues.apache.org/jira/browse/HTTPCLIENT-2371) by logging request re-execution at INFO priority in case of an I/O error, status 503 or status 429. It also makes logging consistent between classic and async implementations


ok2c:
@arturobernalg Could you please double-check?

PR #675, opened 2025-07-15
Deprecate setContentDecoderRegistry to populate contentDecoder

arturobernalg:
Adapt `InputStreamFactory` map into `contentDecoder` via `DecompressingEntity` adapters

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
@@ -28,6 +28,7 @@
 package org.apache.hc.client5.http.impl.classic;
 
 import java.io.Closeable;
+import java.io.InputStream;
 import java.net.ProxySelector;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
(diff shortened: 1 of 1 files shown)


ok2c:
@arturobernalg I spotted another problem. The `LinkedHashMap<ContentCoding, UnaryOperator<HttpEntity>>` map used by `HttpClientBuilder` basically means one cannot plug-in 3rd party coders if they are not in present in `ContentCoding`. This is not good. The map should be  `LinkedHashMap<String, UnaryOperator<HttpEntity>>`. This needs to be corrected first.

arturobernalg:
> @arturobernalg I spotted another problem. The `LinkedHashMap<ContentCoding, UnaryOperator<HttpEntity>>` map used by `HttpClientBuilder` basically means one cannot plug-in 3rd party coders if they are not in present in `ContentCoding`. This is not good. The map should be `LinkedHashMap<String, UnaryOperator<HttpEntity>>`. This needs to be corrected first.

@ok2c  I just made the change

ok2c:
@arturobernalg I would like to propose an alternative approach. Could you live with #678?

ok2c:
Superseded by #678

PR #677, opened 2025-07-15
[HTTPCLIENT-2383] Make LaxRedirectStrategy to allow redirects if sensitive headers are part of the request

dani0600:


httpclient5/src/test/java/org/apache/hc/client5/http/impl/TestDefaultRedirectStrategy.java
```diff
@@ -261,13 +261,6 @@ void testRedirectAllowed() throws Exception {
                 BasicRequestBuilder.get("/").build(),
                 null));
 
-        Assertions.assertTrue(redirectStrategy.isRedirectAllowed(
```
ok2c, line 264: @dani0600 Was a reason for removing this test?

dani0600: From what I see, it’s a duplicated test with the one above :)

ok2c: @dani0600 Ah. I see.

ok2c (approved):

PR #680, opened 2025-07-17
HTTPCLIENT-2381 Enable automatic mapping 

arturobernalg:
Enable env-proxy support ([HTTPCLIENT-2381](Enable env-proxy support (HTTPCLIENT-2381) – add EnvironmentProxyConfigurer, wire it into HttpClientBuilder, revert extra fallbacks, and introduce JUnit test (System-Lambda) with dep-management cleanup.)) – add `EnvironmentProxyConfigurer`, wire it into `HttpClientBuilder`, revert extra fallbacks, and introduce JUnit test (System-Lambda) with dep-management cleanup.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
@@ -233,6 +233,7 @@ public class HttpClientBuilder {
     private boolean authCachingDisabled;
     private boolean connectionStateDisabled;
     private boolean defaultUserAgentDisabled;
+    private boolean proxyAutodetectionDisabled;
     private ProxySelector proxySelector;
 
     private List<Closeable> closeables;
@@ -792,6 +793,19 @@ public class HttpClientBuilder {
     }
 
     /**
+     * Disables automatic proxy detection for clients created by this builder.
+     * <p>
+     * When disabled, and unless an explicit proxy or route planner is configured,
+     * the builder falls back to {@link org.apache.hc.client5.http.impl.routing.DefaultRoutePlanner}.
+     * </p>
+     * @return this instance.
+     */
+    public final HttpClientBuilder disableProxyAutodetection() {
+        this.proxyAutodetectionDisabled = true;
+        return this;
+    }
+
+    /**
      * Sets the {@link ProxySelector} that will be used to select the proxies
      * to be used for establishing HTTP connections. If a non-null proxy selector is set,
      * it will take precedence over the proxy settings configured in the client.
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
```diff
@@ -838,6 +839,11 @@ protected Function<HttpContext, HttpClientContext> contextAdaptor() {
     }
 
     public CloseableHttpClient build() {
+
```
ok2c, line 855: @arturobernalg I am a bit unsure HttpClient should be meddling with system variables behind the scene. This can really have unexpected side-effects. Could we document `EnvironmentProxyConfigurer` in javadocs and advise the users to explicitly call `EnvironmentProxyConfigurer#apply` in order to carry over proxy config from the runtime environment?

arturobernalg: @ok2c  Oleg, one last thought before I push: would you prefer we gate the transfer behind a dedicated flag, e.g. `hc5_apply_proxy_env=true` so users who really want the mapping can simply set it and skip the
explicit `EnvironmentProxyConfigurer.apply()` call?
Default remains “do nothing unless asked”

ok2c: @arturobernalg I do not have a use case, so I can make this call. _Personally_ I would prefer just an explicit `EnvironmentProxyConfigurer#apply` call, but I leave it up to you.

arturobernalg: @ok2c  please do anther pass

ok2c: @arturobernalg I am not sure I see what has changed. `EnvironmentProxyConfigurer#apply` call is still there instead of “do nothing unless asked”.

arturobernalg: please @ok2c  do another pass

ok2c (commented):
@arturobernalg You should also add a similar option to `HttpAsyncClientBuilder`.

Please also invite the original reporter to take a look at the proposed fix. From I understand (but I can well be wrong) the reporter actually wants us to read system properties by default instead of opt-in. I am not sure this change-set is going to be enough.

ok2c:
@arturobernalg Actually I think the fix is misplaced. This logic should go into `SystemDefaultRoutePlanner` and our code must not change the system properties _at all_ but populate `HttpRoute`s instead based on environment

httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.java
```diff
@@ -237,6 +237,9 @@ private ExecInterceptorEntry(
 
     private List<Closeable> closeables;
 
+    private boolean applyEnvProxy;
```
ok2c: @arturobernalg This should no longer be necessary, should it?

arturobernalg: > @arturobernalg This should no longer be necessary, should it?

done

PR #681, opened 2025-07-18
HTTPCLIENT-1822: async transparent content decompression

arturobernalg:
This patch fulfils HTTPCLIENT-1822 by delivering transparent (de)compression support in the async pipeline.

DecompressingStringAsyncEntityConsumer – decodes gzip/deflate/brotli/zstd … responses using the codecs in `ContentCodecRegistry` and returns the plain body as a String.

CompressingAsyncEntityProducer – wraps any outgoing producer and re-encodes its payload with the chosen Content-Encoding, also via ContentCodecRegistry; enables gzip (or other) uploads.

 
AsyncClientDecompressionExample` shows response-side usage.
AsyncClientCompressionExample` demonstrates gzip uploads.

With these additions an async client can both receive and send compressed entities without extra boiler-plate.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DeflatingAsyncEntityProducer.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DeflatingAsyncEntityProducer.java
@@ -1,0 +1,239 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.async.methods;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.zip.Deflater;
+
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.nio.AsyncEntityProducer;
+import org.apache.hc.core5.http.nio.DataStreamChannel;
+import org.apache.hc.core5.util.Args;
+
+/**
+ * {@code AsyncEntityProducer} that streams the output of another producer
+ * through the raw DEFLATE compression algorithm.
+ *
+ * <p>The delegate’s bytes are read in small chunks, compressed with
+ * {@link java.util.zip.Deflater} and written immediately to the HTTP I/O
+ * layer.  Memory use is therefore bounded even for very large request
+ * entities.</p>
+ *
+ * @since 5.6
+ */
+public final class DeflatingAsyncEntityProducer implements AsyncEntityProducer {
+
+    /**
+     * inbound copy‐buffer
+     */
+    private static final int IN_BUF = 8 * 1024;
+    /**
+     * outbound staging buffer
+     */
+    private static final int OUT_BUF = 8 * 1024;
+
+    private final AsyncEntityProducer delegate;
+    private final String contentType;
+    private final Deflater deflater = new Deflater(Deflater.DEFAULT_COMPRESSION, /*nowrap=*/true);
+
+    /**
+     * holds compressed bytes not yet sent downstream
+     */
+    private final ByteBuffer pending = ByteBuffer.allocate(OUT_BUF);
+    private final byte[] in = new byte[IN_BUF];
+
+    private final AtomicBoolean delegateEnded = new AtomicBoolean(false);
+    private boolean finished = false;
+
+    public DeflatingAsyncEntityProducer(final AsyncEntityProducer delegate) {
+        this.delegate = Args.notNull(delegate, "delegate");
+        this.contentType = delegate.getContentType();
+        /* place pending into “read-mode” with no data */
+        pending.flip();
+    }
+
+    // ------------------------------------------------------------------ metadata
+
+    @Override
+    public boolean isRepeatable() {
+        return false;
+    }
+
+    @Override
+    public long getContentLength() {
+        return -1;
+    }     // unknown
+
+    @Override
+    public String getContentType() {
+        return contentType;
+    }
+
+    @Override
+    public String getContentEncoding() {
+        return "deflate";
+    }
+
+    @Override
+    public boolean isChunked() {
+        return true;
+    }
+
+    @Override
+    public Set<String> getTrailerNames() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public int available() {
+        if (pending.hasRemaining()) {
+            return pending.remaining();
+        }
+        return delegate.available();
+    }
+
+    // ------------------------------------------------------------------ core
+
+    @Override
+    public void produce(final DataStreamChannel channel) throws IOException {
+        /* 1 — flush any leftover compressed bytes first */
+        if (flushPending(channel)) {
+            return;  // back-pressure: outer channel could not accept more
+        }
+
+        /* 2 — pull more data from delegate */
+        delegate.produce(new InnerChannel(channel));
+
+        /* 3 — if delegate ended, finish the deflater */
+        if (delegateEnded.get() && !finished) {
+            deflater.finish();
+            deflateToPending();
+            flushPending(channel);
+            if (!pending.hasRemaining()) {
+                finished = true;
+                channel.endStream();
+            }
+        }
+    }
+
+    /**
+     * copy as much as possible from {@link #pending} to the wire
+     */
+    private boolean flushPending(final DataStreamChannel ch) throws IOException {
+        while (pending.hasRemaining()) {
+            final int written = ch.write(pending);
+            if (written == 0) {
+                return true; // back-pressure
+            }
+        }
+        pending.clear().flip();              // no data left → empty read-mode
+        return false;
+    }
+
+    /**
+     * drain {@link #deflater} into {@link #pending}
+     */
+    private void deflateToPending() {
+        /* switch pending to write-mode */
+        pending.compact();
+        final byte[] out = pending.array();
+        int total;
+        do {
+            total = deflater.deflate(out, pending.position(), pending.remaining(),
+                    Deflater.NO_FLUSH);
+            pending.position(pending.position() + total);
+            if (!pending.hasRemaining() && total > 0) {
+                /* buffer full: grow to the next power of two */
+                final ByteBuffer bigger = ByteBuffer.allocate(pending.capacity() * 2);
+                pending.flip();
+                bigger.put(pending);
+                pending.clear();
+                pending.put(bigger);
+            }
+        } while (total > 0);
+        pending.flip(); // back to read-mode
+    }
+
+    // ------------------------------------------------------------------ inner channel that receives raw bytes
+
+    private final class InnerChannel implements DataStreamChannel {
+        private final DataStreamChannel outer;
+
+        InnerChannel(final DataStreamChannel outer) {
+            this.outer = outer;
+        }
+
+        @Override
+        public void requestOutput() {
+            outer.requestOutput();
+        }
+
+        @Override
+        public int write(final ByteBuffer src) throws IOException {
+            int consumed = 0;
+            while (src.hasRemaining()) {
+                final int chunk = Math.min(src.remaining(), in.length);
+                src.get(in, 0, chunk);
+                deflater.setInput(in, 0, chunk);
+                consumed += chunk;
+                deflateToPending();
+                if (flushPending(outer)) {    // honour back-pressure
+                    break;
+                }
+            }
+            return consumed;
+        }
+
+        @Override
+        public void endStream() {
+            delegateEnded.set(true);
+        }
+
+        @Override
+        public void endStream(final List<? extends Header> trailers) {
+            endStream();
+        }
+    }
+
+    // ------------------------------------------------------------------ error / cleanup
+
+    @Override
+    public void failed(final Exception cause) {
+        delegate.failed(cause);
+    }
+
+    @Override
+    public void releaseResources() {
+        delegate.releaseResources();
+        deflater.end();
+    }
+}
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
@@ -1,0 +1,204 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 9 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DecompressingAsyncEntityConsumer.java
```diff
@@ -0,0 +1,102 @@
...
+ * codecs registered in {@link ContentCodecRegistry}.
+ *
+ * <p>
+ * This class buffers the entire response entity content into memory before
+ * applying decompression. Not suitable for large payloads.
+ * </p>
+ *
+ * @since 5.6
+ */
+public class DecompressingAsyncEntityConsumer<T> implements AsyncEntityConsumer<T> {
```
ok2c: @arturobernalg This class looks completely redundant. I pulled your change-set, removed the class, ran the tests and the example and they all worked for me.

ok2c:
@arturobernalg I am not sure about this one. I would prefer a proper solution (even if it is GZIP only) without full content buffering or nothing at all.

arturobernalg:
> without full content buffering or nothing at all

@ok2c I've implemented streaming DEFLATE and GZIP compression/decompression without full content buffering, . I'm not sure if this fully aligns with what you requested.

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/CompressingAsyncEntityProducer.java
```diff
@@ -0,0 +1,181 @@
...
+        ch.write(ByteBuffer.wrap(compressed, pos, chunk));
+        cursor.addAndGet(chunk);
+    }
+
+    private static HttpEntity encode(final HttpEntity src, final String token) throws IOException {
+        final ContentCoding coding = ContentCoding.fromToken(token);
+        if (coding == null) {
+            throw new IOException("Unknown coding: " + token);
+        }
+        final java.util.function.UnaryOperator<HttpEntity> op = ContentCodecRegistry.encoder(coding);
```
garydgregory: This shouldn't need to be a fully qualified name.

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/GzipDecompressingStringAsyncEntityConsumer.java
```diff
@@ -0,0 +1,334 @@
...
+
+    private enum GzipState {
+        HEADER_MAGIC1, HEADER_MAGIC2, HEADER_METHOD, HEADER_FLAGS, HEADER_MTIME1, HEADER_MTIME2, HEADER_MTIME3, HEADER_MTIME4,
+        HEADER_XFL, HEADER_OS, HEADER_EXTRA_LEN_LO, HEADER_EXTRA_LEN_HI, HEADER_EXTRA, HEADER_NAME, HEADER_COMMENT,
+        HEADER_HCRC_LO, HEADER_HCRC_HI, DATA, TRAILER_CRC1, TRAILER_CRC2, TRAILER_CRC3, TRAILER_CRC4,
+        TRAILER_ISIZE1, TRAILER_ISIZE2, TRAILER_ISIZE3, TRAILER_ISIZE4, DONE, ERROR
+    }
+
+    private GzipState state = GzipState.HEADER_MAGIC1;
+    private int flags = 0;
```
garydgregory: I don't think we should initialize instance variables to default values (like in C!), it feels like clutter.

ok2c:
@arturobernalg I am not in a position to tell you what to do. At the moment I see only `DeflateCompressingAsyncEntityProducer` as promising. 

Why does `DeflateDecompressingStringAsyncEntityConsumer` work with `String`s only? Could not it work with any arbitrary `EntityConsumer`? 

I would focus on these two classes for now and drop the rest.

arturobernalg:
> @arturobernalg I am not in a position to tell you what to do. At the moment I see only `DeflateCompressingAsyncEntityProducer` as promising.
> 
> Why does `DeflateDecompressingStringAsyncEntityConsumer` work with `String`s only? Could not it work with any arbitrary `EntityConsumer`?
> 
> I would focus on these two classes for now and drop the rest.

@ok2c  please do another pass. I'll implement the gzip in another PR>

ok2c:
@arturobernalg I guess I failed to properly express myself. I actually thought you were going to build `DeflatingAsyncEntityProducer` and `InflatingAsyncEntityCunsumer` using `Deflater` / `Inflater` API directly without  any of `InputStream` / `OutputStream` code that basically does not work well with async transport. I can still see `HttpEntity` / `InputStream` / `OutputStream` all over the place. I cannot stop you but i do not find such implementations very useful.

arturobernalg:
> build `DeflatingAsyncEntityProducer` and `InflatingAsyncEntityCunsumer` using `Deflater` / `Inflater` API directly

@ok2c implemented.

ok2c (commented):
@arturobernalg `DeflatingAsyncEntityProducer` and `InflatingAsyncEntityCunsumer` look good. 

I would drop `CompressingAsyncEntityProducer`. 

Could you also add an async version of `ContentCompressionExec` to the request execution pipeline?

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientDeflateCompressionExample.java
```diff
@@ -0,0 +1,106 @@
...
+
+            final String json = "{\"msg\":\"hello deflated world\"}";
+            final AsyncEntityProducer raw =
+                    new StringAsyncEntityProducer(json, ContentType.APPLICATION_JSON);
+
+            final AsyncEntityProducer deflated = new DeflatingAsyncEntityProducer(raw);
+
+            final SimpleHttpRequest request = SimpleRequestBuilder
+                    .post("https://httpbin.org/post")
+                    .addHeader("Content-Encoding", "deflate")
```
ok2c: @arturobernalg [RequestContent](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestContent.java#L136) request interceptor should do that automatically. This should not be needed. Please double-check.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientDeflatePostExample.java
```diff
@@ -0,0 +1,111 @@
...
+            /* ------------- original payload ------------------ */
+            final String json = "{\"msg\":\"hello compressed world\"}";
+            final AsyncEntityProducer raw = new StringAsyncEntityProducer(json, ContentType.APPLICATION_JSON);
+
+            /* ------------- wrap with DEFLATE ------------------ */
+            final AsyncEntityProducer deflated = new DeflatingAsyncEntityProducer(raw);
+
+            /* ------------- build request ---------------------- */
+            final SimpleHttpRequest req = SimpleRequestBuilder.post("https://httpbin.org/post")
+                    .addHeader("Content-Encoding", "deflate")
```
ok2c: @arturobernalg Same.

ok2c (commented):
@arturobernalg Very good otherwise.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientDeflateCompressionExample.java
arturobernalg: @ok2c  you are right. Is not need it.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -0,0 +1,212 @@
...
+                                                    final EntityDetails details)
+                    throws HttpException, IOException {
+
+                if (!enabled) {
+                    return cb.handleResponse(rsp, details);
+                }
+
+                final String raw = details != null && details.getContentEncoding() != null
+                        ? details.getContentEncoding()
+                        : Optional.ofNullable(rsp.getFirstHeader(HttpHeaders.CONTENT_ENCODING))
```
ok2c: @arturobernalg This is unnecessary and can potentially produce unexpected results. The interceptor should rely on the properties of `EntityDetails` only.

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DeflatingAsyncEntityProducer.java
```diff
@@ -0,0 +1,240 @@
...
+     */
+    private final ByteBuffer pending = ByteBuffer.allocate(OUT_BUF);
+    private final byte[] in = new byte[IN_BUF];
+
+    private final AtomicBoolean delegateEnded = new AtomicBoolean(false);
+    private boolean finished = false;
+
+    public DeflatingAsyncEntityProducer(final AsyncEntityProducer delegate) {
+        this.delegate = Args.notNull(delegate, "delegate");
+        this.contentType = ContentType.parse(delegate.getContentType());
```
ok2c: @arturobernalg What is the point of parsing the content-type just to have to convert it back to String again?

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientDeflateCompressionExample.java
```diff
@@ -0,0 +1,105 @@
...
+ * Demonstrates how to POST a JSON body compressed on-the-fly with
+ * {@link DeflatingAsyncEntityProducer}.  The program talks to
+ * <a href="https://httpbin.org/post">httpbin</a>, which echoes your request,
+ * making it easy to verify that the {@code Content-Encoding: deflate}
+ * header was honoured.
+ *
+ * <p>Key take-aways:</p>
+ * <ol>
+ *   <li>Wrap any existing {@code AsyncEntityProducer} in the compressor.</li>
+ *   <li>Add the encoding header yourself – HttpClient does not do that for you.</li>
```
ok2c: @arturobernalg The javadocs should be updated

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientInflateDecompressionExample.java
```diff
@@ -0,0 +1,94 @@
...
+public class AsyncClientInflateDecompressionExample {
+
+    public static void main(final String[] args) throws Exception {
+        try (final CloseableHttpAsyncClient client = HttpAsyncClients.createDefault()) {
+            client.start();
+
+            final SimpleHttpRequest request =
+                    SimpleRequestBuilder.get("https://httpbin.org/deflate").build();
+
+            final InflatingAsyncEntityConsumer<String> inflating =
```
ok2c: @arturobernalg This should now be done automatically by the exec interceptor, should it not?

arturobernalg: @ok2c  Shouldn't expect a `AsyncEntityConsumer<?>` ???? Sorry I'm not understands well the issue

ok2c: @arturobernalg Decompression / inflation should be completely transparent to the user, should it not? The purpose of`ContentCompressionAsyncExec` that you built is precisely to make the `InflatingAsyncEntityConsumer` wrapping automatically.

The problem actually is now the actual entity consumer will get wrapped with `InflatingAsyncEntityConsumer` twice.

arturobernalg: @ok2c I initially wrapped at the `AsyncEntityConsumer` level, but that broke for plain `AsyncDataConsumers` and led to double-wrapping issues. Moving the decoder to operate on `AsyncDataConsumer` fixes both problems, works for every downstream type, and keeps the implementation fully streaming.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -0,0 +1,171 @@
...
+                }
+
+                /* header lookup – unchanged from previous version */
+                String raw = details != null ? details.getContentEncoding() : null;
+                if (raw == null) {
+                    final Header h = rsp.getFirstHeader(HttpHeaders.CONTENT_ENCODING);
+                    if (h != null) raw = h.getValue();
+                }
+
+                AsyncDataConsumer downstream = cb.handleResponse(rsp, details);
```
ok2c: @arturobernalg Careful here. We should rewrite `EntityDetails` when we actually about to do automatic content decompression and make it return content-length as `-1` and content-encoding as `null` to signal downstream the content is going to be transformed on the fly.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientDeflatePostExample.java
```diff
@@ -0,0 +1,110 @@
...
+ *   <li>Add the mandatory {@code Content-Encoding} request header.</li>
+ *   <li>Consume the JSON echo using {@link StringAsyncEntityConsumer}.</li>
+ * </ol>
+ *
+ * <p>Swap {@code DeflatingAsyncEntityProducer} for your future
+ * {@code GzipAsyncEntityProducer} once gzip support is added.</p>
+ *
+ * @since 5.6
+ */
+public final class AsyncClientDeflatePostExample {
```
ok2c: @arturobernalg I do not quite understand how this example differs from `AsyncClientDeflateCompressionExample` and what additional value it provides. I would drop it.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
arturobernalg: On more time  @ok2c

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -0,0 +1,208 @@
...
+            public AsyncDataConsumer handleResponse(final HttpResponse rsp,
+                                                    final EntityDetails details)
+                    throws HttpException, IOException {
+
+                if (!enabled) {
+                    return cb.handleResponse(rsp, details);
+                }
+
+                String coding = details != null ? details.getContentEncoding() : null;
+                if (coding == null && rsp.containsHeader(HttpHeaders.CONTENT_ENCODING)) {
```
ok2c: @arturobernalg Please use `details#getContentEncoding()` only. It is the only attribute that is relevant here.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -0,0 +1,208 @@
...
+                if (coding != null) {
+                    patched = wrapEntityDetails(details);
+                }
+                AsyncDataConsumer downstream = cb.handleResponse(rsp, patched);
+
+                if (coding != null) {
+                    final HeaderElement[] el = BasicHeaderValueParser.INSTANCE
+                            .parseElements(coding, new ParserCursor(0, coding.length()));
+                    for (int i = el.length - 1; i >= 0; i--) {
+                        final String token = el[i].getName().trim().toLowerCase(Locale.ROOT);
```
ok2c: @arturobernalg No need to trim here. It has already been done by the parser.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -0,0 +1,208 @@
...
+    private static EntityDetails wrapEntityDetails(final EntityDetails original) {
+        return new EntityDetails() {
+            @Override
+            public long getContentLength() {
+                return -1;
+            }
+
+            @Override
+            public String getContentType() {
+                return original != null ? original.getContentType() : null;
```
ok2c: @arturobernalg `original` should never be nulll here. We cannot make an entity if there was none to begin with.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
ok2c: @arturobernalg This stills needs to be fixed.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
ok2c: @arturobernalg What about `#getTrailerNames`?

arturobernalg:
> @arturobernalg Two things still need to be fixed.
> 
> Alternatively I can pull in your change, tweak them a bit and commit it.

@ok2c I just made the missing change. Please feel free to do any tweak that you think is need it.

ok2c:
Committed as 3fbbd2237216a74cceacf8458655ff19d0120bb4.

@arturobernalg Very good job

PR #683, opened 2025-07-19
HTTPCLIENT-1482: expect-continue only on reused connections

arturobernalg:
This change introduces an opt-in strategy that limits automatic emission of the Expect: 100-continue header to requests executed over a reused (pooled) connection.
It eliminates “No response from server” ambiguity on non-idempotent requests without imposing the extra RTT on brand-new sockets.
Key points

    ExpectContinueTrigger enum – two modes: ALWAYS (default, preserves 5.x behaviour) and IF_REUSED.

    RequestConfig extension – new field, getter, and builder setter setExpectContinueTrigger(...).

    RequestExpectContinue interceptor – now consults the trigger and skips the header when the connection’s EndpointDetails#getRequestCount() is 0.

    No binary incompatibility; existing code continues to operate exactly as before.

    Unit tests added for both trigger modes.

The behaviour requested in Jira ticket HTTPCLIENT-1482 is now available to callers while keeping the classic client’s default semantics unchanged.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/config/ExpectContinueTrigger.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/config/ExpectContinueTrigger.java
@@ -1,0 +1,50 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


ok2c:
@arturobernalg It would be good to add tests to `TestRequestExpectContinue` for the new config options.

arturobernalg:
> `TestRequestExpectContinue`

@ok2c my bad. I should have added tests  Please take a look

ok2c:

PR #685, opened 2025-07-24
Add TUNNEL_PROXY support to AsyncConnectExec for proxy chain tunneling

arturobernalg:
Implement async handling of multi-hop proxy routes with CONNECT requests Update test to verify successful proxy chain establishment

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncConnectExec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AsyncConnectExec.java
@@ -331,7 +331,62 @@ public final class AsyncConnectExec implements AsyncExecChainHandler {
                 // of two proxies, where P1 must be tunnelled to P2.
                 // route: Source -> P1 -> P2 -> Target (3 hops)
                 // fact:  Source -> P1 -> Target       (2 hops)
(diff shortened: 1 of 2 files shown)


ok2c:
@arturobernalg Have you been able to test it in real-life settings with multiple proxy servers forming a proxy chain? 

Then, we would also need to make similar changes to the classic `ConnectExec`.

arturobernalg:
> @arturobernalg Have you been able to test it in real-life settings with multiple proxy servers forming a proxy chain?
> 
> Then, we would also need to make similar changes to the classic `ConnectExec`.

The classic already have a implementations  --> https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ConnectExec.java#L165

ok2c:
> > @arturobernalg Have you been able to test it in real-life settings with multiple proxy servers forming a proxy chain?
> > Then, we would also need to make similar changes to the classic `ConnectExec`.
> 
> The classic already have a implementations --> https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ConnectExec.java#L165

@arturobernalg That code is likely wrong or broken or both. There is not a single integration test case that exercises it. Instead of adding similarly broken code in the async pipeline we should rather remove it from the classic one.

PR #688, opened 2025-07-25
Add fully-streaming async GZIP producer/consumer

arturobernalg:
Adds streaming async support for Content-Encoding: gzip.
Includes `GzipAsyncEntityProducer` + `GzipAsyncEntityConsumer` with full header/trailer validation.
Unit tests & tiny examples provided;

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/GzipAsyncEntityConsumer.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/GzipAsyncEntityConsumer.java
@@ -1,0 +1,379 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/GzipAsyncEntityConsumer.java
```diff
@@ -0,0 +1,331 @@
...
+    private final Queue<Byte> hdrFixed = new ArrayDeque<Byte>(10);
+    private int flg = 0;
+    private int extraRemaining = -1;
+    private boolean wantHdrCrc = false;
+    private int hdrCrcCalc = 0;   // incremental CRC-16
+
+    private boolean headerDone = false;
+    private long uncompressed = 0;
+
+    /* ---------- completion plumbing ---------- */
```
garydgregory, line 91: I think we should use the Javadoc format instead of this ---- very noisy style.

PR #689, opened 2025-07-26
remove dependency httpcore5: httpcore5-h2 is depend on httpcore5

RayTigerZ:


ok2c:
@RayTigerZ What would be the benefit of doing so?

RayTigerZ:
@ok2c it would simplify dependencies manager to httpcomponents-client and project that depended on httpcomponents-client

RayTigerZ:
@ok2c Could you tell the purpose of the current dependency declaration?

ok2c:
> @ok2c Could you tell the purpose of the current dependency declaration?

@RayTigerZ I understand this is a mater of taste. I personally like when mandatory dependencies are _explicitly_ stated and not dragged in as transitive ones.

RayTigerZ:
> > @ok2c Could you tell the purpose of the current dependency declaration?
> 
> @RayTigerZ I understand this is a mater of taste. I personally like when mandatory dependencies are _explicitly_ stated and not dragged in as transitive ones.

understand

PR #693, opened 2025-07-30
[HTTPCLIENT-2178] Implement Happy Eyeballs V2 (RFC 8305) 

arturobernalg:
This pull request introduces a minimal implementation of Happy Eyeballs Version 2 (RFC 8305) for the asynchronous client connection operator in Apache HttpComponents Client 5.x. It focuses on a simple, straightforward approach to dual-stack (IPv6/IPv4) address resolution and staggered connection attempts, while integrating a basic RFC 6724 address selection mechanism.

Key highlights:

- Minimal RFC 6724 Support: Implements essential rules for address sorting (e.g., precedence, scope, label matching, longest prefix), with source address discovery via UDP probes. Non-critical rules (e.g., deprecated addresses, home addresses) are marked as TODO for future enhancements.

- Staggered Connections: Uses a single-threaded ScheduledExecutorService for delaying connection attempts (default: 250ms), ensuring compliance with RFC 8305's latency minimization without unnecessary complexity.

- Integration: Easily configurable via PoolingAsyncClientConnectionManagerBuilder.setConnectionOperator(). Includes shutdown methods for resource management.

### DIFF
--- a/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientHappyEyeballs.java
+++ b/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientHappyEyeballs.java
@@ -1,0 +1,313 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.examples;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Future;
+
+import org.apache.hc.client5.http.AddressSelectingDnsResolver;
+import org.apache.hc.client5.http.SystemDefaultDnsResolver;
+import org.apache.hc.client5.http.async.methods.SimpleHttpRequest;
+import org.apache.hc.client5.http.async.methods.SimpleHttpResponse;
+import org.apache.hc.client5.http.async.methods.SimpleRequestBuilder;
+import org.apache.hc.client5.http.async.methods.SimpleRequestProducer;
+import org.apache.hc.client5.http.async.methods.SimpleResponseConsumer;
+import org.apache.hc.client5.http.config.ConnectionConfig;
+import org.apache.hc.client5.http.config.ProtocolFamilyPreference;
+import org.apache.hc.client5.http.config.RequestConfig;
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.async.HttpAsyncClients;
+import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManager;
+import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManagerBuilder;
+import org.apache.hc.client5.http.ssl.ClientTlsStrategyBuilder;
+import org.apache.hc.core5.concurrent.FutureCallback;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.URIScheme;
+import org.apache.hc.core5.http.message.StatusLine;
+import org.apache.hc.core5.http.nio.ssl.TlsStrategy;
+import org.apache.hc.core5.io.CloseMode;
+import org.apache.hc.core5.util.TimeValue;
+import org.apache.hc.core5.util.Timeout;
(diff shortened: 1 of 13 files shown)


ok2c:
@rschmitt Could you please take a look at this PR as you are the requester of this feature?

rschmitt:
This is still on my to-do list, I'll get to it when I can.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientHappyEyeballs.java
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * <p><strong>References</strong></p>
+ * <ul>
+ *   <li>RFC&nbsp;8305: Happy Eyeballs Version 2</li>
+ *   <li>RFC&nbsp;6724: Default Address Selection for IPv6</li>
+ * </ul>
+ *
+ * @since 5.6
+ */
+public final class AsyncClientHappyEyeballs {
```
rschmitt, line 111: Remember, I added a runner for these:

```
$ ./run-example.sh AsyncClientHappyEyeballs
Executing request GET http://ipv6-test.com/
GET http://ipv6-test.com/->HTTP/1.1 400 Bad Request
Executing request GET https://ipv6-test.com/
SimpleBody{content length=442, content type=text/html; charset=iso-8859-1}
GET https://ipv6-test.com/->HTTP/1.1 400 Bad Request
SimpleBody{content length=442, content type=text/html; charset=iso-8859-1}
Shutting down
```

The runner can pass in command line arguments, so it'd be nice if this class could take a URI as an argument, e.g.:

```
$ ./run-example.sh AsyncClientHappyEyeballs http://neverssl.com/
```

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientHappyEyeballs.java
```diff
@@ -0,0 +1,158 @@
...
+                    .build();
+
+            System.out.println("Executing request " + request);
+            final Future<SimpleHttpResponse> future = client.execute(
+                    SimpleRequestProducer.create(request),
+                    SimpleResponseConsumer.create(),
+                    new FutureCallback<SimpleHttpResponse>() {
+
+                        @Override
+                        public void completed(final SimpleHttpResponse response) {
```
rschmitt, line 204: Do we have access to the resolved IP address from here? It'd be nice to print that

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/HappyEyeballsV2AsyncClientConnectionOperator.java
```diff
@@ -0,0 +1,821 @@
...
+ * care-of, native transport) are <strong>not yet implemented</strong>.
+ * Their placeholders are marked with TODO comments.
+ *
+ * @see <a href="https://www.rfc-editor.org/rfc/rfc8305">RFC&nbsp;8305</a>
+ * @see <a href="https://www.rfc-editor.org/rfc/rfc6724">RFC&nbsp;6724</a>
+ * @since 5.6
+ */
+@Experimental
+public final class HappyEyeballsV2AsyncClientConnectionOperator
+        implements AsyncClientConnectionOperator, ModalCloseable {
```
rschmitt: I think that `AsyncClientConnectionOperator` is too high-level for this. There's unrelated stuff in here like `upgrade`, for one thing. For another, there's an interface called `ConnectionInitiator` that seems like it would be a really good fit for an algorithm that implements staggered, concurrent, asynchronous connection establishment.

In fact, one thing we can consider is teaching the existing `MultihomeConnectionInitiator`/`MultihomeIOSessionRequester` to perform concurrent connection attempts, rather than sequential ones. That would provide the most important and most difficult part of RFC 8305, and it would work with any endpoint with multiple IP addresses, irrespective of IPv4/IPv6 configuration.

rschmitt (changes requested):
I think we should break this up into two or three smaller tasks.

1. We already have connection initiation code that tries multiple IP addresses _sequentially_, and I think we should start by adding an [option](https://www.rfc-editor.org/rfc/rfc8305#section-8) (maybe in `ConnectionConfig`) that allows the async client to _stagger_ those connection attempts instead. We could ship this disabled by default and then enable it later.
2. We can add a `DnsResolver` (or similar) that implements [RFC 6724 destination address selection rules](https://www.rfc-editor.org/rfc/rfc6724#section-6). This could be done by wrapping a delegate `DnsResolver`
3. We can add a `ConnectionConfig` to control protocol family preference, e.g. `PREFER_IPV4`, `PREFER_IPV6`, `IPV4_ONLY`, `IPV6_ONLY`, and `INTERLEAVE`.

The combination of staggered-concurrent connection attempts, destination address selection rules, and interleaved protocol families would give us basically all the `MUST`s and `SHOULD`s in RFC 8305, but the functionality would be much better integrated into existing code and abstractions, and we could ship it in smaller increments.

rschmitt:
@ok2c What do you think of the idea of upgrading the existing `MultihomeConnectionInitiator`/`MultihomeIOSessionRequester` to support the RFC 8305 connection algorithm?

ok2c:
> @ok2c What do you think of the idea of upgrading the existing `MultihomeConnectionInitiator`/`MultihomeIOSessionRequester` to support the RFC 8305 connection algorithm?

@arturobernalg @rschmitt I am very much in favor of splitting the change-set into several smaller ones starting with `MultihomeConnectionInitiator`/`MultihomeIOSessionRequester`.

arturobernalg:
all right @rschmitt  .. i do the changes

arturobernalg:
@rschmitt  please give another change to the PR

arturobernalg:
@rschmitt  @ok2c  can we merge the minimal path (staggered multihome + config flag) as **experimental** and iterate on the rest?

ok2c:
@arturobernalg This feature was requested by @rschmitt and do not think i need to get involved unless absolutely necessary. I suppose if one cannot get a complex feature through a review, splitting it into smaller change-sets is the best thing to do.

arturobernalg:
@rschmitt  I’m splitting this into smaller change-sets. I’ll land the [RFC6724](https://github.com/apache/httpcomponents-client/pull/778) DNS resolver + tests first (ordering / family preference semantics), and follow up with the other pieces in a separate PR.

PR #699, opened 2025-08-05
HTTPCLIENT-2386: Classic transport to use connect timeout as a default if TLS timeout has not been explicitly set

ok2c:
This change-set re-aligns the behavior of the classic and async connection operations and makes both transports use connect timeout as a fallback value for the TLS timeout when explicitly set.


arturobernalg (approved):
LGTM

garydgregory:
👍 I think this will help a lot of users that get confused by what timeout kicks in when.

rschmitt (commented):
Where is the original soTimeout restored after the handshake is completed?

ok2c:
> Where is the original soTimeout restored after the handshake is completed?

@rschmitt Good catch. Please do another pass.

ok2c:
@rschmitt Any objections to committing this change-set?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java
```diff
@@ -217,16 +218,23 @@ public void connect(
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("{} {} connected {}->{}", ConnPoolSupport.getId(conn), endpointHost, conn.getLocalAddress(), conn.getRemoteAddress());
                 }
-                conn.setSocketTimeout(soTimeout);
```
rschmitt, line 220: I'm not sure what this does, since we set the socket timeout directly on the socket (with a null check) on line 191.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java
```diff
@@ -217,16 +218,23 @@ public void connect(
...
                     }
+                    final TlsConfig tlsConfig = attachment instanceof TlsConfig ? (TlsConfig) attachment : TlsConfig.DEFAULT;
+                    final int soTimeout = socket.getSoTimeout();
+                    final Timeout handshakeTimeout = tlsConfig.getHandshakeTimeout() != null ? tlsConfig.getHandshakeTimeout() : connectTimeout;
+                    if (handshakeTimeout != null) {
+                        socket.setSoTimeout(handshakeTimeout.toMillisecondsIntBound());
+                    }
                     final SSLSocket sslSocket = tlsSocketStrategy.upgrade(socket, tlsName.getHostName(), tlsName.getPort(), attachment, context);
                     conn.bind(sslSocket, socket);
+                    socket.setSoTimeout(soTimeout);
```
rschmitt, line 237: I think this is much clearer than before.

rschmitt (approved):

PR #700, opened 2025-08-05
Transparent async content (de)compression with gzip

arturobernalg:
Adds `InflatingGzipDataConsumer,` `DeflatingGzipEntityProducer` and example/tests so the async client automatically advertises gzip,  inflates gzip/ responses on the fly and streams request bodies through gzip when wrapped. No buffering, CRC + ISIZE trailer written little-endian,

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DeflatingGzipEntityProducer.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DeflatingGzipEntityProducer.java
@@ -1,0 +1,250 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/DeflatingGzipEntityProducer.java
```diff
@@ -0,0 +1,250 @@
...
+    public DeflatingGzipEntityProducer(final AsyncEntityProducer delegate) {
+        this.delegate = Args.notNull(delegate, "delegate");
+        outBuf.flip(); // start in “read mode” with no data
+    }
+
+    /* ------------------- metadata ------------------- */
+
+    @Override
+    public boolean isRepeatable() {
+        return false;
```
ok2c: @arturobernalg GZIP coding should be repeatable, if the underlying entity itself is repeatable, should it not?
`delegate.isRepeatable`

ok2c (approved):

PR #701, opened 2025-08-06
Enable new TcpKeepAlive options on Java 8

rschmitt:
This change sets the new KeepAlive-related options directly through the JDK, instead of going through `SocketSupport`. There are two reasons for this.

First, the `SocketSupport` code path doesn't work on Java 8, as it attempts to reflectively call `Socket.setOption()`, which was added in newer versions.

Second, the options themselves _do_ exist on Java 8. The Java socket options were backported to Java 8 in 8u261 (July 2020):

https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8194298

As for Windows, it supports `TCP_KEEPIDLE` and `TCP_KEEPINTVL` since Windows 10 version 1709 (October 2017):

https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-tcp-socket-options

I think it's safe at this point to simply use these features like any other socket option. Unfortunately, it doesn't look like Mojo signatures are distributed for newer maintenance releases of Java 8, so I've had to update `hc.animal-sniffer.signature.ignores`.

Finally, I've added test coverage in TestTlsHandshakeTimeout to exercise this feature, as well as to assert that KeepAlive doesn't interfere with socket timeouts.


garydgregory:
Sounds good to me!

PR #707, opened 2025-08-16
Add RFC 8297 (103 Early Hints) support to HttpClient5 (async)

arturobernalg:
Introduces EarlyHintsListener to surface 103 Early Hints without altering final response flow.

Wires listener via HttpClientBuilder#setEarlyHintsListener(...) and HttpAsyncClientBuilder#setEarlyHintsListener(...).

Handles 103 in classic (MainClientExec) and async (EarlyHintsAsyncExec).

Early Hints ([RFC 8297)](https://datatracker.ietf.org/doc/html/rfc8297) allows servers to advise clients (typically via Link preload/preconnect) before the final response. Libraries should expose these hints while keeping normal processing intact.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExec.java
@@ -1,0 +1,110 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl.async;
+
+import java.io.IOException;
+
+import org.apache.hc.client5.http.EarlyHintsListener;
+import org.apache.hc.client5.http.async.AsyncExecCallback;
+import org.apache.hc.client5.http.async.AsyncExecChain;
+import org.apache.hc.client5.http.async.AsyncExecChainHandler;
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpResponse;
+import org.apache.hc.core5.http.HttpStatus;
+import org.apache.hc.core5.http.nio.AsyncDataConsumer;
+import org.apache.hc.core5.http.nio.AsyncEntityProducer;
+
+/**
+ * Execution chain handler that delivers {@code 103 Early Hints}
+ * informational responses to a user-provided
+ * {@link org.apache.hc.client5.http.EarlyHintsListener}
+ * without affecting processing of the final (non-1xx) response.
+ *
+ * <p>This handler forwards each {@code 103} informational response to the
+ * listener. All other responses (including the final response) are delegated
+ * unchanged.</p>
+ *
+ * <p>For security and interoperability, applications typically act only on
+ * headers considered safe in Early Hints (for example, {@code Link} with
+ * {@code rel=preload} or {@code rel=preconnect}).</p>
+ *
+ * @see org.apache.hc.client5.http.EarlyHintsListener
+ * @see org.apache.hc.core5.http.HttpStatus#SC_EARLY_HINTS
+ * @see org.apache.hc.core5.http.nio.ResponseChannel#sendInformation(org.apache.hc.core5.http.HttpResponse, org.apache.hc.core5.http.protocol.HttpContext)
+ * @since 5.6
+ */
+
+public final class EarlyHintsAsyncExec implements AsyncExecChainHandler {
+    private final EarlyHintsListener listener;
+
+    public EarlyHintsAsyncExec(final EarlyHintsListener listener) {
+        this.listener = listener;
+    }
+
+    @Override
+    public void execute(final HttpRequest request,
+                        final AsyncEntityProducer entityProducer,
+                        final AsyncExecChain.Scope scope,
+                        final AsyncExecChain chain,
+                        final AsyncExecCallback callback) throws HttpException, IOException {
+
+        if (listener == null) {
+            chain.proceed(request, entityProducer, scope, callback);
+            return;
+        }
+
+        chain.proceed(request, entityProducer, scope, new AsyncExecCallback() {
+            @Override
+            public void handleInformationResponse(final HttpResponse response)
+                    throws HttpException, java.io.IOException {
+                if (response.getCode() == HttpStatus.SC_EARLY_HINTS) {
+                    listener.onEarlyHints(response, scope.clientContext);
+                }
+                callback.handleInformationResponse(response);
+            }
+
+            @Override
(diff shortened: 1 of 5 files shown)


garydgregory:
Hi all,

The page https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/103 suggests `103` should only be for HTTP/2 or later. Should this PR only apply to HTTP/2?

arturobernalg:
> Hi all,
> 
> 
> 
> The page https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/103 suggests `103` should only be for HTTP/2 or later. Should this PR only apply to HTTP/2?

For what I understand MDN’s note is a deployment recommendation, not a protocol limit. RFC 8297 does not restrict 103 to HTTP/2+

garydgregory:
> > Hi all,
> > The page https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/103 suggests `103` should only be for HTTP/2 or later. Should this PR only apply to HTTP/2?
> 
> For what I understand MDN’s note is a deployment recommendation, not a protocol limit. RFC 8297 does not restrict 103 to HTTP/2+

Check. I wonder if the fact that this is an RFC marked "Experimental" should be of concern, or at least mentioned in the Javadoc.

arturobernalg:
> > > Hi all,
> 
> > > The page https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/103 suggests `103` should only be for HTTP/2 or later. Should this PR only apply to HTTP/2?
> 
> > 
> 
> > For what I understand MDN’s note is a deployment recommendation, not a protocol limit. RFC 8297 does not restrict 103 to HTTP/2+
> 
> 
> 
> Check. I wonder if the fact that this is an RFC marked "Experimental" should be of concern, or at least mentioned in the Javadoc.
> 
> 

The RFC text is immutable, so the body still says “Experimental” (that’s the original 2017 publication). However, the IESG later changed the status to Proposed Standard; that action is recorded separately and Datatracker now shows the badge accordingly. Official status-change doc (Feb 5, 2025) is here.  ￼
For reference, the original RFC body that still reads “Experimental” is here (unchanged by design).

garydgregory:
Thank you for the details @arturobernalg 👌

rschmitt:
> The page https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/103 suggests `103` should only be for HTTP/2 or later.

And rightly so. HTTP/1.1's framing is inherently unreliable and they're still finding [desync attacks](https://portswigger.net/research/http1-must-die) in the wild. Additionally, this type of scheme works best with async IO, so that the `103` response can be handled as soon as it comes in, and only HTTP/2 is reliably implemented with async IO (since it's a multiplexed protocol). I also find it strange that this RFC provides no way for the client to signal its support for this feature, analogous to `Expect: 100-continue`. Who the hell would just turn this feature on in production for all clients?

I agree that the client should at least _tolerate_ `103` responses, but I don't have a strong opinion about how to expose them. The dedicated callback in the PR seems reasonable.

ok2c:
@arturobernalg I agree with @rschmitt I would not add this feature to the classic transport at all. The implementation for the classic i/o looks clumsy and it is not worth it.

arturobernalg:
> @arturobernalg I agree with @rschmitt I would not add this feature to the classic transport at all. The implementation for the classic i/o looks clumsy and it is not worth it.

@ok2c  I’ll scope Early Hints to the async client only and drop the classic wiring.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExec.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpResponse;
+import org.apache.hc.core5.http.HttpStatus;
+import org.apache.hc.core5.http.nio.AsyncEntityProducer;
+import org.apache.hc.core5.http.protocol.HttpContext;
+
+/**
+ * Execution chain handler that delivers {@code 103 Early Hints} (RFC&nbsp;8297)
```
ok2c: @arturobernalg I would prefer if we made no references to specific RFCs in our source code. If in the future the RFC gets superseded by another specification we would not need to chase all the references to that RFC in our source code.

Commit message would be the right place to mention it.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExec.java
```diff
@@ -0,0 +1,108 @@
...
+    public EarlyHintsAsyncExec(final EarlyHintsListener listener) {
+        this.listener = listener;
+    }
+
+    @Override
+    public void execute(final HttpRequest request,
+                        final AsyncEntityProducer entityProducer,
+                        final AsyncExecChain.Scope scope,
+                        final AsyncExecChain chain,
+                        final AsyncExecCallback callback) throws HttpException, java.io.IOException {
```
ok2c: @arturobernalg Why full class name?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExec.java
```diff
@@ -0,0 +1,108 @@
...
+            public void handleInformationResponse(final HttpResponse response)
+                    throws HttpException, java.io.IOException {
+                if (response.getCode() == HttpStatus.SC_EARLY_HINTS) {
+                    listener.onEarlyHints(response, scope.clientContext);
+                }
+                callback.handleInformationResponse(response);
+            }
+
+            @Override
+            public org.apache.hc.core5.http.nio.AsyncDataConsumer handleResponse(
```
ok2c: @arturobernalg Same. Why? There is no conflict and this makes the code more difficult to read

httpclient5/src/test/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExecTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @AfterEach
+    void tearDown() {
+        if (server != null) {
+            server.close(CloseMode.GRACEFUL);
+        }
+    }
+
+    @Test
+    @org.junit.jupiter.api.Timeout(10)
```
ok2c, line 82: @arturobernalg Same. Unless I am missing something,

ok2c (commented):
@arturobernalg Looks good overall. An example / sample app would be nice, though.

httpclient5/src/test/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExecTest.java
arturobernalg: @ok2c The only reason I qualified the annotation is the name clash with `org.apache.hc.core5.util.Timeout` used in `setSoTimeout(Timeout.ofSeconds(5))`.

ok2c: @arturobernalg Understood.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/EarlyHintsAsyncExec.java
ok2c: @arturobernalg This one still uses full class name. Should it?

httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncEarlyHintsEndToEndDemo.java
```diff
@@ -0,0 +1,174 @@
...
+ *   <li>completes the exchange with a final {@code 200 OK} and a short body.</li>
+ * </ol>
+ * The async client registers an Early Hints listener, prints any received {@code 103}
+ * headers, and then prints the final status and body.</p>
+ *
+ * <p>Use this sample to see how to wire {@code setEarlyHintsListener(...)} and verify that
+ * Early Hints do not interfere with normal response processing.</p>
+ */
+
+public class AsyncEarlyHintsEndToEndDemo {
```
ok2c: @arturobernalg Can we name it `AsyncClientEarlyHintsEndToEnd` or `AsyncClientEarlyHintsEndToEndExample` for consistency with existing examples?

arturobernalg: Done

arturobernalg:
> @arturobernalg There is still one unresolved comment

solve. sorry @ok2c  I overlooked it

ok2c (approved):

PR #712, opened 2025-08-23
Clarify Javadoc for responseTimeout and socketTimeout interaction

LichKing-lee:
### What problem does this solve?

The Javadoc for responseTimeout and socketTimeout does not clearly describe their interaction. Many users assume they are independent, but in classic I/O responseTimeout actually overrides the connection’s socketTimeout during request execution. This can cause confusion when diagnosing timeout behavior.

### What does this PR do?

This PR updates the Javadoc of both RequestConfig#setResponseTimeout and ConnectionConfig#setSocketTimeout to explicitly document that behavior:
- responseTimeout overrides socketTimeout for the duration of a request.
- socketTimeout serves as the default unless overridden.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
@@ -543,6 +543,12 @@ public class RequestConfig implements Cloneable {
          * HTTP transports with message multiplexing.
          * </p>
          * <p>
+         * Please note that response timeout is not a deadline. Its absolute value
+         * can be exceeded, for example, in case of automatic request re-execution.
+         * Please make sure the automatic request re-execution policy has been
+         * configured appropriately.
+         * </p>
+         * <p>
          * This parameter may override a socket timeout configured at the connection
          * or I/O layers for the duration of a message execution. It is not a hard
          * deadline; its effective duration can be exceeded, for example, by automatic


httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
```diff
@@ -149,9 +149,19 @@ public Builder setSocketTimeout(final int soTimeout, final TimeUnit timeUnit) {
         }
 
         /**
-         * Determines the default socket timeout value for I/O operations.
+         * Determines the default socket timeout value for I/O operations on
+         * connections created by this configuration.
+         * A timeout value of zero is interpreted as an infinite timeout.
          * <p>
-         * Default: {@code null} (undefined)
```
ok2c, line 154: @LichKing-lee I do not think this line should be removed.

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
```diff
@@ -149,9 +149,19 @@ public Builder setSocketTimeout(final int soTimeout, final TimeUnit timeUnit) {
...
+         * This value serves as the baseline timeout applied when a connection is first
+         * created and managed by the connection pool.
+         * </p>
+         * <p>
+         * Please note: in classic (blocking) I/O execution, if a request-level
```
ok2c: @LichKing-lee I do not think this is the right place to mention any specific i/o layers or anything that the protocol may or may not do. `RequestConfig` is of no relevance here. What you can mention here instead is that the socket  timeout when specified at the connection layer overrides the default set by the i/o layer and in its turn can be overridden by the protocol layer

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -533,23 +533,25 @@ public Builder setConnectTimeout(final long connectTimeout, final TimeUnit timeU
         }
 
         /**
-         * Determines the timeout until arrival of a response from the opposite
-         * endpoint.
-         * <p>
+         * Determines the timeout until arrival of a response from the opposite endpoint.
```
ok2c: @LichKing-lee Please avoid formatting changes in the same change-set with the content changes

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -533,23 +533,25 @@ public Builder setConnectTimeout(final long connectTimeout, final TimeUnit timeU
...
-         * Please make sure the automatic request re-execution policy has been
-         * configured appropriately.
+         * Please note that response timeout is not a deadline. Its absolute value can
+         * be exceeded, for example, in case of automatic request re-execution. Please
+         * make sure the automatic request re-execution policy has been configured
+         * appropriately.
          * </p>
          * <p>
-         * Default: {@code null}
+         * In classic (blocking) I/O execution, if this value is set, it will be applied
```
ok2c: @LichKing-lee Same here. You can mention that this parameter can override the setting of the connection layer or the i/o layer but `ConnectionConfig` is of no relevance here

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
LichKing-lee: Thanks for guidance.
I noticed through testing that in Apache HttpComponents, responseTimeout and socketTimeout do not operate completely independently at execution time, responseTimeout effectively overrides the socket’s read timeout. My intention with this PR was to make that behavior clearer in the documentation.

Would the following revised wording be acceptable?

httpclient5/src/main/java/org/apache/hc/client5/http/config/ConnectionConfig.java
```diff
@@ -149,7 +149,15 @@ public Builder setSocketTimeout(final int soTimeout, final TimeUnit timeUnit) {
         }
 
         /**
-         * Determines the default socket timeout value for I/O operations.
+         * Determines the default socket timeout value for I/O operations on
+         * connections created by this configuration.
+         * A timeout value of zero is interpreted as an infinite timeout.
+         * <p>
+         * This value acts as a baseline configured by the I/O layer. It can be
```
ok2c, line 156: @LichKing-lee Actually this is not quite correct. The baseline is established by the respective i/o layer (through `SocketConfig` for the classic one or `IOReactorConfig` for the async one). Those settings apply to all newly created connections / channels indiscriminately. `ConnectionConfig` is one layer up and can apply selectively to specific hosts. This is reason `ConnectionConfig` exists, so one could configure different connection settings for specific hosts only.

httpclient5/src/main/java/org/apache/hc/client5/http/config/RequestConfig.java
```diff
@@ -549,6 +549,13 @@ public Builder setConnectTimeout(final long connectTimeout, final TimeUnit timeU
          * configured appropriately.
          * </p>
          * <p>
+         * This parameter may override a socket timeout configured at the connection
+         * or I/O layers for the duration of a message execution. It is not a hard
```
ok2c, line 553: @LichKing-lee Not quite. It does not apply to the entire duration of a message exchange. It defines the maximum period of inactivity between sending of a request message and receipt of a response message _head_. Once a response head has been received the connection socket timeout gets reset back to its previous value and the content of the response message gets consumed with the socket timeout configured at the connection or i/o layers.

ok2c (changes requested):

ok2c:
@LichKing-lee I committed your changes as 67f56a27fc846258c36939c4b59d5479bb056cf4 with some minor changes. Thank you for contributing this improvement.

PR #719, opened 2025-09-09
HTTPCLIENT-2396 - add Server-Sent Events client to HttpClient5.

arturobernalg:
New SSE module for HttpClient5 with `EventSource`, `SseExecutor`, and spec-compliant parsing.
Pluggable reconnect strategies, resumable IDs, and Retry-After hints; H1/H2 ready.
Defaults are simple; advanced tuning exposed via builders and strategies.
Includes docs, examples, and tests; API surface ready for review.

### DIFF
--- a/httpclient5-sse/src/main/java/org/apache/hc/client5/http/sse/EventSource.java
+++ b/httpclient5-sse/src/main/java/org/apache/hc/client5/http/sse/EventSource.java
@@ -1,0 +1,134 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.sse;
+
+import java.util.Map;
+
+import org.apache.hc.core5.annotation.Contract;
+
+/**
+ * Represents a Server-Sent Events (SSE) connection to a remote resource.
+ *
+ * <p>This interface exposes the minimal control surface for an SSE stream:
+ * starting, cancelling, header management, and basic state and offset (Last-Event-ID).
+ * Implementations are provided by this module (for example, via
+ * a factory such as {@code SseExecutor#open(...)}), which wires the
+ * {@link EventSourceListener} that receives events.</p>
+ *
+ * <h3>Thread-safety</h3>
+ * <p>Implementations should be safe for concurrent use: {@link #start()} and
+ * {@link #cancel()} are expected to be idempotent, and header methods should be
+ * safe to call at any time prior to (re)connect.</p>
+ *
+ * <h3>Last-Event-ID</h3>
+ * <p>{@code Last-Event-ID} follows the SSE spec: the most recent non-null event
+ * id is remembered by the implementation and can be sent on subsequent reconnects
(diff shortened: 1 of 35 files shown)


ok2c:
@arturobernalg Would it be OK for you to move implementation classes to `impl` package? It would make it easier to focus on the publicly visible APIs.

ok2c:
@arturobernalg What is your motivation for proposing this new module? Do you have an application that would be using it? Is there any chance of someone else contributing to it as well?

That would be the second module where you would be the sole maintainer.

arturobernalg:
> @arturobernalg What is your motivation for proposing this new module? Do you have an application that would be using it? Is there any chance of someone else contributing to it as well?
> 
> That would be the second module where you would be the sole maintainer.

My motivation is to provide minimal, low-level support for Server-Sent Events on top of the existing async HttpClient APIs. Right now users who need SSE either implement their own ad-hoc parsing and reconnect logic, or pull in a separate client library.

I do have concrete use cases for this in long-lived monitoring / event streams and would use this module in those applications instead of maintaining custom SSE code.

The module is small, self-contained and builds only on existing HttpClient primitives. The public API is intentionally limited, so the maintenance cost should stay low even if I remain the primary maintainer.

rschmitt:
> @arturobernalg Would it be OK for you to move implementation classes to `impl` package? It would make it easier to focus on the publicly visible APIs.

I don't recommend doing this. Package-private (default) visibility is the only reliable Java language mechanism we have to encapsulate implementation details. If we put implementation details in an `impl` package, they all need to be declared `public` just so we can make internal calls to them from our actually-public APIs!

ok2c:
> The module is small, self-contained and builds only on existing HttpClient primitives. The public API is intentionally limited, so the maintenance cost should stay low even if I remain the primary maintainer.

@arturobernalg There were instances when a complete module had been contributed to the project in the past: caching in client and reactive in core, but in both instances those contributions were backed by a commercial entity. At the very least those modules had a user base. In your case there is no user base, so it is unclear how this new solution would be applied in real life settings. If there is no documentation no one will even find out about it unless they look at the source code. I have no objections to inclusion of this module into the project but I think your priorities needs to shift from producing more experimental features to documenting / applying those you have already contributed.

ok2c:
> I don't recommend doing this. Package-private (default) visibility is the only reliable Java language mechanism we have to encapsulate implementation details. If we put implementation details in an `impl` package, they all need to be declared `public` just so we can make internal calls to them from our actually-public APIs!

@rschmitt That seems to imply that public API classes would have some implementation code in them that needed to be kept private / not exposed to the API consumers, because here is nothing stopping classes in impl from re-using package private code within that package. 

I understand this is more of a matter of taste, so let's agree to disagree here and let @arturobernalg decide what approach to take.

httpclient5-sse/src/main/java/org/apache/hc/client5/http/sse/EventSource.java
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Represents a Server-Sent Events (SSE) connection to a remote resource.
+ *
+ * <p>This interface exposes the minimal control surface for an SSE stream:
+ * starting, cancelling, header management, and basic state and offset (Last-Event-ID).
+ * Implementations are provided by this module (for example, via
+ * a factory such as {@code SseExecutor#open(...)}), which wires the
+ * {@link EventSourceListener} that receives events.</p>
+ *
+ * <h3>Thread-safety</h3>
```
ok2c, line 42: @arturobernalg Could you please also add `@Contract` annotations as well?

rschmitt:
> @rschmitt That seems to imply that public API classes would have some implementation code in them that needed to be kept private / not exposed to the API consumers, because here is nothing stopping classes in impl from re-using package private code within that package.

To clarify, my concern with an `impl` package is that at least one API in it needs to be declared `public` in order to facilitate internal calls within the client, and those public APIs could conceivably be declared package-private instead if they were in the same Java package as the caller. I think this aspect of code organization is frequently overlooked, which is one reason why every mature Java library ends up adding its own `@Internal` annotation. The other reason is Java's crude language-level visibility mechanisms. (Jigsaw fixes this, but I think it was delivered too late to matter much.)

arturobernalg:
> > I don't recommend doing this. Package-private (default) visibility is the only reliable Java language mechanism we have to encapsulate implementation details. If we put implementation details in an `impl` package, they all need to be declared `public` just so we can make internal calls to them from our actually-public APIs!
> 
> @rschmitt That seems to imply that public API classes would have some implementation code in them that needed to be kept private / not exposed to the API consumers, because here is nothing stopping classes in impl from re-using package private code within that package.
> 
> I understand this is more of a matter of taste, so let's agree to disagree here and let @arturobernalg decide what approach to take.

HttpClient already uses impl subpackages extensively (impl.classic, impl.async, etc.), so following the same pattern here would be consistent with the existing structure

ok2c:
@arturobernalg Please avoid using deprecated APIs in new code.

arturobernalg:
> @arturobernalg Please avoid using deprecated APIs in new code.

@ok2c checking. I'll do the change this weekend.

PR #721, opened 2025-09-12
MDW-2391 fix - call proper shutdown instead of shutdown now

koszta5:
[HTTPCLIENT-2391](https://issues.apache.org/jira/browse/HTTPCLIENT-2391) - fix

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AbstractHttpAsyncClientBase.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AbstractHttpAsyncClientBase.java
@@ -116,7 +116,7 @@ abstract class AbstractHttpAsyncClientBase extends CloseableHttpAsyncClient {
         }
         ioReactor.initiateShutdown();
         ioReactor.close(closeMode);
-        executorService.shutdownNow();
+        executorService.shutdown();
         internalClose(closeMode);
     }
 


garydgregory:
This seems like such an easy regression failure waiting to happen without a unit test.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/AbstractHttpAsyncClientBase.java
```diff
@@ -116,7 +116,7 @@ public final void close(final CloseMode closeMode) {
         }
         ioReactor.initiateShutdown();
         ioReactor.close(closeMode);
-        executorService.shutdownNow();
+        executorService.shutdown();
```
ok2c, line 119: @koszta5 Hmm. Why is `#shutdownNow` improper? Can you help me understand?

koszta5: please - read the jira ticket description (https://issues.apache.org/jira/browse/HTTPCLIENT-2391). Due to implementation of the ThreadPool (AutoShutdownDelegatedExecutorService)  shutdownNow() does not remove a phantom reference to the delegated thread pool

Shutdown does

koszta5: PS calling shutdown() after shutdownNow() as originally suggested cant be done (Thread pool does not allow that)

ok2c: @koszta5 I am very tempted to close the ticket as invalid right now and move onto other things, but in case I have trouble understanding written English, here is what the ticket says:
```
* The AutoShutdownDelegatedExecutorService correctly calls the shutdownNow() method on the underlying executor service.
* However, the phantom reference is never dereferenced and remains held internally.
```

What is `AutoShutdownDelegatedExecutorService` exactly and why does not it de-references the the phantom reference with `#shutdownNow`? How is this exactly a problem with HttpClient? Please help me understand if you expect me to merge this change-set.

koszta5: if you give me a little bit more time next week I can provide heap dump details / proof and possibly even a reproducer in a test
But to put it real simple
- in AbstractHttpAsyncClientBase constructor executor service is created as follows
```
this.executorService = Executors.newSingleThreadExecutor(threadFactory);
```
- the above leads to AutoShutdownDelegatedExecutorService (see implementation of newSingleThreadExecutor() )
- see constructor of AutoShutdownDelegatedExecutorService (it creates a phantom reference via cleanable variable in this constructor)
- see shutdown() method in AutoShutdownDelegatedExecutorService - it is overriden and runs super method but also (very importantly!) cleanable.clean() which removes the phantom reference
- on the other hand if you just call shutdownNow() on AutoShutdownDelegatedExecutorService it does not run the cleaner (its not overriden :( :( ) and does not remove the phantom reference which is very important for GC. 

Note that this problem happens at least in JDK21. Not sure about other releases. You could argue that its a JDK bug and it might - other way to look at it that its improper API usage.

I had to fork our httpclient and I can confirm that this fix works and we are no longer leaking to metaspace.

ok2c: > Note that this problem happens at least in JDK21. Not sure about other releases. You could argue that its a JDK bug and it might - other way to look at it that its improper API usage.

@koszta5 Yes, it does look like a bug in JDK21 to me. I do not see how calling `#shutdownNow` instead of `#shutdown` is wrong here just by looking at the `ExecutionService` javadocs. In fact, we do want the damn thing to shutdown immediately and forget about any unfinished tasks.

I do see however we probably should do a different shutdown sequence based on the `CloseMode` flag, and in case of `GRACEFUL` closure the use `#shutdown` followed by `#awaitTermination` may be more appropriate.

I am willing to look into that.

koszta5: ok that would resolve the problem too. Im onboard with what you are saying. 

I will try to raise this as a jdk bug

ok2c: @koszta5 I am currently crushed with my day work and a major HTTP/2 rewrite in core. Please do put a comment in the original JIRA ticket about the `CloseMode` flag and will try to get around to fixing the problem sometime next week.

koszta5: will do thanks a lot!

ok2c: @koszta5 Here is the fix I would like to propose #722

koszta5: thanks - testing it now

ok2c:
Superseded by 91d32f87846401bf64a427ebe510f3d817fb9c93

PR #722, opened 2025-09-13
HTTPCLIENT-2391: improved GRADEGUL shutdown of ExecutorService used internally by async clients

ok2c:


koszta5:
please give me ~24 hours to watch my metrics 
I have installed this to our runtime and should be able to give you confirmation tomorrow

koszta5:
I can confirm that this works - does not lead to raising non-heap. 

Additionally - I have the JDK bug reported - looks like somebody already also ran into this 2 months before me :D (JDK bug - https://bugs.java.com/bugdatabase/view_bug?bug_id=JDK-8367726)

PR #723, opened 2025-09-14
HTTPCLIENT-2395: Avoid double-encoding of RFC 5987 filename* and pass through pre-encoded values

arturobernalg:
Make ConnPoolSupport.formatStats null-safe to prevent NPEs during lease/shutdown. Add multipart tests covering EXTENDED mode and pre-encoded filename*.


httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/HttpRFC7578Multipart.java
```diff
@@ -101,8 +101,12 @@ protected void formatMultipartHeader(final MultipartPart part, final OutputStrea
                     writeBytes("=\"", out);
                     if (value != null) {
                         if (name.equalsIgnoreCase(MimeConsts.FIELD_PARAM_FILENAME_START)) {
-                            final String encodedValue = "UTF-8''" + PercentCodec.RFC5987.encode(value);
-                            writeBytes(encodedValue, StandardCharsets.US_ASCII, out);
+                            if (value.startsWith("UTF-8''")) {
+                                writeBytes(value, StandardCharsets.US_ASCII, out);
+                            } else {
+                                final String encodedValue = "UTF-8''" + PercentCodec.RFC5987.encode(value);
```
ok2c: @arturobernalg Could you please use `StringBuilder` and `PercentCodec#encode(StringBuilder, CharSequence)` here? Small optimization but still a bit less garbage.

arturobernalg: @ok2c  done

arturobernalg:
@ok2c  cherry-pick to 5.5.x. as well?

PR #724, opened 2025-09-14
Bug fix: connections managers to ensure open connections have socket timeout set based on ConnectionConfig upon lease

ok2c:
I was looking at the pooled connection activation / passivation logic and found out that connections do not reset the socket timeout on leased open connections, which can lead to some random timeout setting from the previous and likely completely unrelated message exchange. This is a defect the way I see it.

@rschmitt Please take a look and let me know if you disagree.

@arturobernalg Could you please double-check?


rschmitt (approved):
I guess either contract could be plausible, but the new behavior makes more sense to me.

PR #725, opened 2025-09-15
Bug fix: Use 1 second timeout when closing out connections inside a connection pool lock

ok2c:
There was a [report](https://lists.apache.org/thread/ll2q7kh0jtffq8j95v5w17w2zty7g6yb) on the dev list recently about clients getting stuck trying to close out a TLS connection while holding a lock on the connection pool, effectively rendering those clients completely disabled. I thought it should not never happen and it turned out I was wrong. The issue was known and got fixed in `core` but the fix never got applied to `client`.

Basically this change-set makes sure the socket timeout is set to 1 second before connections get closed out inside the pool lock.

@arturobernalg  Please double-check.


arturobernalg (approved):
LGTM

ok2c:
@arturobernalg By the way, if you have spare bandwidth and inclination, there is one potential improvement idea one could experiment with. Presently `PoolingHttpClientConnectionManager` always closes out connections while holding the global pool lock. Even with a timeout of 1 second that can essentially freeze the clients blocked waiting for a connection lease for over one second, which can be a lot for many applications. What one could do is to queue up connections to be closed with a custom `DisposalCallback` and then close the queued connections out outside of the pool lock. This approach has its risks but it would be a substantial improvement if it can be made to work properly.

PR #726, opened 2025-09-17
Pooling managers: offload connection disposal from the pool lock.

arturobernalg:
Close sockets off the pool lock via an async disposer. Improves lease latency under eviction/state mismatch with slow close.Includes classic/async tests and default-disposer baseline.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/OffLockDisposalCallback.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/OffLockDisposalCallback.java
@@ -1,0 +1,61 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl.io;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.io.CloseMode;
+import org.apache.hc.core5.io.ModalCloseable;
+import org.apache.hc.core5.pool.DisposalCallback;
+
+@Internal
+final class OffLockDisposalCallback<T extends ModalCloseable> implements DisposalCallback<T> {
+
+    private final DisposalCallback<T> delegate;
+    private final Queue<T> gracefulQueue = new ConcurrentLinkedQueue<>();
+
+    OffLockDisposalCallback(final DisposalCallback<T> delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void execute(final T closeable, final CloseMode mode) {
+        if (mode == CloseMode.IMMEDIATE) {
+            delegate.execute(closeable, CloseMode.IMMEDIATE);
+        } else {
+            gracefulQueue.offer(closeable);
+        }
+    }
+
+    void drain() {
+        for (T c; (c = gracefulQueue.poll()) != null; ) {
+            delegate.execute(c, CloseMode.GRACEFUL);
+        }
+    }
+}
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
@@ -72,6 +72,7 @@ import org.apache.hc.core5.http.protocol.HttpContext;
 import org.apache.hc.core5.io.CloseMode;
 import org.apache.hc.core5.pool.ConnPoolControl;
 import org.apache.hc.core5.pool.DefaultDisposalCallback;
+import org.apache.hc.core5.pool.DisposalCallback;
 import org.apache.hc.core5.pool.LaxConnPool;
 import org.apache.hc.core5.pool.ManagedConnPool;
 import org.apache.hc.core5.pool.PoolConcurrencyPolicy;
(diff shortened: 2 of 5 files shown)


ok2c:
@arturobernalg This functionality is necessary for the classic connection management only where connections can get blocked in the `#close` method. The async connection management is unaffected as `#close` method there merely enqueues a command to be executed by the i/o reactor asynchronously. 

Please reduce the scope to `PoolingHttpClientConnectionManager` only

arturobernalg:
> @arturobernalg This functionality is necessary for the classic connection management only where connections can get blocked in the `#close` method. The async connection management is unaffected as `#close` method there merely enqueues a command to be executed by the i/o reactor asynchronously.
> 
> Please reduce the scope to `PoolingHttpClientConnectionManager` only

Done

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/AsyncDisposalCallback.java
```diff
@@ -0,0 +1,128 @@
...
+
+        Item(final X c, final CloseMode m) {
+            this.closeable = c;
+            this.mode = m;
+        }
+    }
+
+    private final BlockingQueue<Item<C>> queue;
+    private final AtomicBoolean running = new AtomicBoolean(true);
+    private final Thread worker;
```
ok2c: @arturobernalg I think you are over-complicating it. Just drain the queue at the end of `#lease`, `#release` and other methods that can call `PoolEntry#discardConnection`. As long as it is done outside the pool lock it is good enough.

arturobernalg: > @arturobernalg I think you are over-complicating it. Just drain the queue at the end of `#lease`, `#release` and other methods that can call `PoolEntry#discardConnection`. As long as it is done outside the pool lock it is good enough.

what about this last change-set  @ok2c

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/OffLockDisposalCallback.java
```diff
@@ -0,0 +1,104 @@
...
+    private final ReentrantLock qlock = new ReentrantLock();
+    private final AtomicBoolean draining = new AtomicBoolean();
+
+    OffLockDisposalCallback(final DisposalCallback<T> delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void execute(final T closeable, final CloseMode mode) {
+        qlock.lock();
```
ok2c: @arturobernalg Very good. I think though we can still make it simpler

1. Let's enqueue connections in GRACEFUL mode and close them immediately in IMMEDIATE. This will also eliminate the `Item` inner class
2. We can use `ConcurrentLinkedQueue` and eliminate the need to have a lock. Less locks, less chances something can get dead-locked. There is no harm in multiple threads draining the queue concurrently as long as do not attempt to close out the same connection.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/OffLockDisposalCallback.java
arturobernalg: @ok2c  Please  do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/AsyncDisposalCallback.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.hc.core5.pool.DisposalCallback;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Asynchronously disposes {@code SocketModalCloseable} instances on a daemon thread,
+ * delegating the actual close to {@link DefaultDisposalCallback}.
+ */
+@Internal
+final class AsyncDisposalCallback<C extends SocketModalCloseable>
```
arturobernalg: removed.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -741,6 +753,9 @@ public void close(final CloseMode closeMode) {
             final PoolEntry<HttpRoute, ManagedHttpClientConnection> poolEntry = poolEntryRef.get();
             if (poolEntry != null) {
                 poolEntry.discardConnection(closeMode);
+                if (disposer != null) {
```
ok2c: @arturobernalg I see no harm in delaying `#drainDisposals` execution until the endpoint is passed to `#release`. 

BUT, Please look at the `LeaseRequest#get`. Connections can be closed in the course of that method. So, `#drainDisposals`  should be called there. Please investigate.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
arturobernalg: Done.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -702,11 +712,14 @@ public void setValidateAfterInactivity(final TimeValue validateAfterInactivity)
     static class InternalConnectionEndpoint extends ConnectionEndpoint implements ConnectionHolder, Identifiable {
 
         private final AtomicReference<PoolEntry<HttpRoute, ManagedHttpClientConnection>> poolEntryRef;
+        private final OffLockDisposalCallback<ManagedHttpClientConnection> disposer;
```
ok2c: @arturobernalg This should no longer be necessary

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -409,6 +415,7 @@ public ConnectionEndpoint get(
                     }
                 } finally {
                     lock.unlock();
```
ok2c: @arturobernalg This is a lease lock., not a global pool lock. It is perfectly OK to call `#drainDisposals` inside the method, for instance, at line 395 (prior to ` final ManagedHttpClientConnection conn = poolEntry.getConnection();`)

arturobernalg: > @arturobernalg This is a lease lock., not a global pool lock. It is perfectly OK to call `#drainDisposals` inside the method, for instance, at line 395 (prior to ` final ManagedHttpClientConnection conn = poolEntry.getConnection();`)

@ok2c drainDisposals() now runs inside LeaseRequest#get right before poolEntry.getConnection() (lease lock only), with the final safety drain retained; tests updated accordingly.

ok2c: @arturobernalg The chances of anything going wrong in the remaining code is super-slim. And if it did happen the queue would get drained by another call / thread. But I think one `#drainDisposals` call should be enough here. Pick one you prefer. I leave it up to you.

arturobernalg:
> Now the big question is whether this approach is stable enough to be activated by default or shall we mark it experimental and make optional in 5.6 release series?

@ok2c 
opt-in experimental  for 5.6, gather data, then flip on by default in 5.7 if no regressions... Please take another look.

ok2c (approved):

PR #728, opened 2025-09-27
HTTPCLIENT-2398 -  Cap async execution queue to break recursive re-entry

arturobernalg:
Implements Option 1 from the discussion: add a configurable cap on the `async` execution pipeline to prevent the recursive callback loop that led to `StackOverflowError`.


ok2c:
@arturobernalg Could you please rebase this change-set?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
```diff
@@ -59,13 +75,15 @@ static class ReUseData {
         final Object state;
         final TimeValue validDuration;
 
-      ReUseData(final Object state, final TimeValue validDuration) {
-        this.state = state;
-        this.validDuration = validDuration;
-      }
+        ReUseData(final Object state, final TimeValue validDuration) {
```
ok2c: @arturobernalg tabs?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
```diff
@@ -218,8 +250,7 @@ public Cancellable connectEndpoint(
             return Operations.nonCancellable();
         }
         final RequestConfig requestConfig = context.getRequestConfigOrDefault();
-        @SuppressWarnings("deprecation")
-        final Timeout connectTimeout = requestConfig.getConnectTimeout();
+        @SuppressWarnings("deprecation") final Timeout connectTimeout = requestConfig.getConnectTimeout();
```
ok2c: @arturobernalg Why? Is this needed?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
```diff
@@ -241,7 +272,7 @@ public void completed(final AsyncConnectionEndpoint endpoint) {
                         }
                     }
 
-        }));
+                }));
```
ok2c: @arturobernalg Is this needed? Formatting noise?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
```diff
@@ -282,10 +313,115 @@ public EndpointInfo getEndpointInfo() {
...
+            public void cancel() {
+                try {
+                    h.cancel();
+                } finally {
+                    finish();
+                }
+            }
+
+            @Override
+            public void releaseResources() {
```
ok2c: @arturobernalg I imagine calling `#finish` in `#releaseResources` only should be enough and probably more semantically correct. 

Can we also avoid wrapping each and every `AsyncClientExchangeHandler` instead? Can we do it only when `sharedQueued` is being used? By the way, you may want to use dynamic proxy here in order to avoid having to override every method in the interface.

ok2c (changes requested):
@arturobernalg This is a good start. Please also do not forget about `InternalH2AsyncExecRuntime`.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
arturobernalg: @ok2c  please do another pass

ok2c:
@arturobernalg Looks good. Can we add the same functionality to `H2AsyncClientBuilder` as well?

arturobernalg:
> ks good. Can we add the same functionality to `H2AsyncClientBuilder` as well?

@ok2c  please do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalH2AsyncExecRuntime.java
```diff
@@ -184,35 +206,34 @@ public Cancellable connectEndpoint(
         }
         final HttpRoute route = endpoint.route;
         final RequestConfig requestConfig = context.getRequestConfigOrDefault();
-        @SuppressWarnings("deprecation")
-        final Timeout connectTimeout = requestConfig.getConnectTimeout();
+        @SuppressWarnings("deprecation") final Timeout connectTimeout = requestConfig.getConnectTimeout();
```
ok2c: @arturobernalg There is again plenty of white space / formatting noise. Please try to avoid it.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalH2AsyncExecRuntime.java
```diff
@@ -246,10 +267,64 @@ public EndpointInfo getEndpointInfo() {
...
+        }
+    }
+
+    private void releaseSlot() {
+        if (sharedQueued != null) {
+            sharedQueued.decrementAndGet();
+        }
+    }
+
+    private AsyncClientExchangeHandler guard(final AsyncClientExchangeHandler handler) {
```
ok2c: @arturobernalg I am not super crazy about code re-use at all cost but in this case there is just too much duplicate code. Please consider extracting into a separate proxy class similar to `RequestEntityProxy` / `ResponseEntityProxy`

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
```diff
@@ -66,6 +70,9 @@ static class ReUseData {
 
     }
 
+    private static final ConcurrentMap<AsyncClientConnectionManager, AtomicInteger> QUEUE_COUNTERS =
```
ok2c: @arturobernalg This is bad. We need to find a way to maintain the counter without using a static hash map. The limit should ideally be per connection. If this is too difficult, then, per client instance.

ok2c (changes requested):

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/InternalHttpAsyncExecRuntime.java
arturobernalg: @ok2c  please do another pass

ok2c: @arturobernalg Do not see how this is any better. The static map is still there and what you have is essentially a major resource leak right there.

ok2c (changes requested):

arturobernalg:
@ok2c  got it. No more static var.

ok2c:
@arturobernalg This is better. Please mark new setters in the builder classes as experimental.

Okay, this is what we have now

Users could now limit the number of concurrent active requests on a per client basis. How useful is that, though? Say, one sets max concurrent request number to 1000, then opens 100 active connections and equally distributes requests over those connections. 10 concurrent requests per connection is very little. However, having just one connection and trying to execute 1000 requests concurrently over it would likely be wrong.

What would be more useful is to cap the total number of requests on a per connection basis. 

What could also be useful is enqueuing requests instead of failing them and executing them as soon as the number of concurrent requests drops below the maximum. Moreover, the same queuing mechanism could be used to re-queue and retry requests that did not get executed due to the opposite endpoint closing the connection. 

This change-set is a start in the right direction. Let's keep it though clearly marked as experimental for now.

arturobernalg:
> What could also be useful is enqueuing requests instead of failing them and executing them as soon as the number of concurrent requests drops below the maximum. Moreover, the same queuing mechanism could be used to re-queue and retry requests that did not get executed due to the opposite endpoint closing the connection.

@ok2c Marked the new builder setters as @Experimental. Agreed on the limitations of a per-client cap; I’ll keep this change-set scoped to fail-fast limiting for now. I’ll follow up with separate PRs for (1) per-connection caps and (2) optional bounded queuing / re-queue of not-yet-started requests.

ok2c:
>  I’ll follow up with separate PRs for (1) per-connection caps

@arturobernalg Please note this will likely have to be done in core.

arturobernalg:
> >  I’ll follow up with separate PRs for (1) per-connection caps
> 
> 
> 
> @arturobernalg Please note this will likely have to be done in core.

@ok2c Ill take a look the option to move it to the core

arturobernalg:
> > I’ll follow up with separate PRs for (1) per-connection caps
> 
> @arturobernalg Please note this will likely have to be done in core.

@ok2c  I replaced the `AsyncClientExchangeHandler` dynamic proxy with a simple delegating wrapper so we don’t pay reflection overhead on the hot path; behavior stays the same, with the slot released exactly once on `releaseResources()`. Please do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ReleasingAsyncClientExchangeHandler.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public void releaseResources() {
+        try {
+            handler.releaseResources();
+        } finally {
+            if (released.compareAndSet(false, true)) {
+                try {
+                    onRelease.run();
+                } catch (final RuntimeException ignore) {
```
ok2c: @arturobernalg Why do need to catch and ignore all `RuntimeException`s here? This does not look quite right to me. Why not let it propagate?

arturobernalg: > @arturobernalg Why do need to catch and ignore all `RuntimeException`s here? This does not look quite right to me. Why not let it propagate?

@ok2c  Agreed. There’s no good reason to swallow `RuntimeException` here; onRelease is an internal decrement callback and should not throw. I removed the catch and keep it guarded by the CAS in the finally block.

PR #731, opened 2025-10-06
RFC 7639 ALPN 

arturobernalg:
Add ConnectAlpnProvider and inject ALPN header in ConnectExec/AsyncConnectExec. Provide builder hooks for fixed list or provider-driven values.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
@@ -1,0 +1,111 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.client5.http.impl;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.Internal;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.ProtocolException;
+import org.apache.hc.core5.http.message.MessageSupport;
+import org.apache.hc.core5.net.PercentCodec;
+import org.apache.hc.core5.util.Args;
+
+/**
+ * Codec for the HTTP {@code ALPN} header field (RFC 7639).
+ *
+ * @since 5.7
+ */
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+@Internal
+public final class AlpnHeaderSupport {
+
+    private AlpnHeaderSupport() {
+    }
+
+    /**
+     * Formats a list of raw ALPN protocol IDs into a single {@code ALPN} header.
+     */
+    public static Header formatValue(final List<String> protocolIds) {
+        Args.notEmpty(protocolIds, "protocolIds");
+        return MessageSupport.headerOfTokens(HttpHeaders.ALPN, protocolIds, AlpnHeaderSupport::encodeId);
+    }
+
+    /**
+     * Parses an {@code ALPN} header into decoded protocol IDs.
+     *
+     * @throws ProtocolException if a token is not a well-formed percent-encoded protocol ID.
+     */
+    public static List<String> parseValue(final Header header) throws ProtocolException {
+        final List<String> tokens = new ArrayList<>();
+        MessageSupport.parseTokens(header, tokens::add);
+        final List<String> out = new ArrayList<>(tokens.size());
+        for (final String token : tokens) {
+            out.add(decodeId(token));
+        }
+        return out;
+    }
+
+    /**
+     * Encodes a single raw protocol ID to canonical token form using the HTTP token codec
+     * from core, which keeps RFC 7230 {@code tchar} octets literal and percent-encodes the
+     * rest (including {@code '%'}) with uppercase hexadecimal.
+     */
+    public static String encodeId(final String id) {
+        Args.notBlank(id, "id");
+        return PercentCodec.HTTP_TOKEN.encode(id);
+    }
+
+    /**
+     * Decodes a percent-encoded token to a raw protocol ID using UTF-8.
+     * <p>
+     * A {@code '%'} that is not followed by two hexadecimal digits is a malformed
+     * token and is rejected as a protocol error.
+     *
+     * @throws ProtocolException if the token contains malformed percent-encoding.
+     */
+    public static String decodeId(final String token) throws ProtocolException {
+        Args.notBlank(token, "token");
+        for (int i = 0; i < token.length(); i++) {
+            if (token.charAt(i) == '%') {
+                if (i + 2 >= token.length()
+                        || Character.digit(token.charAt(i + 1), 16) < 0
+                        || Character.digit(token.charAt(i + 2), 16) < 0) {
+                    throw new ProtocolException("Malformed percent-encoding in ALPN protocol id: " + token);
+                }
+                i += 2;
+            }
+        }
+        return PercentCodec.decode(token, StandardCharsets.UTF_8);
+    }
+
+}
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
@@ -70,6 +70,7 @@ import org.apache.hc.client5.http.impl.auth.BearerSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.DigestSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.ScramSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.SystemDefaultCredentialsProvider;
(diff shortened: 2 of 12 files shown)


ok2c:
@arturobernalg Could you please rebase this change-set?

arturobernalg:
> @arturobernalg Could you please rebase this change-set?

@ok2c  done.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+        return sb.toString();
+    }
+
+    /**
+     * Decodes percent-encoded token to raw ID using UTF-8.
+     * Accepts lowercase hex; malformed/incomplete sequences are left literal.
+     */
+    public static String decodeId(final String token) {
```
ok2c: @arturobernalg what is the reason having to re-implement percent coding and not using `PercentCodec` from core?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private static final char[] HEXADECIMAL = "0123456789ABCDEF".toCharArray();
+
+    private AlpnHeaderSupport() {
+    }
+
+    /**
+     * Formats a list of raw ALPN protocol IDs into a single {@code ALPN} header value.
+     */
+    public static String formatValue(final List<String> protocolIds) {
```
ok2c: @arturobernalg Use `MessageSupport#headerOfTokens` instead

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,183 @@
...
+            sb.append(encodeId(id));
+            first = false;
+        }
+        return sb.toString();
+    }
+
+    /**
+     * Parses an {@code ALPN} header value into decoded protocol IDs.
+     */
+    public static List<String> parseValue(final String value) {
```
ok2c: @arturobernalg Use  `MessageSupport#parseTokens(Header, Consumer)` instead.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,151 @@
...
+     * Formats a list of raw ALPN protocol IDs into a single {@code ALPN} header value.
+     */
+    public static String formatValue(final List<String> protocolIds) {
+        Args.notEmpty(protocolIds, "protocolIds");
+        final List<String> tokens = new ArrayList<>(protocolIds.size());
+        for (final String id : protocolIds) {
+            tokens.add(encodeId(id));
+        }
+        final Header header = MessageSupport.headerOfTokens(HEADER_NAME, tokens);
+        return header.getValue();
```
ok2c: @arturobernalg Noooooooo, please do not do that! You create a header with a char buffer, than you you get the value of it as a string, then you make char buffer with a copy of the string and then create a header. Please do not do that.

Please focus more on quality of your contributions than their volume,

My other two comments have not been addressed.

arturobernalg: @ok2c  sorry about that. I think all are fine now

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,149 @@
...
+    public static Header formatValue(final List<String> protocolIds) {
+        Args.notEmpty(protocolIds, "protocolIds");
+        return MessageSupport.headerOfTokens(HttpHeaders.ALPN, protocolIds, AlpnHeaderSupport::encodeId);
+    }
+
+    /**
+     * Parses an {@code ALPN} header into decoded protocol IDs.
+     */
+    public static List<String> parseValue(final Header header) {
+        if (header == null) {
```
ok2c: @arturobernalg This is an internal method. It does not need to do strict input validation. This is the responsibility of public methods. Keep it simple and small.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,149 @@
...
+                out.add(decodeId(token));
+            }
+        });
+        return out;
+    }
+
+    /**
+     * Encodes a single raw protocol ID to canonical token form.
+     */
+    public static String encodeId(final String id) {
```
ok2c, line 83: @arturobernalg I asked this already, I have to ask it again. What is the reason for having to re-implement percent coding and not using PercentCodec from core? What is wrong with PercentCodec encoding?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Decodes percent-encoded token to raw ID using UTF-8.
+     * Malformed / incomplete sequences are left literal.
+     */
+    public static String decodeId(final String token) {
+        Args.notBlank(token, "token");
+        try {
+            return PercentCodec.decode(token, StandardCharsets.UTF_8);
+        } catch (final IllegalArgumentException ex) {
+            return token;
```
ok2c: @arturobernalg This is a protocol error. Should not we treat it as such?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/AlpnHeaderSupport.java
arturobernalg: @ok2c RFC 7639 requires canonical token encoding: octets MUST NOT be percent-encoded if they are valid RFC 7230 tchar (except ‘%’), and percent-encoding MUST use uppercase hex.

ok2c: @arturobernalg There are two publicly accessible instances and they both do not fit the requirements?
```
    public static final PercentCodec RFC3986 = new PercentCodec(UNRESERVED);
    public static final PercentCodec RFC5987 = new PercentCodec(RFC5987_UNRESERVED);
```
> percent-encoding MUST use uppercase hex

This I do not quite understand. Can you give me an example of what `PercentCodec` produces and what this new RFC requires?

ok2c: @arturobernalg At any rate, could you contribute the correct codec to core? That would be the proper way to start.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncClientBuilder.java
```diff
@@ -936,6 +941,26 @@ public final HttpAsyncClientBuilder disableRequestPriority() {
...
+     * If {@code ids} is {@code null} or empty, no {@code ALPN} header will be added.</p>
+     *
+     * <p>This is a convenience method equivalent to installing a {@link ConnectAlpnProvider}
+     * that always returns the same list.</p>
+     *
+     * @param ids ALPN protocol IDs to advertise (for example {@code "h2"} and {@code "http/1.1"})
+     * @return this builder
+     * @since 5.6
+     */
+    public HttpAsyncClientBuilder setConnectAlpn(final String... ids) {
```
ok2c: @arturobernalg What should happen if the caller passes `h3`, `http/0.75`, `srtuff` to this method? 

Why do we even need this method at all? There is already `HttpVersionPolicy` settable on a per route basis and this is exactly what you should be using should you decide to proceed further with this feature.

ok2c:
@arturobernalg This change can still make it into `5.7` if you are willing to invest time into it.

arturobernalg:
> @arturobernalg This change can still make it into `5.7` if you are willing to invest time into it.

@ok2c  please do another pass

httpclient5/src/main/java/org/apache/hc/client5/http/ConnectAlpnProvider.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Returns the ALPN protocol IDs to advertise for a tunnel to {@code target}
+     * over the given {@code route}.
+     *
+     * @param target the origin server the tunnel will connect to (non-null)
+     * @param route  the planned connection route, including proxy info (non-null)
+     * @return list of protocol IDs (e.g., {@code "h2"}, {@code "http/1.1"});
+     * {@code null} or empty to omit the header
+     */
+    List<String> getAlpnForTunnel(HttpHost target, HttpRoute route);
```
ok2c: @arturobernalg Same question that I already asked you before. What do you think should happen if this provider returns something other than `h2` or `http/1.1` from supported `ApplicationProtocol` enum? Why do we need this interface at all?

ok2c (commented):
@arturobernalg Much, much better! However, we need to figure out a better of way of passing the actual TLS policy to the interceptors. I would try to solve it by adding `HttpVersionPolicy` as an attribute to `HttpClientContext`

httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java
```diff
@@ -552,6 +553,8 @@ public void connect(final ConnectionEndpoint endpoint, final TimeValue timeout,
         final HttpHost firstHop = route.getProxyHost() != null ? route.getProxyHost() : route.getTargetHost();
         final SocketConfig socketConfig = resolveSocketConfig(route);
         final ConnectionConfig connectionConfig = resolveConnectionConfig(route);
+        final TlsConfig tlsConfig = resolveTlsConfig(route.getTargetHost());
+        HttpClientContext.castOrCreate(context).setHttpVersionPolicy(tlsConfig.getHttpVersionPolicy());
```
ok2c: @arturobernalg One last bit. Let's do `HttpClientContext#cast` here instead and assign the attribute only if the context is non null. There is no point creating a new context here.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/nio/PoolingAsyncClientConnectionManager.java
```diff
@@ -504,13 +505,15 @@ public Future<AsyncConnectionEndpoint> connect(
         if (LOG.isDebugEnabled()) {
             LOG.debug("{} connecting endpoint to {} ({})", ConnPoolSupport.getId(endpoint), firstHop, connectTimeout);
         }
+        final TlsConfig targetTlsConfig = resolveTlsConfig(route.getTargetHost());
+        HttpClientContext.castOrCreate(context).setHttpVersionPolicy(targetTlsConfig.getHttpVersionPolicy());
```
ok2c: @arturobernalg Same

ok2c (changes requested):

ok2c (changes requested):
@arturobernalg Almost  there. See my comments

arturobernalg:
> @arturobernalg Almost there. See my comments

Please do a

> @arturobernalg Almost there. See my comments

@ok2c Please do another pass.

ok2c (approved):
@arturobernalg Please add RFC 7639 to the list of supported RFCs in README.md and feel free to merge.

PR #778, opened 2026-01-09
Rfc6724 resolver ordering semantics

arturobernalg:
Introduce Rfc6724AddressSelectingDnsResolver and unit tests.
Define INTERLEAVE as “no bias” (preserve RFC6724 order); keep ONLY/PREFER behavior unchanged.


httpclient5/src/test/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolverTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+class Rfc6724AddressSelectingDnsResolverTest {
+
+    private DnsResolver delegate;
+
+    @BeforeEach
+    void setUp() {
+        delegate = Mockito.mock(DnsResolver.class);
```
rschmitt: Seems like `InMemoryDnsResolver` would also be fine here?

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+            }
+        }
+        return new PolicyEntry(null, 40, 1);
+    }
+
+    private static final Comparator<Info> RFC6724_COMPARATOR = (a, b) -> {
+        final InetAddress DA = a.dst;
+        final InetAddress DB = b.dst;
+        final InetAddress SourceDA = a.src;
+        final InetAddress SourceDB = b.src;
```
rschmitt: These should be named something like `aDst`, `bDst`, `aSrc`, and `aDst` respectively, so it doesn't look like you're calling static methods later on

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+        final InetAddress SourceDB = b.src;
+        final Attr attrDA = a.dstAttr;
+        final Attr attrDB = b.dstAttr;
+        final Attr attrSourceDA = a.srcAttr;
+        final Attr attrSourceDB = b.srcAttr;
+
+        final int preferDA = -1;
+        final int preferDB = 1;
+
+        // Rule 1: Avoid unusable destinations.
```
rschmitt, line 578: This is specifically section 6 (destination address selection); there are separate rules for source address selection

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+    private static Scope classifyScope(final InetAddress ip) {
+        if (ip.isLoopbackAddress()) {
+            return Scope.INTERFACE_LOCAL;
+        }
+        if (ip.isLinkLocalAddress()) {
+            return Scope.LINK_LOCAL;
+        }
+        if (ip.isMulticastAddress()) {
+            if (ip instanceof Inet6Address) {
+                // RFC 4291: low 4 bits of second byte are scope.
```
rschmitt: Also refer to RFC 6724 §3.1 which explains how scope comparisons work between multicast and unicast addresses

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+     */
+    public Rfc6724AddressSelectingDnsResolver(
+            final DnsResolver delegate,
+            final ProtocolFamilyPreference familyPreference) {
+        this.delegate = java.util.Objects.requireNonNull(delegate, "delegate");
+        this.familyPreference = familyPreference != null ? familyPreference : ProtocolFamilyPreference.INTERLEAVE;
+    }
+
+    @Override
+    public InetAddress[] resolve(final String host) throws UnknownHostException {
```
rschmitt, line 122: I think that debug logging should just log the resolution result. Everything else should be logged at trace level.

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+        }
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("RFC6724 input candidates: {}", fmt(addrs));
+        }
+
+        final List<InetSocketAddress> sockAddrs = new ArrayList<>(addrs.size());
+        for (final InetAddress a : addrs) {
+            sockAddrs.add(new InetSocketAddress(a, PROBE_PORT));
+        }
+        final List<InetAddress> srcs = srcAddrs(sockAddrs);
```
rschmitt: There's no clear reason to create this intermediate collection `sockAddrs`

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+        final List<InetSocketAddress> sockAddrs = new ArrayList<>(addrs.size());
+        for (final InetAddress a : addrs) {
+            sockAddrs.add(new InetSocketAddress(a, PROBE_PORT));
+        }
+        final List<InetAddress> srcs = srcAddrs(sockAddrs);
+
+        final List<Info> infos = new ArrayList<>(addrs.size());
+        for (int i = 0; i < addrs.size(); i++) {
+            final InetAddress dst = addrs.get(i);
+            final InetAddress src = srcs.get(i);
```
rschmitt: I don't like seeing indexing into arrays; it would probably make more sense to just build the `Info` structs one at a time and add them to the list as you go

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+                        info.srcAttr.scope, info.srcAttr.precedence, info.srcAttr.label);
+            }
+        }
+
+        infos.sort(RFC6724_COMPARATOR);
+
+        final List<InetAddress> out = new ArrayList<>(infos.size());
+        for (final Info info : infos) {
+            out.add(info.dst);
+        }
```
rschmitt: You can just stream-map-collect this in a single line

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+            InetAddress src = null;
+            try (final DatagramSocket s = new DatagramSocket()) {
+                s.connect(dest); // does not send packets; OS picks source addr/if
+                src = s.getLocalAddress();
+            } catch (final SocketException ignore) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("RFC6724 could not infer source address for {}: {}", dest, ignore.toString());
+                }
+            }
+            srcs.add(src);
```
rschmitt: This code path will insert a `null` that you have to check for in all downstream code. `ipAttrOf` treats it as a `GLOBAL` address for some reason, etc. Why do this? It seems weird. Does the RFC speak to this case?

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+            }
+        }
+    }
+
+    private static List<InetAddress> srcAddrs(final List<InetSocketAddress> addrs) {
+        final List<InetAddress> srcs = new ArrayList<>(addrs.size());
+        for (final InetSocketAddress dest : addrs) {
+            InetAddress src = null;
+            try (final DatagramSocket s = new DatagramSocket()) {
+                s.connect(dest); // does not send packets; OS picks source addr/if
```
rschmitt: How'd you find this trick? This to me is the most interesting part of the PR; I wasn't sure how to get source addresses/interfaces for each potential destination address.

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    private static Scope classifyScope(final InetAddress ip) {
+        if (ip.isLoopbackAddress()) {
+            return Scope.INTERFACE_LOCAL;
+        }
+        if (ip.isLinkLocalAddress()) {
+            return Scope.LINK_LOCAL;
+        }
+        if (ip.isMulticastAddress()) {
```
rschmitt, line 460: We're writing this for use in an HTTP client, not as a general-purpose resolver (e.g. [JEP 418](https://openjdk.org/jeps/418)), and I'm pretty sure TCP connections can only be established to unicast addresses. Should we just filter out multicast addresses?

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
rschmitt: It looks like this also has the effect of delegating source address selection to the JDK, since much of the RFC is about how to deal with multiple candidate source addresses for a given destination, e.g. this example from §10.1:

```
   Destination: 2001:db8:1::1
   Candidate Source Addresses: 2001:db8:1::1 (deprecated) or
   2001:db8:2::1
   Result: 2001:db8:1::1 (prefer same address)
```

httpclient5/src/main/java/org/apache/hc/client5/http/config/ProtocolFamilyPreference.java
```diff
@@ -0,0 +1,67 @@
...
+     * Use only IPv6 addresses.
+     */
+    IPV6_ONLY,
+
+    /**
+     * Interleave address families (v6, then v4, then v6, …) when multiple
+     * addresses are available. When staggered connects are enabled, the first
+     * address of the other family is delayed by a small offset.
+     */
+    INTERLEAVE
```
rschmitt, line 69: This is identical to `DEFAULT`. Why not just implement interleaving here? It's potentially useful, even without staggered concurrent connection attempts; for example, connection failures can trigger a retry on a different address family, since I believe our default behavior is to retry on the next IP address when more than one is returned by the resolver.

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+                src = s.getLocalAddress();
```
rschmitt: You probably want to inject a thing here that you can mock for unit testing purposes, e.g. a `Function<InetSocketAddress, InetAddress>` (or an equivalent that `throws IOException`).

httpclient5/src/test/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolverTest.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.hc.client5.http.config.ProtocolFamilyPreference;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+class Rfc6724AddressSelectingDnsResolverTest {
```
rschmitt: I think we need more unit test coverage here. I'd also like to see an integration test or a runnable example (with debug/trace logging enabled) that resolves a given DNS name. That would afford some non-trivial manual testing using loopback addresses, `/etc/hosts`, Docker networking, etc.

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,595 @@
...
+                    final boolean isV6 = a instanceof Inet6Address;
+                    if (preferV6 && isV6 || !preferV6 && !isV6) {
+                        first.add(a);
+                    } else {
+                        second.add(a);
+                    }
+                }
+                final List<InetAddress> merged = new ArrayList<>(rfcSorted.size());
+                merged.addAll(first);
+                merged.addAll(second);
```
rschmitt: Since we're already relying on the JDK to provide a stable sort, you can write this as:

```java
List<InetAddress> merged =
    rfcSorted.stream()
        .sorted(Comparator.comparing(
            a -> ((a instanceof Inet6Address) != preferV6)
        ))
        .toList();
```

rschmitt (changes requested):
Main thing here is that I'd like to see a proper implementation of `INTERLEAVE`, more unit test coverage, and an example program we can use for manual testing.

arturobernalg:
@rschmitt  Please do another pass. I think i solve all your remarks

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/ManualRfc6724ResolverIT.java
```diff
@@ -0,0 +1,87 @@
...
+
+class ManualRfc6724ResolverIT {
+
+    private static final String PROP_ENABLE = "httpclient.rfc6724.it";
+    private static final String PROP_HOST = "httpclient.rfc6724.host";
+
+    @Test
+    void resolve_and_dump_order() throws Exception {
+        assumeTrue(Boolean.getBoolean(PROP_ENABLE),
+                "Enable with -Dhttpclient.rfc6724.it=true");
```
rschmitt: I'd rather control this with JUnit 5 tags somehow. The advantage of what you've done here is that we don't have to define an entire Maven profile for this test, but we could add support in our build for controlling test selection via system properties.

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+ * <p>
+ * {@link ProtocolFamilyPreference#DEFAULT} keeps the RFC 6724 sorted order intact (no family bias).
+ * {@link ProtocolFamilyPreference#INTERLEAVE} interleaves IPv6 and IPv4 addresses (v6, v4, v6, …),
+ * preserving the relative order within each family as produced by RFC 6724 sorting.
+ * </p>
+ *
+ * @since 5.7
+ */
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+public final class Rfc6724AddressSelectingDnsResolver implements DnsResolver {
```
rschmitt: I'm not crazy about the name -- for one thing, it's obviously cryptic; for another, this RFC could be superceded by a newer one (as has happened multiple times with Happy Eyeballs and HTTP itself), and we'll be stuck with the old RFC number as part of our public API symbols

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+    }
+
+    @Override
+    public InetAddress[] resolve(final String host) throws UnknownHostException {
+        final InetAddress[] resolved = delegate.resolve(host);
+
+        if (resolved == null) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("{} resolved '{}' -> null", simpleName(), host);
+            }
```
rschmitt, line 128: Why pass `simpleName()` here? The logger configuration should add the class name

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+            LOG.trace("{} delegate returned {} addresses for '{}': {}", simpleName(), resolved.length, host, fmt(resolved));
+        }
+
+        final List<InetAddress> candidates = filterCandidates(resolved, familyPreference);
+
+        if (candidates.isEmpty()) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("{} resolved '{}' -> []", simpleName(), host);
+            }
+            return new InetAddress[0];
```
rschmitt: Why an empty array here, but `null` on line 128?

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+        // HTTP/TCP is for unicast destinations; multicast is not a valid connect target.
+        if (ip.isMulticastAddress()) {
+            return false;
+        }
+        return true;
+    }
+
+    private static List<InetAddress> filterCandidates(
+            final InetAddress[] resolved,
+            final ProtocolFamilyPreference pref) {
```
rschmitt, line 201: This could be nicely implemented using streaming operations: `Arrays.stream(resolved).filter(a -> isUsableDestination(a)).filter(pref::shouldInclude).collect(toList())`

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
rschmitt: I'd still like to see this change implemented so I don't have to worry about the possibility of `ArrayIndexOutOfBoundsException` or of `src` and `dst` getting misaligned somehow

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
rschmitt: Comment still applies

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+    private List<InetAddress> inferSourceAddresses(final List<InetSocketAddress> destinations) {
+        final List<InetAddress> srcs = new ArrayList<>(destinations.size());
+
+        for (final InetSocketAddress dest : destinations) {
+            InetAddress src = null;
+            try {
+                src = sourceAddressResolver.resolveSource(dest);
+            } catch (final SocketException ignore) {
+                if (LOG.isTraceEnabled()) {
+                    LOG.trace("RFC6724 could not infer source address for {}: {}", dest, ignore.toString());
```
rschmitt: I realize the exception is being suppressed, but I don't think you should call this `ignore` unless it's literally a dead store

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
rschmitt: This should be dead code now, right? Don't these get filtered out now before this method is called?

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+                }
+            }
+            srcs.add(src);
+        }
+
+        if (LOG.isTraceEnabled()) {
+            final List<String> printable = new ArrayList<>(srcs.size());
+            for (final InetAddress a : srcs) {
+                printable.add(addr(a));
+            }
```
rschmitt: Looks like duplicate code; just call `fmt`

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+        return out;
+    }
+
+    private static List<String> fmt(final List<InetAddress> arr) {
+        final List<String> out = new ArrayList<>(arr.size());
+        for (final InetAddress a : arr) {
+            out.add(addr(a));
+        }
+        return out;
+    }
```
rschmitt: You could consolidate these by changing the method signature to take `Stream<InetAddress>`, which can be obtained by calling `list.stream()` or `Arrays.stream(addrs)`

httpclient5/src/main/java/org/apache/hc/client5/http/AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+                case 0x4: {
+                    return ADMIN_LOCAL;
+                }
+                case 0x5: {
+                    return SITE_LOCAL;
+                }
+                case 0x8: {
+                    return ORG_LOCAL;
+                }
+                default: {
```
rschmitt: You may consider changing this to `case 0xe` and having the `default` branch throw `IllegalArgumentException` or `AssertionError`

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
```diff
@@ -0,0 +1,670 @@
...
+        int i = 0;
+        int j = 0;
+        while (i < first.size() || j < second.size()) {
+            if (i < first.size()) {
+                out.add(first.get(i++));
+            }
+            if (j < second.size()) {
+                out.add(second.get(j++));
+            }
+        }
```
rschmitt: I think this would be a little cleaner with `Iterator`s. It saves you from having to index into arrays (I'm a little neurotic about that, I realize), and it may help optimize out runtime array bounds checks.

httpclient5/src/test/java/org/apache/hc/client5/http/examples/Rfc6724ResolverExample.java
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.hc.client5.http.examples;
+
+import java.net.InetAddress;
+
+import org.apache.hc.client5.http.DnsResolver;
+import org.apache.hc.client5.http.Rfc6724AddressSelectingDnsResolver;
+import org.apache.hc.client5.http.SystemDefaultDnsResolver;
+import org.apache.hc.client5.http.config.ProtocolFamilyPreference;
+
+public final class Rfc6724ResolverExample {
```
rschmitt:
```
src/hc/client # ./run-example.sh Rfc6724ResolverExample www.amazon.com
Host: www.amazon.com
Preference: DEFAULT
  IPv6 2600:1409:9800:1680:0:0:0:3bd4
  IPv6 2600:1409:9800:168d:0:0:0:3bd4
  IPv4 184.27.192.135

src/hc/client # ./run-example.sh Rfc6724ResolverExample www.amazon.com INTERLEAVE
Host: www.amazon.com
Preference: INTERLEAVE
  IPv6 2600:1409:9800:168d:0:0:0:3bd4
  IPv4 184.27.192.135
  IPv6 2600:1409:9800:1680:0:0:0:3bd4
```

Neat! I'd be cool if I could see the before-and-after (maybe you could intercept the "before" by wrapping the system default resolver, then passing the wrapper in to the RFC 6724 resolver), but it's not essential.

httpclient5/src/test/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolverTest.java
```diff
@@ -0,0 +1,385 @@
...
+        final Class<?> resolverClass = Rfc6724AddressSelectingDnsResolver.class;
+
+        final Method addr = resolverClass.getDeclaredMethod("addr", InetAddress.class);
+        addr.setAccessible(true);
+
+        final Method fmtArr = resolverClass.getDeclaredMethod("fmt", InetAddress[].class);
+        fmtArr.setAccessible(true);
+
+        final Method fmtList = resolverClass.getDeclaredMethod("fmt", List.class);
+        fmtList.setAccessible(true);
```
rschmitt: Just declare these methods package-private and call them directly!

httpclient5/src/test/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolverTest.java
rschmitt: Do we need any additional test coverage for the prefix matching stuff? That seems like quite a bit of non-trivial code

httpclient5-testing/src/test/java/org/apache/hc/client5/testing/ManualRfc6724ResolverIT.java
```diff
@@ -0,0 +1,87 @@
...
+import java.net.Inet6Address;
+import java.net.InetAddress;
+
+import org.apache.hc.client5.http.DnsResolver;
+import org.apache.hc.client5.http.Rfc6724AddressSelectingDnsResolver;
+import org.apache.hc.client5.http.SystemDefaultDnsResolver;
+import org.apache.hc.client5.http.config.ProtocolFamilyPreference;
+import org.junit.jupiter.api.Test;
+
+class ManualRfc6724ResolverIT {
```
rschmitt: What does this actually test? There are no assertions. If this basically just repeats the `examples` class then I'd rather not add it.

rschmitt:
There are some things from my first review that still haven't been addressed. I suggest going through the comments one by one and hitting "Resolve conversation" as you address each one, so you can track what is and isn't done.

httpclient5/src/main/java/org/apache/hc/client5/http/Rfc6724AddressSelectingDnsResolver.java
arturobernalg: I prefer keeping null only when the delegate returns null, and returning an empty array when all resolved candidates are filtered out to avoid introducing NPEs for callers.

rschmitt: I'm not sure that it's sensible for a resolver to return `null` or an empty collection as opposed to throwing an exception. But for now, the contract is to return `null`:

```
    /**
     * Returns the IP address for the specified host name, or null if the given
     * host is not recognized or the associated IP address cannot be used to
     * build an InetAddress instance.
     *
     * @see InetAddress
     *
     * @param host
     *            The host name to be resolved by this resolver.
     * @return The IP address associated to the given host name, or null if the
     *         host name is not known by the implementation class.
     */
    InetAddress[] resolve(String host) throws UnknownHostException;
```

arturobernalg:
> There are some things from my first review that still haven't been addressed. I suggest going through the comments one by one and hitting "Resolve conversation" as you address each one, so you can track what is and isn't done.

Hi @rschmitt  I believe I’ve addressed all your remarks in the latest update.

arturobernalg:
Hi @rschmitt ,                               
                                                                                                                                                                                                                                                                                           
  All your review comments are addressed. Additional fixes since last time:                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                           
  - IPv4 scope mapping fixed per RFC 6724 (127/8, 169.254/16 → link-local; everything else → global)                                                                                                                                                                                       
  - ::1 → link-local (interface-local is multicast-only)
  - Rule 9 now applies to IPv4 pairs too                                                                                                                                                                                                                                                   
  - No early return before filtering (single-address edge case)                                                                                                                                                                                                                            
  - resolve(host, port) throws instead of returning unresolved (prevents bypassing family filtering)                                                                                                                                                                                       
  - Rule 9 documented as heuristic (JDK limitation)

PR #781, opened 2026-01-11
HTTPCLIENT-1165: Add optional request collapsing for async cache misses

arturobernalg:
This change adds an optional request collapsing mechanism to the async caching executor to prevent multiple concurrent requests from fetching the same cacheable resource on a cold miss. When enabled, a single leader request is allowed to proceed to the backend while followers wait and then re-check the cache once the leader completes. The feature is disabled by default and is covered by unit tests for the collapser and an integration-style test validating that backend requests are reduced to one under concurrency.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
@@ -58,6 +58,7 @@ import org.apache.hc.client5.http.validator.ETag;
 import org.apache.hc.core5.annotation.Contract;
 import org.apache.hc.core5.annotation.ThreadingBehavior;
 import org.apache.hc.core5.concurrent.CancellableDependency;
+import org.apache.hc.core5.concurrent.ComplexCancellable;
 import org.apache.hc.core5.concurrent.ComplexFuture;
 import org.apache.hc.core5.concurrent.FutureCallback;
 import org.apache.hc.core5.http.ContentType;
@@ -98,6 +99,8 @@ class AsyncCachingExec extends CachingExecBase implements AsyncExecChainHandler 
     private final HttpAsyncCache responseCache;
     private final DefaultAsyncCacheRevalidator cacheRevalidator;
     private final ConditionalRequestBuilder<HttpRequest> conditionalRequestBuilder;
+    private final boolean requestCollapsingEnabled;
+    private final CacheRequestCollapser collapser;
 
     AsyncCachingExec(final HttpAsyncCache cache, final DefaultAsyncCacheRevalidator cacheRevalidator, final CacheConfig config) {
         super(config);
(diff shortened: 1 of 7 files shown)


ok2c:
@arturobernalg There is a lot of complexity in this change-set for what I personally see is little practical gain. 

Question. How well is this tested? How sure you are this is going to work reliably?

arturobernalg:
> @arturobernalg There is a lot of complexity in this change-set for what I personally see is little practical gain.
> 
> Question. How well is this tested? How sure you are this is going to work reliably?

@ok2c 
I added additional coverage for the unsafe paths: leader failure, leader cancellation, non-cacheable responses, Vary, request entities and only-if-cached. I also kept request collapsing disabled by default and scoped to each caching executor.
The local smoke example produces 20 origin requests without collapsing and 1 with collapsing enabled for concurrent cold misses.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
```diff
@@ -98,13 +99,25 @@ class AsyncCachingExec extends CachingExecBase implements AsyncExecChainHandler
...
 
     AsyncCachingExec(final HttpAsyncCache cache, final DefaultAsyncCacheRevalidator cacheRevalidator, final CacheConfig config) {
+        this(cache, cacheRevalidator, config, false);
+    }
+
+    AsyncCachingExec(
+            final HttpAsyncCache cache,
+            final DefaultAsyncCacheRevalidator cacheRevalidator,
+            final CacheConfig config,
+            final boolean requestCollapsingEnabled) {
```
ok2c: @arturobernalg I would move `requestCollapsingEnabled` to `CacheConfig` and documented it being a hint. Some implementations (like async) may choose to respect it, some may not.

ok2c (changes requested):

httpclient5-cache/src/test/java/org/apache/hc/client5/http/cache/example/AsyncClientCacheRequestCollapsing.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.cache.CacheConfig;
+import org.apache.hc.client5.http.impl.cache.CachingHttpAsyncClientBuilder;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.io.CloseMode;
+
+/**
+ * Demonstrates optional request collapsing for concurrent async cache misses.
+ */
+public class AsyncClientCacheRequestCollapsing {
```
ok2c, line 55: @arturobernalg Your examples tend to look like integration tests. This is not the point of examples. They need to demonstrate the use of APIs abd be informative. Please drop the server side code and just focus on the client APIs.

ok2c (approved):

PR #807, opened 2026-02-24
Validate Connection: Upgrade on protocol switch responses

arturobernalg:
ProtocolSwitchStrategy now rejects 101 Switching Protocols responses that advertise Upgrade without the required Connection: Upgrade token.

[RFC 2817 notes:](https://datatracker.ietf.org/doc/html/rfc2817#section-3.1)
"the upgrade keyword MUST be supplied within a Connection header field (section 14.10) whenever Upgrade is present in an HTTP/1.1 message". :contentReference[oaicite:0]{index=0}

[RFC 9110 requires](https://www.rfc-editor.org/rfc/rfc9110.html#name-upgrade):
"A sender of Upgrade MUST also send an "Upgrade" connection option in the Connection header field (Section 7.6.1)". :contentReference[oaicite:1]{index=1}

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
@@ -26,6 +26,7 @@
  */
 package org.apache.hc.client5.http.impl;
 
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.hc.core5.annotation.Internal;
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/ProtocolSwitchStrategy.java
```diff
@@ -87,6 +91,49 @@ public ProtocolVersion switchProtocol(final HttpMessage response) throws Protoco
         }
     }
 
+    private boolean containsConnectionUpgrade(final HttpMessage message) throws ProtocolException {
```
ok2c: @arturobernalg  Use `MessageSupport#parseHeaders` or `MessageSupport#parseTokens` instead.

ok2c (changes requested):

ok2c (approved):

PR #810, opened 2026-03-06
HTTPCLIENT-2261 - Enable HTTP/2 CONNECT tunneling for HTTP/2 clients …

arturobernalg:
…through HTTP/2 proxies

Wire HTTP/2 tunnel establishment into InternalH2ConnPool for tunneled routes by using H2OverH2TunnelSupport to convert an existing proxy HTTP/2 connection into a stream-backed tunnel session


arturobernalg:
We need this first https://github.com/apache/httpcomponents-core/pull/648

ok2c:
@arturobernalg This is a log of code. We need a compatibility test case similar to `ApacheHTTPDSquidCompatibilityIT` to be sure it actually works correctly. I am not entirely sure this is all worth the efforts.

arturobernalg:
> @arturobernalg This is a log of code. We need a compatibility test case similar to `ApacheHTTPDSquidCompatibilityIT` to be sure it actually works correctly. I am not entirely sure this is all worth the efforts.

@ok2c  add it. please do another passs

PR #815, opened 2026-03-26
Named Pipe transport support for Windows   

arturobernalg:
Added Windows Named Pipe transport as the Windows counterpart to Unix Domain Socket support introduced earlier. This enables HttpClient to communicate over Named Pipes such as \\.\pipe\docker_engine using the same connection management and request execution infrastructure.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -84,3 +84,34 @@ jobs:
         EOF
     - name: Build with Maven
       run: ./mvnw -V --file pom.xml --no-transfer-progress -DtrimStackTrace=false -Djunit.jupiter.execution.parallel.enabled=false -Dhc.build.toolchain.version="${HC_BUILD_TOOLCHAIN_VERSION}" -Pdocker
+
+  # Windows Named Pipe tests — requires Windows for \\.\pipe\... support
+  windows-npipe:
+
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        java: [ 17, 21 ]
+      fail-fast: false
+
+    steps:
+    - uses: actions/checkout@v5
+    - uses: actions/cache@v4
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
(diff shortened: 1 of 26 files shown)


rschmitt:
We'd be crossing a Rubicon by exposing support for a platform-specific IO primitive via JNI. I don't think we've done that before. I'd rather have a story for improved custom transport support, which I think would be more sustainable than threading through yet more special cases. `docker-java` implemented both UDS and named pipe transports via `DetachedSocketFactory`, and the main issue with that approach is that it requires the use of internal/experimental APIs; we would solve more problems by stabilizing that extension point, and offering a similar one in the async client (the `java.io.Socket` API is itself legacy).

If we were to implement this as a special case/first-class feature, I'd like to consider doing it on top of the Unix domain socket support. As far as I can tell, the named pipe support in `docker-java` is a legacy feature and an obsolete workaround for the lack of UDS support in Windows. It might be reasonable to add anyway, but it's not obvious that there's a lot of demand for it.

arturobernalg:
> We'd be crossing a Rubicon by exposing support for a platform-specific IO primitive via JNI. I don't think we've done that before. I'd rather have a story for improved custom transport support, which I think would be more sustainable than threading through yet more special cases. `docker-java` implemented both UDS and named pipe transports via `DetachedSocketFactory`, and the main issue with that approach is that it requires the use of internal/experimental APIs; we would solve more problems by stabilizing that extension point, and offering a similar one in the async client (the `java.io.Socket` API is itself legacy).
> 
> If we were to implement this as a special case/first-class feature, I'd like to consider doing it on top of the Unix domain socket support. As far as I can tell, the named pipe support in `docker-java` is a legacy feature and an obsolete workaround for the lack of UDS support in Windows. It might be reasonable to add anyway, but it's not obvious that there's a lot of demand for it.

The real issue here is not named pipes per se, but whether HttpClient should grow another transport-specific request option.
I used Unix domain socket support as the closest existing precedent, but a generic transport extension point for both classic and async clients would be cleaner.

I am happy to rework the proposal in that direction if everyone is agree.

PR #823, opened 2026-04-16
HTTPCLIENT-2418 - Fix SimpleBody default charset handling for JSON content

arturobernalg:
SimpleBody currently falls back to ASCII when Content-Type is application/json without an explicit charset, which breaks UTF-8 payloads.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/SimpleBody.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/SimpleBody.java
@@ -55,8 +55,8 @@ public final class SimpleBody {
         if (body.length() > 2048) {
             return new SimpleBody(null, body, contentType);
         }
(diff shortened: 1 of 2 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/SimpleBody.java
```diff
@@ -50,13 +50,31 @@ public final class SimpleBody {
...
     }
 
+    private static Charset getCharset(final ContentType contentType) {
+        if (contentType != null) {
+            final Charset charset = contentType.getCharset();
+            if (charset != null) {
+                return charset;
+            }
+            final String mimeType = contentType.getMimeType();
+            if (contentType.isSameMimeType(ContentType.APPLICATION_JSON)
```
ok2c: @arturobernalg Oh, god. This is ugly. Just use always UTF-8 and be done with it.

ok2c (approved):

PR #830, opened 2026-05-19
HTTPCLIENT-2421: Use effective retry interval for response timeout check

arturobernalg:
Use the effective response retry interval when checking the response timeout. 
The retry interval may be derived from the Retry-After response header, not only from the default retry interval.


davsclaus (approved):

ok2c (approved):

arturobernalg:
@ok2c should i cherry-pick to 5.6.x??

ok2c:
@arturobernalg This represents a behavioral change. I would rather not.

PR #832, opened 2026-05-23
REST interface scanner code improvement

ok2c:
@arturobernalg I reworked some of your REST interface scanning code and made it use core parsing and formatting routines where available. Please review.


arturobernalg:
> @arturobernalg I reworked some of your REST interface scanning code and made it use core parsing and formatting routines where available. Please review.

@ok2c 
Looks good overall.

Should this use `producesContentTypes` instead? @Produces maps to Accept; @Consumes should only drive the request entity Content-Type.

ok2c:
@arturobernalg I suspect the current implementation of `@Produces` and `@Consumes` is wrong but this is a subject for a subsequent change-set. I wanted to make sure all your original tests still passed after the refactoring.

arturobernalg (approved):

PR #839, opened 2026-06-07
HTTPCLIENT-2423: Propagate trace context in observation interceptors

arturobernalg:
Expose outbound requests as Micrometer sender carriers for header injection.


philsttr (approved):
Looks great!  Thanks @arturobernalg

PR #840, opened 2026-06-16
Add support for HTTP QUERY method

desiderantes:
After the changes were merged in https://github.com/apache/httpcomponents-core/pull/499 and with the changes in https://github.com/apache/httpcomponents-core/pull/673 this PR adds some missing pieces for HTTP Query method support, as defined on RFC 10008. I've omitted other possible locations to improve because they're marked as deprecated.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
@@ -29,6 +29,8 @@ package org.apache.hc.client5.http.impl.cache;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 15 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
```diff
@@ -192,13 +216,56 @@ public String generateKey(final URI requestUri) {
...
+        final String hash = Method.QUERY.isSame(request.getMethod()) ? getBodyHash(request) : null;
         try {
-            return generateKey(new URI(s));
+            return generateKey(new URI(s), hash);
         } catch (final URISyntaxException ex) {
             return s;
         }
     }
 
+    private static byte[] getRequestBodyBytes(final HttpRequest request) {
```
ok2c: @desiderantes I do not think this is how we should be doing it. This is not going to work well with the async caching layer. We will likely need to extend or to replace `CacheKeyGenerator` interface in order to solve the problem properly. 

As far as I understand QUERY method caching is not a hard requirement. Please remove caching logic from the change-set for now. It can be introduced at a later point with a separate pull request.

ok2c (changes requested):

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
desiderantes: I've reverted the changes. I'll come back to the caching side after this PR gets merged.

PR #845, opened 2026-07-05
escape quotes and strip line breaks in rfc7578 multipart parameters

dxbjavid:
In EXTENDED mode the RFC 7578 writer builds the Content-Disposition line itself rather than going through MimeField.getBody(), and for parameters other than filename (in practice the form field name) it writes the value straight out with no quoting. So a field name that contains a double quote or a CR/LF ends up breaking out of the quoted string and injecting extra header lines into the part, which matters when the name comes from somewhere untrusted. The STRICT and LEGACY writers already avoid this because they escape quotes and strip line breaks; this just brings the RFC 7578 path in line by backslash-escaping quotes and replacing line breaks the same way. Added a small regression test in HttpRFC7578MultipartTest covering a name with an embedded quote and CRLF.


ok2c (approved):

dxbjavid:
TY!

PR #846, opened 2026-07-08
Add QUERY method caching support

desiderantes:
This part was dropped from #840 but I reworked it based on the feedback I got. Now CacheKeyGenerator is not a singleton but a regular class that can receive a body extraction function.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
@@ -40,6 +40,16 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.stream.StreamSupport;
 
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.function.Function;
+import org.apache.hc.client5.http.async.methods.SimpleHttpRequest;
+import org.apache.hc.core5.http.ClassicHttpRequest;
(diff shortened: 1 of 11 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
```diff
@@ -138,13 +173,62 @@ public String generateKey(final URI requestUri) {
...
+                    if (entity.isRepeatable()) {
+                        final byte[] bytes = EntityUtils.toByteArray(entity);
+                        ((ClassicHttpRequest) request).setEntity(new ByteArrayEntity(bytes, ContentType.parse(entity.getContentType())));
+                        return bytes;
+                    }
+                } catch (final Exception ignore) {
+                }
+            }
+        }
+        return null;
```
ok2c, line 207: @desiderantes I apologize it took me to long to get back to you. 

Your approach is not unreasonable but it can still lead to cache entries with partial / broken keys, if the content body they are enclosing cannot be extracted. The key generator _must_ always produce a valid key. This method _must_ not return null.

What we need to do instead is to ensure those requests that cannot be correctly represented as `SimpleHttpRequest` get routed along the execution chain bypassing the caching layer.

Please as the first step review my pull request #850

desiderantes: ok, yes, if null is not an option, this doesn't work. I'll take a look at that PR

desiderantes:
Closed in favour of #852

PR #847, opened 2026-07-12
Harden WebSocket transport and permessage-deflate for RFC 6455/7692 conformance

arturobernalg:
Fixed permessage-deflate fragmentation and the SYNC_FLUSH drain in both stacks, derived the client masking key from SecureRandom, and failed the connection on RSV1 control frames and on a data frame received mid-fragment.
  Bounded the HTTP/2 server inbound window and outbound queue, rejected a 64-bit length with the MSB set, enforced the 125-byte control frame limit, and truncated the CLOSE reason.
  Deduplicated extension offers by name, released the Deflater and Inflater on session end, accepted any 2xx Extended CONNECT response, and guarded the H2 transport against a cancelled selection key on shutdown.
  Added tests for each; the module suite passes and checkstyle is clean.


ok2c:
@arturobernalg Once you merge this change-set I will start with HttpClient 5.7-alpha1 release

PR #850, opened 2026-07-15
Added support for request body to the cache key generator

ok2c:
This change-set extends the cache key generator with support for requests with an enclosed content body such as QUERY.

@desiderantes Please review and let me know if this approach works for you.

@arturobernalg Please double-check.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
@@ -302,7 +302,7 @@ class AsyncCachingExec extends CachingExecBase implements AsyncExecChainHandler 
                 final CacheHit root = result != null ? result.root : null;
                 if (hit == null) {
                     if (requestCollapsingEnabled && root == null && !requestCacheControl.isOnlyIfCached()) {
-                        final String cacheKey = CacheKeyGenerator.INSTANCE.generateKey(target, cacheRequest);
+                        final String cacheKey = CacheKeyGenerator.INSTANCE.generateKey(target, cacheRequest, SimpleHttpRequest::getBodyBytes);
                         final CacheRequestCollapser.Token token = collapser.enter(cacheKey);
                         if (token.isLeader()) {
                             handleCacheMiss(requestCacheControl, null, target, cacheRequest, scope, chain, new AsyncExecCallback() {
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
@@ -38,6 +38,7 @@ import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.stream.StreamSupport;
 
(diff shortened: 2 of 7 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
```diff
@@ -135,9 +140,33 @@ public String generateKey(final URI requestUri) {
...
+     * @return cache key
+     *
+     * @since 5.7
+     */
+    public <T extends HttpRequest> String generateKey(final HttpHost host,
+                                                      final T request,
+                                                      final Function<T, String> bodyExtractor) {
         final String s = CacheSupport.requestUriRaw(host, request);
+        final String body = bodyExtractor != null ? bodyExtractor.apply(request) : null;
+        if (body != null) {
```
desiderantes, line 167: Is it supposed to fail now if the body extractor succeeds?

ok2c: @desiderantes This is a temporary sanity check as the caching layer simply does not work correctly with entity enclosing requests. This is something I presume you are going to fix in your pull request.

desiderantes: good, just wanted to confirm

desiderantes (approved):
I think it's solid and simple, thank you! I'll rebase my PR and simplify it after this lands.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
```diff
@@ -302,7 +302,7 @@ public void completed(final CacheMatch result) {
                 final CacheHit root = result != null ? result.root : null;
                 if (hit == null) {
                     if (requestCollapsingEnabled && root == null && !requestCacheControl.isOnlyIfCached()) {
-                        final String cacheKey = CacheKeyGenerator.INSTANCE.generateKey(target, cacheRequest);
+                        final String cacheKey = CacheKeyGenerator.INSTANCE.generateKey(target, cacheRequest, SimpleHttpRequest::getBodyText);
```
arturobernalg: Would it make more sense to use byte[] here and SimpleHttpRequest::getBodyBytes at the call sites? getBodyText() can collapse different binary or invalid byte sequences into the same String, which would make the eventual cache key ambiguous.

desiderantes: hmm... it is also possible that subclasses of Request might have a certain representation for certain content types (like a sorted object/array for json, or ordered form fields, etc), which might affect the caching as different bodies should be cached because they converge to the same canonical representation

arturobernalg: That still does not require String. A request implementation could canonicalize the content and return the resulting bytes. That preserves the option for content-type-specific normalization without forcing arbitrary content through text decoding.

desiderantes: true, not all content is text-like, sorry, this is a good suggestion

ok2c: @desiderantes @arturobernalg I can hardly imagine a QUERY request with binary query content but so be it. Please do another pass.

desiderantes (approved):

PR #852, opened 2026-07-16
HTTP cache support for the QUERY method (RFC 10008)

hunghhdev:
Follow-up to #840. The caching layer can now store and serve responses to `QUERY` requests: cache keys incorporate a digest of the request content and content type, and both classic and async interceptors buffer `QUERY` bodies instead of bypassing the cache. Anything that cannot be buffered safely still falls back to direct execution.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
@@ -72,10 +72,12 @@ import org.apache.hc.core5.http.HttpHost;
 import org.apache.hc.core5.http.HttpRequest;
 import org.apache.hc.core5.http.HttpResponse;
 import org.apache.hc.core5.http.HttpStatus;
+import org.apache.hc.core5.http.Method;
 import org.apache.hc.core5.http.impl.BasicEntityDetails;
 import org.apache.hc.core5.http.nio.AsyncDataConsumer;
 import org.apache.hc.core5.http.nio.AsyncEntityProducer;
 import org.apache.hc.core5.http.nio.CapacityChannel;
+import org.apache.hc.core5.http.nio.DataStreamChannel;
 import org.apache.hc.core5.http.nio.entity.BasicAsyncEntityProducer;
 import org.apache.hc.core5.http.nio.entity.StringAsyncEntityProducer;
 import org.apache.hc.core5.net.URIAuthority;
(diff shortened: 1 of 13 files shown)


ok2c:
@hunghhdev Awesome! Out of curiosity is this effort coordinated with the original author of #840 or is it purely coincidental? 

Please note there is #850 that still needs to land that improves the process of cache key generation. 

Overall looks very good.

desiderantes:
> @hunghhdev Awesome! Out of curiosity is this effort coordinated with the original author of #840 or is it purely coincidental? 
> 
> Please note there is #850 that still needs to land that improves the process of cache key generation. 
> 
> Overall looks very good.

No, this PR was not coordinated, I just saw this. If you prefer this code instead, I can stop #846.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
```diff
@@ -412,12 +414,78 @@ public void cancelled() {
...
+        return null;
+    }
+
+    /**
+     * Drains the given repeatable entity producer into a byte array without an I/O reactor.
+     * Returns {@code null} if the producer fails to make progress, in which case it is reset
+     * by {@link AsyncEntityProducer#releaseResources()} and can still be used to execute
+     * the request directly.
+     */
+    private static byte[] drainEntityProducer(final AsyncEntityProducer entityProducer) throws IOException {
```
ok2c, line 446: @hunghhdev When draining the entity producer we should enforce a total cap on body length in order to avoid some "security professional" reporting it as a potential DoS vulnerability. I imagine requests with content body larger than 25K should be considered non-cacheable

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExec.java
```diff
@@ -221,12 +222,21 @@ private static ClassicHttpResponse convert(final SimpleHttpResponse cacheRespons
...
-        if (request.getEntity() != null) {
-            return null;
+    SimpleHttpRequest prepareRequest(final ClassicHttpRequest request) throws IOException {
+        final HttpEntity entity = request.getEntity();
+        if (entity == null) {
+            return SimpleRequestBuilder.copy(request).build();
+        }
+        // QUERY content must be read in full in order to determine the cache key
+        if (Method.QUERY.isSame(request.getMethod()) && entity.getContentEncoding() == null) {
+            final byte[] content = EntityUtils.toByteArray(entity);
```
ok2c: @hunghhdev Likewise, we should avoid buffering of content body larger than 25K. That also makes it necessary for the content entity to be repeatable.

ok2c (changes requested):

ok2c:
> No, this PR was not coordinated, I just saw this. If you prefer this code instead, I can stop #846.

@desiderantes This is the first time we, as a project, have multiple contributions of the same feature. I suggest we proceed with @hunghhdev contribution as it is almost complete and it is a first-time contribution. In the meantime, if you have time and inclination what also needs to be done is to extend `HttpCacheEntry` and the related factory and serializer to store request body along with the request head.

ok2c:
@hunghhdev Very good. Please re-base off the latest master and update the cache key generation logic

hunghhdev:
> @hunghhdev Very good. Please re-base off the latest master and update the cache key generation logic

Rebased off the latest master

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheKeyGenerator.java
```diff
@@ -161,17 +170,39 @@ public String generateKey(final HttpHost host, final HttpRequest request) {
      */
     public <T extends HttpRequest> String generateKey(final HttpHost host,
                                                       final T request,
-                                                      final Function<T, byte[]> bodyExtractor) {
+                                                      final Function<T, SimpleBody> bodyExtractor) {
```
ok2c: @hunghhdev I would like to avoid a direct coupling with the `Simple*` APIs because it is basically an intermediate layer intended to help transition from the classic i/o to  the async i/o. The project actually strongly discourages its use. I may end up replacing it with the cache specific internal classes in the future.

If you want to include the content-type in the content digest you can get it from the request object as a header value

ok2c (changes requested):

ok2c (approved):
@hunghhdev Looks good. I will merge this change-set tomorrow if I hear no objections.

ok2c:
@hunghhdev Many thanks for this contribution! There is still one more thing that needs to be done: `HttpCacheEntry` and related classes should be extended to include the request body if given. Unless @desiderantes has already started working on this change.

hunghhdev:
> @hunghhdev Many thanks for this contribution! There is still one more thing that needs to be done: `HttpCacheEntry` and related classes should be extended to include the request body if given. Unless @desiderantes has already started working on this change.

Thanks for merging! I'll take the HttpCacheEntry extension as a follow-up. @desiderantes hope you don't mind me picking this one up too

desiderantes:
> > @hunghhdev Many thanks for this contribution! There is still one more thing that needs to be done: `HttpCacheEntry` and related classes should be extended to include the request body if given. Unless @desiderantes has already started working on this change.
> 
> Thanks for merging! I'll take the HttpCacheEntry extension as a follow-up. @desiderantes hope you don't mind me picking this one up too 

I have no strong feelings, one way or the other

PR #853, opened 2026-07-16
Do not call the backend when an only-if-cached request cannot be served from cache (RFC 9111 5.2.1.7)

hunghhdev:
`AsyncCachingExec.handleCacheMiss` is missing a `return` after generating the 504 (Gateway Timeout) for an `only-if-cached` request, so it falls through and calls the backend anyway. Added the missing `return` and a test, mirroring the classic `CachingExec` behavior


ok2c:
Cherry-picked to 5.6.x

PR #859, opened 2026-07-22
Extended HttpCacheEntry and related classes to store the enclosed request body

hunghhdev:
Follow-up to #852: `HttpCacheEntry`, its factory and the entry serializer now store the enclosed request body along with the request head.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheEntryFactory.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheEntryFactory.java
@@ -130,6 +130,7 @@ public class HttpCacheEntryFactory {
                 latestVariant.getRequestMethod(),
                 latestVariant.getRequestURI(),
                 headers(latestVariant.requestHeaderIterator()),
+                latestVariant.getRequestContent(),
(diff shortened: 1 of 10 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheEntryFactory.java
```diff
@@ -152,6 +153,27 @@ public HttpCacheEntry create(final Instant requestInstant,
...
+     * Create a new {@link HttpCacheEntry} with the given request content and {@link Resource}.
+     *
+     * @param requestInstant   Date/time when the request was made (Used for age calculations)
+     * @param responseInstant  Date/time that the response came back (Used for age calculations)
+     * @param host             Target host
+     * @param request          Original client request (a deep copy of this object is made)
+     * @param requestContent   Content enclosed in the original client request or {@code null}
+     * @param response         Origin response (a deep copy of this object is made)
+     * @param resource         Resource representing origin response body
+     */
```
ok2c, line 171: @hunghhdev Please add `@since 5.7` tags to all new methods. Looks very good otherwise.

ok2c (approved):

PR #860, opened 2026-07-24
  HTTPCLIENT-2428: adjust reactor capacity in inflating AsyncDataConsumers 

arturobernalg:
Inflating consumers forwarded the downstream capacity (decompressed bytes) to the reactor unchanged, so decompression could expand it and overshoot the requested capacity.
They now wrap the `CapacityChannel` and scale the increment down by a fixed expansion factor before forwarding it.
`InflatingBrotliDataConsumer` also stops overriding the downstream capacity with `Integer.MAX_VALUE`, which defeated back-pressure.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/InflatingCapacityChannel.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/InflatingCapacityChannel.java
@@ -1,0 +1,74 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/InflatingCapacityChannel.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    InflatingCapacityChannel(final CapacityChannel delegate, final int expansionFactor) {
+        this.delegate = Args.notNull(delegate, "Capacity channel");
+        this.expansionFactor = Args.positive(expansionFactor, "Expansion factor");
+    }
+
+    @Override
+    public void update(final int increment) throws IOException {
+        if (increment <= 0) {
```
ok2c: @arturobernalg I would also make a special check and pass `Integer#MAX_VALUE` along without expansion factor.

ok2c (approved):

httpclient5/src/main/java/org/apache/hc/client5/http/async/methods/InflatingCapacityChannel.java
arturobernalg: @ok2c Done.   `Integer.MAX_VALUE` is now forwarded unchanged

ok2c (approved):

PR #863, opened 2026-07-28
reject control characters in bearer token in BearerScheme

dxbjavid:
BearerScheme writes the token straight into the Authorization header, and unlike BasicScheme it never checks the credential for control characters. A bearer token carrying CR/LF is neutralised by the HTTP/1.1 line formatter, which replaces those bytes with spaces, but the HTTP/2 request converter passes header values through unchanged, so over h2 such a token can inject extra header material into the request. This applies the same control-character guard BasicScheme already uses, rejecting the token before the header is built.


dxbjavid:
TY!

PR #864, opened 2026-08-01
Mark the SSE and WebSocket modules as experimental

arturobernalg:
Annotated the public API types of the httpclient5-sse and httpclient5-websocket modules with @Experimental. Both modules expose a large API surface that is not yet stable, and the annotation marks them as subject to change so that a relaxed compatibility policy applies for the 5.7 release series; the japicmp binary and source compatibility checks are already skipped for both modules. Types that are already marked @Internal are left unchanged.


ok2c:
@arturobernalg I imagine adding a note to the top level `package-info.java` about the module still considered experimental would be enough. I leave it up to you to decide if you want to mark every public class `@Experimenal` or just mark the top level javadoc.

ok2c (approved):

ok2c (approved):

PR #867, opened 2026-08-11
RFC 9211: optional Cache-Status response header

arturobernalg:
Adds optional support for the RFC 9211 Cache-Status response header to the caching executors.

When enabled, the cache records how each response was produced — a cache hit,cor a forward to the origin with the reason (bypass, method, uri-miss, vary-miss, request, stale) and the upstream status code where relevant — and emits a
Cache-Status header identified as Apache-HttpClient.

Disabled by default; enabled with `CacheConfig.setCacheStatusEnabled(true)`.
Applies to both the classic and async cache execution paths.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheContext.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheContext.java
@@ -42,6 +42,7 @@ import org.apache.hc.client5.http.cookie.CookieOrigin;
 import org.apache.hc.client5.http.cookie.CookieSpec;
 import org.apache.hc.client5.http.cookie.CookieSpecFactory;
 import org.apache.hc.client5.http.cookie.CookieStore;
+import org.apache.hc.client5.http.impl.cache.CacheStatus;
 import org.apache.hc.client5.http.protocol.HttpClientContext;
 import org.apache.hc.client5.http.protocol.RedirectLocations;
 import org.apache.hc.core5.annotation.Internal;
@@ -73,6 +74,7 @@ public class HttpCacheContext extends HttpClientContext {
     static final String REQUEST_CACHE_CONTROL = "http.cache.request-control";
     static final String RESPONSE_CACHE_CONTROL = "http.cache.response-control";
     static final String CACHE_ENTRY = "http.cache.entry";
+    static final String CACHE_STATUS = "http.cache.status";
 
     /**
      * @deprecated Use {@link #castOrCreate(HttpContext)}.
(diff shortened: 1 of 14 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachingExecBase.java
```diff
@@ -131,6 +130,41 @@ SimpleHttpResponse generateGatewayTimeout() {
         return SimpleHttpResponse.create(HttpStatus.SC_GATEWAY_TIMEOUT, "Gateway Timeout");
     }
 
+    /**
+     * Internal context attribute holding the per-exchange {@link CacheStatus} used to render the
+     * RFC 9211 {@code Cache-Status} header.
+     */
+    static final String CACHE_STATUS_ATTR = "http.cache.status";
```
ok2c: @arturobernalg Why? Why not making it a proper context attribute like `CacheResponseStatus`, `RequestCacheControl`, `ResponseCacheControl`, etc?

arturobernalg: @ok2c  Good point — done.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/HttpCacheContext.java
```diff
@@ -122,6 +124,7 @@ public static HttpCacheContext create() {
     private RequestCacheControl requestCacheControl;
     private ResponseCacheControl responseCacheControl;
     private HttpCacheEntry cacheEntry;
+    private CacheStatus cacheStatus;
```
ok2c, line 126: @arturobernalg `CacheResponseStatus` and `CacheStatus` now overlap. Is there any chance we could get rid of `responseStatus` attribute and generate its value from `cacheStatus` on the fly?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/AsyncCachingExec.java
```diff
@@ -293,6 +300,15 @@ public void doExecute(
             if (LOG.isDebugEnabled()) {
                 LOG.debug("{} request cannot be served from cache", exchangeId);
             }
+            // The cache lookup is bypassed here (non-cacheable method, or a no-store / no-cache
+            // request), so no stored response was ever selected; this is not fwd=request.
+            final String method = cacheRequest.getMethod();
+            final boolean cacheSupportedMethod = Method.GET.isSame(method)
```
ok2c: @arturobernalg Let's move this bit into `CacheSupport`. Looks good otherwise.

ok2c: @arturobernalg Please make  `CacheableRequestPolicy` and `ResponseCachingPolicy` use this method as well

ok2c (approved):

PR #868, opened 2026-08-12
Honor qualified Cache-Control: private="field" in a shared cache (RFC9111 section 5.2.2.7)

arturobernalg:
A shared cache now stores a response that carries a qualified private directive with the named header fields removed from the stored copy, instead of treating the whole response as `non-cacheable`, while the response returned to the caller retains those fields. The fields are removed from freshly stored entries, from entries updated by a 304 revalidation, and from the root entry of a Vary response. A bare private directive still makes the whole response `non-storable` by a shared cache, and multiple qualified private directives accumulate their field names.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
@@ -189,6 +189,20 @@ class CacheControlHeaderParser {
                 builder.setNoStore(true);
             } else if (name.equalsIgnoreCase(HeaderConstants.CACHE_CONTROL_PRIVATE)) {
                 builder.setCachePrivate(true);
+                if (value != null) {
+                    final Tokenizer.Cursor valCursor = new ParserCursor(0, value.length());
+                    final Set<String> privateFields = new HashSet<>();
+                    while (!valCursor.atEnd()) {
+                        final String token = tokenParser.parseToken(value, valCursor, VALUE_DELIMS);
+                        if (!TextUtils.isBlank(token)) {
+                            privateFields.add(token);
+                        }
+                        if (!valCursor.atEnd()) {
+                            valCursor.updatePos(valCursor.getPos() + 1);
+                        }
+                    }
+                    builder.setPrivateFields(privateFields);
+                }
             } else if (name.equalsIgnoreCase(HeaderConstants.CACHE_CONTROL_PROXY_REVALIDATE)) {
                 builder.setProxyRevalidate(true);
             } else if (name.equalsIgnoreCase(HeaderConstants.CACHE_CONTROL_PUBLIC)) {
--- a/httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/CacheControlParserTest.java
+++ b/httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/CacheControlParserTest.java
@@ -28,6 +28,7 @@ package org.apache.hc.client5.http.impl.cache;
 
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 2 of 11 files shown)


httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/CacheControlParserTest.java
```diff
@@ -207,6 +207,46 @@ void testParseNoCacheFieldsMessy() {
...
+        final Header header = new BasicHeader("Cache-Control", "private, s-maxage=3600");
+        final ResponseCacheControl cacheControl = parser.parseResponse(Collections.singletonList(header).iterator());
+
+        assertTrue(cacheControl.isCachePrivate());
+        assertTrue(cacheControl.getPrivateFields().isEmpty());
+    }
+
+    @Test
+    void testParseMultiplePrivateDirectivesAccumulate() {
+        final Header header = new BasicHeader("Cache-Control", "private=\"X-A\", private=\"X-B\", s-maxage=3600");
```
ok2c, line 234: @arturobernalg What part of the spec actually defines or recommends such behavior? In my opinion only `private=\"X-B\"` should have an effect.

httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/TestAsyncCachingExecPrivateFields.java
```diff
@@ -0,0 +1,108 @@
...
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpServer;
```
ok2c: @arturobernalg Why?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/BasicHttpCache.java
```diff
@@ -95,14 +111,44 @@ public BasicHttpCache() {
...
+            storage.putEntry(cacheKey, sharedCache ? stripPrivateFields(entry) : entry);
         } catch (final ResourceIOException ex) {
             if (LOG.isWarnEnabled()) {
                 LOG.warn("I/O error storing cache entry with key {}", cacheKey);
             }
         }
     }
 
+    static HttpCacheEntry stripPrivateFields(final HttpCacheEntry entry) {
+        final ResponseCacheControl cacheControl = CacheControlHeaderParser.INSTANCE.parse(entry);
```
ok2c: @arturobernalg This looks conceptually wrong. Why do we need to re-parse the headers and re-create `ResponseCacheControl` here? In fact this is the wrong architectural layer for such a decision. This is the responsibility of the protocol layer, not of a caching backend. Pull this logic up into `CachingExec` / `AsyncCachingExec`

ok2c (changes requested):

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -158,19 +174,40 @@ public final ResponseCacheControl parseResponse(final Iterator<Header> headerIte
...
+                        }
+                    }
                     builder.setNoCacheFields(noCacheFields);
                 }
             } else if (name.equalsIgnoreCase(HeaderConstants.CACHE_CONTROL_NO_STORE)) {
                 builder.setNoStore(true);
             } else if (name.equalsIgnoreCase(HeaderConstants.CACHE_CONTROL_PRIVATE)) {
                 builder.setCachePrivate(true);
+                // A repeated private directive replaces any earlier one; only the last takes effect.
+                privateFields.clear();
```
ok2c: @arturobernalg This whole thing can be condensed into just a few lines of code. Why the parsing logic should be any different that that of `CACHE_CONTROL_NO_CACHE`?

httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/TestAsyncCachingExecPrivateFields.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.hc.core5.http.protocol.HttpContext;
+import org.apache.hc.core5.io.CloseMode;
+import org.apache.hc.core5.reactor.ListenerEndpoint;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class TestAsyncCachingExecPrivateFields {
+
+    @Test
+    void testQualifiedPrivateFieldStrippedButRemainderCachedBySharedCache() throws Exception {
```
ok2c: @arturobernalg Does this need to be an integration test? Cannot be the same code be tested in `TestCachingExecChain` or `TestAsyncCachingExecChain`?

ok2c (changes requested):

ok2c:
@arturobernalg I am sorry but I am not in favor of this change. It introduces a lot of ugly complexity for a very little practical gain.

arturobernalg:
@ok2c 
Understood. Thanks for taking another look. I’ll close this one.

PR #869, opened 2026-08-13
RFC 6265bis compliance for the cookie implementation

arturobernalg:
Enforces the __`Secure- `and `__Host-` cookie name prefixes, adds the `SameSite` attribute with the `SameSite=None` requires Secure rule, includes the host-only flag in cookie identity, applies the 4096-byte cookie and 1024-byte attribute size limits and the 400-day maximum lifetime, supports nameless cookies in both parsing and serialization, and prevents a non-secure connection from overwriting a stored secure cookie.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/cookie/RFC6265CookieSpec.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/cookie/RFC6265CookieSpec.java
@@ -27,6 +27,7 @@
 
 package org.apache.hc.client5.http.impl.cookie;
 
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -66,6 +67,12 @@ public class RFC6265CookieSpec implements CookieSpec {
     private final static char EQUAL_CHAR = '=';
     private final static char DQUOTE_CHAR = '"';
     private final static char ESCAPE_CHAR = '\\';
+
+    private final static String SECURE_PREFIX = "__Secure-";
+    private final static String HOST_PREFIX = "__Host-";
+
+    private final static int MAX_NAME_VALUE_LENGTH = 4096;
+    private final static int MAX_ATTRIBUTE_VALUE_LENGTH = 1024;
 
     // IMPORTANT!
     // These private static variables must be treated as immutable and never exposed outside this class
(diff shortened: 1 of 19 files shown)


ok2c:
@arturobernalg `RFC 6265bis` is still a draft, it is not? What is so important about this feature that we need to rush it and implement a draft RFC which has not been accepted as standard yet?

arturobernalg:
> @arturobernalg `RFC 6265bis` is still a draft, it is not? What is so important about this feature that we need to rush it and implement a draft RFC which has not been accepted as standard yet?

@ok2c 
You're right, it's still a draft — but the __Host-/__Secure- prefix and `SameSite` bits it covers are already universal, interoperable browser behaviour, and the prefix checks are a real acceptance/security gap in our cookie store  today. Happy to scope it to just those stable parts and mark it @Experimental, or hold it until 6265bis is published

ok2c:
@arturobernalg I honestly have no idea how much those new attributes are being used / relied upon in everyday life. I have an impression HTTP cookies are getting used less and less. I was even going to propose to disable automatic cookie handling by default.

Anyway, I leave it up to you to make a call here. If you resolve the conflicts I will take another look at the parsing code.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/cookie/RFC6265CookieSpec.java
```diff
@@ -152,22 +159,29 @@ public String toString() {
 
     private RawCookie parseCookie(final CharSequence buffer,
                                   final Tokenizer.Cursor cursor) {
-        final String name = tokenParser.parseToken(buffer, cursor, TOKEN_DELIMS);
-        if (name.isEmpty()) {
-            return null;
+        final String token = tokenParser.parseToken(buffer, cursor, TOKEN_DELIMS);
```
ok2c, line 162: @arturobernalg Is this more complex parsing logic something that `6265bis` mandates?

I am not going to object to these changes as long as you confirm you know what you are doing.

arturobernalg: @ok2c 
Yes. The more permissive parsing is required by the 6265bis user-agent algorithm.
In particular, a Set-Cookie value without = is parsed as a nameless cookie, and the UA parser is intentionally more permissive than the server grammar.

ok2c: @arturobernalg Understood. I have no objections.

PR #870, opened 2026-08-13
Deprecate the never-cache-responses-with-query options and HTTP/1.0 response caching

arturobernalg:
Per the discussion on the query-string heuristic-caching report: assigning heuristic freshness to a response with a query component is no longer prohibited by the HTTP caching specification, so the `CacheConfig` options that suppress caching of such responses no longer serve a purpose.
This deprecates isNeverCacheHTTP10ResponsesWithQuery, isNeverCacheHTTP11ResponsesWithQuery and their builder setters, and deprecates HTTP/1.0 response caching as a whole (noted in the `CacheConfig` Javadoc and at the version check in `ResponseCachingPolicy`). Behaviour is unchanged — the options are still honoured — and an origin that does not want a response cached should send an explicit directive such as Cache-Control: no-cache.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
@@ -129,14 +129,17 @@ class ResponseCachingPolicy {
             return false;
         }
 
-        if (request.getPath().contains("?")) {
-            if (neverCache1_0ResponsesWithQueryString && from1_0Origin(response)) {
-                LOG.debug("Response is not cacheable as it had a query string");
-                return false;
-            } else if (!neverCache1_1ResponsesWithQueryString && !isExplicitlyCacheable(cacheControl, response)) {
-                LOG.debug("Response is not cacheable as it is missing explicit caching headers");
-                return false;
-            }
+        if (neverCache1_0ResponsesWithQueryString
+                && request.getPath().contains("?")
+                && from1_0Origin(response)) {
+            LOG.debug("Response is not cacheable as it had a query string");
+            return false;
+        }
+        if (neverCache1_1ResponsesWithQueryString
+                && request.getPath().contains("?")
+                && !isExplicitlyCacheable(cacheControl, response)) {
+            LOG.debug("Response is not cacheable as it is missing explicit caching headers");
+            return false;
         }
 
         if (cacheControl.isMustUnderstand()) {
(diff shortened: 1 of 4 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
ok2c, line 102: @arturobernalg Damn it. This is wrong. In fact the entire logic is based on the protocol version _hint_ and not t he real transport version. It is broken and wrong.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
```diff
@@ -105,6 +105,7 @@ public boolean isResponseCacheable(final RequestCacheControl requestCacheControl
             }
             return false;
         }
+        // HTTP/1.0 response caching is deprecated and will be rejected here in a future release.
```
ok2c: @arturobernalg Let's drop this comment altogether. I am going to remove the request version check from this class shortly/

ok2c (approved):
@arturobernalg Looks good otherwise.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
ok2c: @arturobernalg I am sorry. I overlooked this bit. It needs to be fixed.

ok2c (changes requested):

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/ResponseCachingPolicy.java
ok2c: @arturobernalg Do not you agree this bit of code should be refactored / fixed? 

```
        if (request.getPath().contains("?")) {
            if (neverCache1_0ResponsesWithQueryString && from1_0Origin(response)) {
                LOG.debug("Response is not cacheable as it had a query string");
                return false;
            } else if (!neverCache1_1ResponsesWithQueryString && !isExplicitlyCacheable(cacheControl, response)) {
                LOG.debug("Response is not cacheable as it is missing explicit caching headers");
                return false;
            }
        }
```

Should not it be something like this, so that in normal execution flow those bits would never get executed?
```
        if (neverCache1_0ResponsesWithQueryString && 
                request.getPath().contains("?") && 
                from1_0Origin(response)) {
            LOG.debug("Response is not cacheable as it had a query string");
            return false;
        }
        if (neverCache1_1ResponsesWithQueryString && 
                request.getPath().contains("?") && 
                !isExplicitlyCacheable(cacheControl, response)) {
            LOG.debug("Response is not cacheable as it is missing explicit caching headers");
            return false;
        }
```

arturobernalg: @ok2c  please do another pass

ok2c (approved):

ok2c:
@arturobernalg I tweaked some of your test case changes to keep the original asserts. I will also port the `neverCache1_1ResponsesWithQueryString` fix to the stable branch.

PR #872, opened 2026-08-14
Verify the server authentication (rspauth) provided in the Digest Authentication-Info response per RFC 7616

arturobernalg:
The client recomputes `rspauth` from the session state of the last request and compares it to the value returned by the server in constant time, so mutual authentication no longer depends solely on the request being accepted. For `qop=auth` the `cnonce` and `nc` echoed by the server are required and must match the values used for that request, and an Authentication-Info that states `qop=auth` must carry `rspauth`, `cnonce` and `nc`. Verification is confined to `qop=auth;` the `rspauth` of `qop=auth-int` hashes the response body, which is not available to the scheme. An Authentication-Info without `rspauth` remains acceptable unless the server states `qop=auth`.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
@@ -36,6 +36,7 @@ import java.security.MessageDigest;
 import java.security.Principal;
 import java.security.SecureRandom;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Formatter;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java
```diff
@@ -182,12 +201,36 @@ public String getRealm() {
...
+    }
+
     @Override
     public void processChallenge(
             final AuthChallenge authChallenge,
             final HttpContext context) throws MalformedChallengeException {
+        try {
+            processChallenge(null, true, authChallenge, context);
+        } catch (final AuthenticationException ex) {
+            throw new MalformedChallengeException(ex.getMessage(), ex);
```
michael-o: This basically duplicates the message, no?

arturobernalg: @michael-o 
You're right, wrapping with ex.getMessage() just duplicated the cause's message. And that catch was effectively dead anyway. please do another pass

michael-o (approved):
I have no formal objection, but cannot reasonable judge on the actual code.

ok2c (approved):
@arturobernalg Likewise, I could not spot anything objectionable code-wise, but cannot comment as far as RFC 7616 conformance is concerned. I trust you know what you are doing.

PR #874, opened 2026-08-16
treat malformed s-maxage as absent in shared-cache auth check

dxbjavid:
A shared cache must not store a response to a request that carried an Authorization header unless the response permits it through s-maxage, must-revalidate or public (RFC 9111 3.5). The guard treats any s-maxage as qualifying, but a malformed value like s-maxage=foo is parsed as 0 rather than -1, so it looks present and the authenticated response is stored in the shared cache and can later be handed to other clients on a revalidated hit. This tracks whether a valid s-maxage was actually supplied and uses that in the check, so a malformed directive is treated as absent as the spec requires; the numeric value is left unchanged so freshness behaviour is not affected.

### DIFF
--- a/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheSupport.java
+++ b/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheSupport.java
@@ -209,7 +209,9 @@ public final class CacheSupport {
             return ageValue;
         } catch (final NumberFormatException ignore) {
         }
-        return 0;
+        // A value that is not a valid delta-seconds token is invalid and must be treated as absent (RFC 9111 4.2.1),
+        // not folded into a usable 0 which would look like a genuine directive.
+        return -1;
     }
 
 }
--- a/httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/CacheControlParserTest.java
+++ b/httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/CacheControlParserTest.java
@@ -62,7 +62,8 @@ class CacheControlParserTest {
     void testParseInvalidCacheValue() {
         final Header header = new BasicHeader("Cache-Control", "max-age=invalid");
         final ResponseCacheControl cacheControl = parser.parseResponse(Collections.singletonList(header).iterator());
-        assertEquals(0L, cacheControl.getMaxAge());
+        // A malformed delta-seconds value is invalid and must be treated as absent (-1), not a usable 0.
+        assertEquals(-1L, cacheControl.getMaxAge());
     }
 
     @Test
--- a/httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/TestCacheSupport.java
+++ b/httpclient5-cache/src/test/java/org/apache/hc/client5/http/impl/cache/TestCacheSupport.java
@@ -106,7 +106,7 @@ class TestCacheSupport {
         Assertions.assertEquals(-1L, CacheSupport.deltaSeconds("-100"));
         Assertions.assertEquals(-1L, CacheSupport.deltaSeconds(""));
         Assertions.assertEquals(-1L, CacheSupport.deltaSeconds(null));
-        Assertions.assertEquals(0L, CacheSupport.deltaSeconds("huh?"));
+        Assertions.assertEquals(-1L, CacheSupport.deltaSeconds("huh?"));
         Assertions.assertEquals(2147483648L, CacheSupport.deltaSeconds("2147483648"));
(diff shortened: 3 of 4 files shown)


httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/ResponseCacheControl.java
```diff
@@ -60,6 +60,13 @@ public final class ResponseCacheControl implements CacheControl {
      * The shared-max-age directive value.
      */
     private final long sharedMaxAge;
+    /**
+     * Indicates whether a valid {@code s-maxage} directive was present in the header. This is distinct from
+     * {@link #sharedMaxAge} being {@code >= 0}, since a syntactically present but malformed value (for example
+     * {@code s-maxage=foo}) is stored as {@code 0} rather than {@code -1}. RFC 9111 treats such an invalid
+     * directive as absent, so this flag records whether the directive genuinely applies.
+     */
+    private final boolean sharedMaxAgePresent;
```
ok2c: @dxbjavid Any chance we could avoid this extra boolean? Can we use something line `-Integer#MAX_VALUE` to represent an invalid value?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -244,6 +247,19 @@ private static long parseSeconds(final String name, final String value) {
         return delta;
     }
 
+    private static boolean isDeltaSeconds(final String value) {
```
ok2c: @dxbjavid Could we move this logic into `#parseSeconds`?

httpclient5-cache/src/main/java/org/apache/hc/client5/http/cache/ResponseCacheControl.java
dxbjavid: good point, dropped the boolean. parseSeconds now returns -Integer.MAX_VALUE when the value isn't a valid delta-seconds, so a malformed directive is distinct from both absent (-1) and a genuine 0. the shared-cache auth check just treats anything below zero as not present, and i guarded the freshness paths so the sentinel is never used as a duration.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
dxbjavid: moved it in, the malformed check lives inside parseSeconds now and returns the sentinel there rather than being handled at the call site.

httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CacheControlHeaderParser.java
```diff
@@ -238,12 +245,31 @@ public final RequestCacheControl parse(final HttpRequest request) {
 
     private static long parseSeconds(final String name, final String value) {
         final long delta = CacheSupport.deltaSeconds(value);
-        if (delta == -1 && LOG.isDebugEnabled()) {
```
ok2c, line 241: @dxbjavid Why not making it even simpler?

```
public static long deltaSeconds(final String s) {
    if (TextUtils.isEmpty(s)) {
        return -1;
    }
    try {
        long ageValue = Long.parseLong(s);
        if (ageValue < 0) {
            ageValue = -1;  // Handle negative age values as invalid
        } else if (ageValue > Integer.MAX_VALUE) {
            ageValue = MAX_AGE.toSeconds();
        }
        return ageValue;
    } catch (final NumberFormatException ignore) {
        return -Integer.MAX_VALUE;
    }
}
```

ok2c: @dxbjavid In fact my original implementation of `#deltaSeconds` is wrong. It should be returning `-1`, not `0` if the value cannot be parsed.

dxbjavid: agreed, that's the cleaner root cause. i've pulled it into deltaSeconds so an unparseable value now returns -1 instead of 0, which means a malformed s-maxage reads as absent everywhere it's used and the original shared-cache auth check catches it with no extra plumbing. that let me drop the sentinel and revert the parser and policy changes entirely, so the fix is now just the one line in deltaSeconds plus the two tests that were asserting the old 0. full cache suite is green.

ok2c (approved):

ok2c:
@dxbjavid cherry-picked to `5.6.x`

dxbjavid:
TY!

PR #875, opened 2026-08-18
Add a model and parser for the RFC 9209 Proxy-Status response field

arturobernalg:
Adds a typed model of a single Proxy-Status member together with the standardized proxy error values, and a parser that reads the Proxy-Status response field according to the RFC 9209 and RFC 8941 Structured Fields grammar. The model is immutable and preserves the Structured Fields value types; the parser validates the standardized and error-specific parameter types and ignores unrelated extension parameters.

The code lives in the client for now and provides the field model and parsing only; it is not integrated with retries, exceptions, execution policy or telemetry.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/ProxyStatus.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/ProxyStatus.java
@@ -1,0 +1,229 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


ok2c (commented):
@arturobernalg I have no objections, but it would be nice to see how exactly you intent to use these classes. Feel free to merge.

arturobernalg:
> @arturobernalg I have no objections, but it would be nice to see how exactly you intent to use these classes. Feel free to merge.

@ok2c I’ve added a small runnable example based on the RFC 9209 504 Gateway Timeout / connection_timeout example.It shows how `ProxyStatusSupport` and `ProxyStatus` can be used directly from an `HttpClient` response.

ok2c (commented):
@arturobernalg I have no objections

PR #876, opened 2026-08-19
decode punycode public suffixes in PublicSuffixMatcher.verify

dxbjavid:
PublicSuffixMatcher.verify (through verifyInternal) resolves the domain against the suffix rules without the normalisation and punycode decoding that getDomainRoot and matches already apply, and the bundled list holds IDN suffixes in their Unicode form. So an ACE-encoded public suffix such as xn--h-2fa.no matches no rule and verify returns true, which lets the cookie PublicSuffixDomainFilter treat a whole IDN TLD as a registrable domain and accept a supercookie scoped to it, even though matches recognises the same suffix correctly. This decodes and lowercases the input in verifyInternal the same way getDomainRoot does, so both the ACE and Unicode forms are rejected consistently while genuine registrable subdomains under an IDN suffix are still allowed.

### DIFF
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
@@ -254,7 +254,14 @@ public final class PublicSuffixMatcher {
 
     @Internal
     public boolean verifyInternal(final String domain) {
-        final DomainRootInfo domainRootInfo = resolveDomainRoot(domain, null);
+        // Match the normalisation performed by getDomainRoot: the rules are held lowercase and in
+        // Unicode form, so an ACE-encoded (xn--) or mixed-case public suffix has to be decoded here
+        // as well, otherwise it fails to match a rule and is mistaken for a registrable domain.
+        String normalized = DnsUtils.normalize(domain);
+        if (normalized != null && normalized.contains("xn-")) {
+            normalized = IDN.toUnicode(normalized);
+        }
+        final DomainRootInfo domainRootInfo = resolveDomainRoot(normalized, null);
         if (domainRootInfo == null) {
             return false;
         }
--- a/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
+++ b/httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
@@ -249,19 +249,20 @@ public final class PublicSuffixMatcher {
          if (domain == null) {
              return false;
          }
-         return verifyInternal(domain.startsWith(".") ? domain.substring(1) : domain);
+         // Normalise here so that verifyInternal can assume its input is already lowercase and in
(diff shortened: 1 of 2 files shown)


ok2c (commented):
@dxbjavid Could we pull the normalization logic up into `#verify` method and leave `#verifyInternal` as is under assumption its input is always normalized?

dxbjavid:
makes sense, done. normalisation now happens in verify and verifyInternal just assumes a normalised input.

one thing worth flagging: DefaultHostnameVerifier.matchIdentity calls verifyInternal directly rather than through verify, so with the logic moved up it now relies on the identity already being in normalised form. that matches the pre-existing behaviour on master, but if you'd like that path covered too i'm happy to normalise the identity at that call site.

httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
```diff
@@ -249,7 +249,15 @@ public boolean verify(final String domain) {
          if (domain == null) {
              return false;
          }
-         return verifyInternal(domain.startsWith(".") ? domain.substring(1) : domain);
+         // Normalise here so that verifyInternal can assume its input is already lowercase and in
+         // Unicode form. The rules are held that way, so an ACE-encoded (xn--) or mixed-case public
+         // suffix has to be decoded first, mirroring getDomainRoot; otherwise it fails to match a
+         // rule and is mistaken for a registrable domain.
+         String normalized = DnsUtils.normalize(domain.startsWith(".") ? domain.substring(1) : domain);
```
ok2c: @dxbjavid Could not we use `DnsUtils#normalizeUnicode` here?

ok2c:
> one thing worth flagging: DefaultHostnameVerifier.matchIdentity calls verifyInternal directly rather than through verify, so with the logic moved up it now relies on the identity already being in normalised form. that matches the pre-existing behaviour on master, but if you'd like that path covered too i'm happy to normalise the identity at that call site.

The reason `#verifyInternal` exists at all is to avoid having to run the input through the normalization process multiple times. `DefaultHostnameVerifier#matchIdentity` normalizes the input at the beginning of the routine and then passes it normalized to the PSM.

httpclient5/src/main/java/org/apache/hc/client5/http/psl/PublicSuffixMatcher.java
dxbjavid: good call, switched to normalizeUnicode. it also drops the fragile contains("xn-") guard and handles the IllegalArgumentException from toUnicode for us, so it is a bit tidier than the getDomainRoot version.

dxbjavid:
right, that clears up my earlier worry then. since matchIdentity already normalises before it reaches the PSM, verifyInternal keeping its normalised-input assumption is fine and there is nothing extra needed at that call site.

ok2c:
@dxbjavid cherrry-picked to `5.6.x`

dxbjavid:
Thanks for the merge.

PR #881, opened 2026-08-24
Add support for Compression Dictionary Transport (RFC 9842)

arturobernalg:
Implement dictionary negotiation, storage, matching, and dcb / dcz decoding


ok2c:
@arturobernalg This is a lot of new code. I will scan it for obvious programming errors or inefficiencies. I, however, cannot do a proper review as far as its conformance to the RFC is concerned. I will have to trust you know what you are doing.

One question, though. Is there a reason this is an async only feature?

arturobernalg:
> One question, though. Is there a reason this is an async only feature?
@ok2c 
No fundamental reason. I initially limited the implementation to the async execution chain because that was the use case I was targeting and the capture/decoding code was based on `AsyncDataConsumer`. The protocol itself is equally applicable to the classic client, so the classic execution chain should support it as well.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -188,6 +313,233 @@ public void failed(final Exception ex) {
...
+            }
+        }
+        return false;
+    }
+
+    private static URI resolveRequestUri(
+            final HttpRequest request,
+            final AsyncExecChain.Scope scope) {
+        try {
+            if (scope != null) {
```
ok2c: @arturobernalg The scope can never be null. Please simplify.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -188,6 +313,233 @@ public void failed(final Exception ex) {
...
+                }
+            }
+
+            final URI requestUri = request.getUri();
+            if (requestUri.isAbsolute()) {
+                return requestUri;
+            }
+
+            if (scope != null) {
+                final URI baseUri = URI.create(scope.route.getTargetHost().toURI() + "/");
```
ok2c: @arturobernalg This looks wrong. You should be using the request authority here, not the target host.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -188,6 +313,233 @@ public void failed(final Exception ex) {
...
+    private CompressionDictionary prepareDictionaryNegotiation(
+            final HttpRequest request,
+            final boolean enabled,
+            final CompressionDictionary candidate,
+            final List<String> requestDictionaryAcceptTokens) throws HttpException {
+        if (!enabled) {
+            return null;
+        }
+
+        if (request.containsHeader(HttpHeaders.ACCEPT_ENCODING)) {
```
ok2c, line 354: @arturobernalg This whole `Accept-Encoding` parsing business looks really bad and terribly inefficient.

ok2c (commented):
@arturobernalg I must admit i do not quite understand what is going on here, but some bits, especially header parsing ones do not good enough to me. I cannot make a call on this change-set

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -188,6 +310,175 @@ public void failed(final Exception ex) {
...
+            final HttpRequest request,
+            final boolean enabled,
+            final CompressionDictionary candidate,
+            final List<String> requestDictionaryAcceptTokens) {
+        if (!enabled) {
+            return null;
+        }
+
+        // A caller-provided Accept-Encoding is honoured verbatim: the client neither rewrites it
+        // nor negotiates a dictionary on top of an explicit choice.
```
ok2c, line 353: @arturobernalg Exactly

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -188,6 +310,175 @@ public void failed(final Exception ex) {
...
+            final URI requestUri,
+            final CookieStore privacyPartition) {
+        if (compressionDictionaryStore == null
+                || privacyPartition == null
+                || requestUri == null
+                || !URIScheme.HTTPS.same(requestUri.getScheme())) {
+            return null;
+        }
+
+        final Header[] headers = response.getHeaders(CompressionDictionaryHeaderSupport.USE_AS_DICTIONARY);
```
ok2c: @arturobernalg This could be done much, much more efficiently. You are parsing and copying header values into a comma delimited list and then parsing it again. Consider using `MessageSupport#parseHeaders` or `MessageSupport#parseHeadersStrict` instead.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
```diff
@@ -188,6 +309,167 @@ public void failed(final Exception ex) {
...
+            final URI requestUri,
+            final CookieStore privacyPartition) {
+        if (compressionDictionaryStore == null
+                || privacyPartition == null
+                || requestUri == null
+                || !URIScheme.HTTPS.same(requestUri.getScheme())) {
+            return null;
+        }
+
+        if (!response.containsHeader(CompressionDictionaryHeaderSupport.USE_AS_DICTIONARY)) {
```
ok2c: @arturobernalg I presume `UseAsDictionary#parse` should be smart enough to return `null` if no dictionary header is present which would make this call unnecessary and overall parsing routine a little more efficient.

ok2c (commented):
@arturobernalg I have no objections, just one minor comment. This is now your area of responsibility. 
Feel free to merge the change-set if you think it is ready or make symmetric changes to the classic transport first. Anyway, there is not much I can do beyond code conceptual sanity checks, parsing improvements, etc.

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/ContentCompressionAsyncExec.java
arturobernalg: @ok2c  please do another pass

ok2c: @arturobernalg Looks good to me. I have no further comments.

PR #884, opened 2026-09-17
HTTPCLIENT-2433: Prevent concurrent HTTP/2 request body replay

mkurz:
This fixes [HTTPCLIENT-2433](https://issues.apache.org/jira/browse/HTTPCLIENT-2433).

When an HTTP/2 authentication challenge completed before the request body had finished, `H2AsyncMainClientExec` previously notified the execution chain that the exchange was complete immediately. The authentication handler could then release and reuse the same repeatable entity producer while the original HTTP/2 stream was still consuming it, corrupting the replayed request body.

The HTTP/2 execution handler now completes the exchange only after both request output and response input have terminated. It also handles a graceful `RST_STREAM(NO_ERROR)` after a complete response as request-side termination, allowing authentication replay to continue without hanging.

Regression coverage includes:

- Uploading a deterministic 2 MB file through challenge-based Basic authentication and verifying that the replay contains the exact original bytes.
- A body-less `401` followed by `RST_STREAM(NO_ERROR)`, verifying that the authenticated retry completes with the exact 2 MB request body.

Reproducer:
https://github.com/mkurz/apache-httpclient5-h2-auth-replay-reproducer

Tests:

- HTTP/2 authentication suites with and without TLS: 28 tests, 0 failures
- Graceful-reset regression test repeatedly verified with and without TLS


ok2c:
@mkurz The test coverage looks awesome! However I need a little while to digest the proposed fix. Please bear with me.

ok2c (commented):
@mkurz Your suggested fix is perfectly fine. 

For the sake of code consistency could we use the same approach as used in [HttpAsyncMainClientExec](httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/HttpAsyncMainClientExec.java#L115) using `AtomicInteger` instead of two `AtomicBoolean`?

mkurz:
@ok2c done!

mkurz:
@ok2c force pushed, diff is here: https://github.com/apache/httpcomponents-client/compare/17cf04895b9f51064c46965d9782c06e74b89230..8483a2a27120582cc5559cbe2a267f95a897c210#diff-274e1a4b963f654fc7d082e31c4198bd09238d6c3b55dbb6c07564f50a2112bf

ok2c:
Cherry-picked to `5.6.x`

ok2c:
@mkurz Many thanks for reporting the problem and contributing the fix

mkurz:
@ok2c thanks for merging them ;)

mkurz:
@ok2c do you have an eta when next core and client patch release will be cut?

ok2c:
@mkurz Soon. I will try to cut the core releases this weekend. The client bug I see as less severe as it affects H2 only client, so it may take a bit longer.

ok2c:
@mkurz I made some major simplifications to the main async exec interceptor code with 4a7295895bafa74c021e466910568965ff9fcc57. Please do feel free to double-check. If you confirm the change-set does not break anything on your end, I will cherry-pick it to `4.6.x` and will include in the upcoming release. If you do not want to mess with master, you can try to work with `5.6.x` instead and cherry-pick the change-set on top of the head.

mkurz:
I tested the simplification on master and cherry-picked it cleanly onto the current `5.6.x`, which already contains the HTTPCLIENT-2433 backport. Both regression tests pass with H2 and H2+TLS, the complete async suites pass on both branches, and the standalone reproducer passed three times on each branch with the exact 2,000,000-byte body and expected SHA-1.

Looks good to me @ok2c, thanks!

ok2c:
Thanks @mkurz Much appreciated.

PR #885, opened 2026-09-18
HTTPCLIENT-2434 - Improve idle connection eviction configuration.

arturobernalg:
Add an overload of evictIdleConnections that allows the eviction interval to be configured independently of the maximum idle time. The existing overload retains the current automatically calculated eviction interval.


httpclient5/src/test/java/org/apache/hc/client5/http/examples/ClientEvictExpiredConnections.java
```diff
@@ -31,30 +31,27 @@
...
 import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.apache.hc.core5.http.message.StatusLine;
-import org.apache.hc.core5.pool.PoolStats;
 import org.apache.hc.core5.util.TimeValue;
 
 /**
- * Example demonstrating how to make sure connections connections expire
- * after 5 seconds of inactivity and how to proactively evict expired connections
- * from the connection pool.
+ * Example demonstrating how to configure connection keep-alive and
```
ok2c: @arturobernalg I think we have a conflict here. Could you please pick up my changes?

httpclient5/src/main/java/org/apache/hc/client5/http/impl/async/H2AsyncClientBuilder.java
```diff
@@ -734,7 +734,10 @@ public final H2AsyncClientBuilder disableAuthCaching() {
      * get closed and evicted from the pool.
      *
      * @return this instance.
+          * @deprecated Configure connection keep-alive settings appropriately and use
+     * {@link #evictExpiredConnections()} instead.
```
ok2c, line 738: @arturobernalg I would also link `RequestConfig#connectionKeepAlive` here.

ok2c: @arturobernalg Looks good. Could we please also add `@see RequestConfig#connectionKeepAlive`? to all deprecated methods?

ok2c (approved):

